PR #6, opened 2019-10-08
Fewer CI jobs

cmyr:


madig:
Still irks me that CI jobs are run twice, on the `push` and the `pull_request`. There must be a way to turn off `push` builds for anything other than tagged commits.

raphlinus:
That doesn't seem unreasonable to me. You definitely want both in the case that the merge is not trivial (ie not a fast-forward).

cmyr:
I think they're building different commits; `push` builds the commit that is pushed, and `pull_request` builds the merge commit, which is generated automatically by github?

madig:
> You definitely want both in the case that the merge is not trivial

What do you mean? I know this behavior from other CI systems, where it's really just the same CI running on the same code, but twice because the PR happens to be coming from a branch of the same repository. It doesn't happen if you send a PR from a fork to the upstream repository because it doesn't trigger a `push` event to the upstream repo.

Maybe the following works:

```YAML
on:
  - push:
      # Sequence of patterns matched against refs/heads
      branches:    
        - master         # Push events on master branch
      # Sequence of patterns matched against refs/tags
      tags:        
        - "*"
  - pull_requests
```

cmyr:
Also I think I'm wrong, it doesn't appear that `pull_request` runs on the auto-generated merge commit. I could swear this was happening in some other situation when I was testing, but it doesn't appear to be happening here. 😕

madig:
Do you mean it should run the CI after merging? I think this may need a more complicated setup...

cmyr:
no, github internally creates a merge commit for you, whether you merge or not. This commit only exists on github, but is exposed via certain APIs. I think this is how they show you whether you have conflicts with the base branch, etc.

madig:
My snippet above does not werk. Hrm.

cmyr:
I  think you want just `branches: master`.

madig:
But you also want to catch tags, unless you want a separate workflow for those. And releases.

cmyr:
ah right I was confused in any case, [the example](https://help.github.com/en/articles/workflow-syntax-for-github-actions#onpushpull_requestbranchestags) looks an awful lot like your snippet. 😕

madig:
That's what I copy-pasted, but how to insert the `pull_request` thing? I apparently can't insert a string next to a map as above.

madig:
This seems to do something:
```YAML
on:
  push:
    branches:
      - master
    tags:
      - "*"
  pull_request: {}
```

PR #7, opened 2019-10-15
A batch of changes motivated by runebender

cmyr:
this has been my working branch for the past week or two, and I'm happy enough with these changes that I'd like to bring them in.

Major things:

- we now just load everything up front, which makes life simpler
- we use `Arc<str>` for glyph names, and keep loaded glyphs behind `Arc`s as well, so that we can play more nicely with `Data`. 
- more stuff is public
- there's a `druid` feature that turns on various conversions and `Data` impls.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,3 @@ druid = { git="https://github.com/xi-editor/druid/", rev="22dfeae", optional=tru
 
 [dev-dependencies]
 failure = "0.1.5"
-
-[features]
-druid_data = ["druid"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ exclude = [
 ]
 
 [dependencies]
-plist = "0.4"
+plist = "0.5"
 serde = "1.0"
 serde_derive = "1.0"
 serde_repr = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,11 +16,11 @@ exclude = [
 
 [dependencies]
 plist = "0.5"
-serde = "1.0"
+serde = { version =  "1.0", features = ["rc"] }
 serde_derive = "1.0"
 serde_repr = "0.1"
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -15,11 +15,12 @@ exclude = [
...
 
 [dependencies]
-plist = "0.4"
-serde = "1.0"
+plist = "0.5"
+serde = { version =  "1.0", features = ["rc"] }
 serde_derive = "1.0"
 serde_repr = "0.1"
 quick-xml = "0.12.0"
+druid = { git="https://github.com/xi-editor/druid/", rev="8164adf", optional=true }
```
raphlinus: Is this for the Data trait? Seems *almost* like a wrong-way dep...

Hmm, now I see, also druid's piet's color type.

cmyr: and kurbo's affine, want to have it pinned to druid instead of those crates specifically. 🤷‍♂

madig:
Have you tried loading e.g. a large CJK font? What's the performance diff between old and new?

cmyr:
Edit: just for fun I plugged in [rayon](https://github.com/rayon-rs/rayon) and got this down to 2/3 seconds, and from what I can see there's lots of overhead there, so I think we should be able to do better. In any case I think this is okay for now!

madig:
Sounds fantastic! Sure, bend it for runebender for now. I just always side-eye Rust for eventually rewriting all the font reading/writing libs in it for maximum greatness, so I have a scripting bend.

madig:
For reference, what Noto CJK font file did you load, and from where?

cmyr:
I'm currently using NotoSansCJKhk-Regular, and I made a UFO from an otf using font-extractor and defcon.

cmyr:
annoyingly I don't seem to have committed the little tool I wrote to do this loading, will do that

PR #14, opened 2019-11-02
Add defaults for MetaInfo

madig:
Is this  how you make default arguments? I'm missing my `fn Ufo::new(meta = MetaInfo::default())`.

### DIFF
--- a/src/ufo.rs
+++ b/src/ufo.rs
@@ -51,12 +51,18 @@ pub enum FormatVersion {
 
 /// The contents of the [`metainfo.plist`] file.
 ///
-/// [`metainfo.plist`]: http://unifiedfontobject.org/versions/ufo1/metainfo.plist/
+/// [`metainfo.plist`]: http://unifiedfontobject.org/versions/ufo3/metainfo.plist/
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct MetaInfo {
     pub creator: String,
     pub format_version: FormatVersion,
+}
+
+impl Default for MetaInfo {
+    fn default() -> Self {
+        MetaInfo { creator: "org.linebender.norad".to_string(), format_version: FormatVersion::V3 }
+    }
 }
 
 /// The contents of the [`fontinfo.plist`][] file.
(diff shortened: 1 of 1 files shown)


src/ufo.rs
```diff
@@ -51,14 +51,20 @@ pub enum FormatVersion {
...
-/// [`metainfo.plist`]: http://unifiedfontobject.org/versions/ufo1/metainfo.plist/
+/// [`metainfo.plist`]: http://unifiedfontobject.org/versions/ufo3/metainfo.plist/
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct MetaInfo {
     pub creator: String,
     pub format_version: FormatVersion,
 }
 
+impl Default for MetaInfo {
```
cmyr: This is sort of a philosophical point.

I feel that `Default` is intended to represent types that have a well-defined "empty" or "placeholder" value. I'm not sure this is true of `MetaInfo`, since I presume anybody constructing a `MetaInfo` may want to provide their own `creator` string?

In that case, I would advocate adding a `new` method to `MetaInfo`, that takes a `name`, but uses `v3`. If the user wants to use another version, they can always construct `MetaInfo` directly, like I [currently do in runebender](https://github.com/linebender/runebender/blob/master/src/data.rs#L180).

As for the overall question of allowing default arguments, a simple approach is to make the type of the argument an `Option<_>`, and then insert a default value in the body of the function if the argument was `None`. There's also a trick where you can have the type of an argument be `impl Into<Option<f64>>`, and then the user can just write `5.5` instead of `Some(5.5)` when calling the function.

Example:

```rust
const DEFAULT_CLOSENESS_THRESHOLD: f64 = 10.0;

fn points_are_close(p1: Point, p2: Point, min_distance: impl Into<Option<f64>>) -> bool {
    let min_distance = min_distance.unwrap_or(DEFAULT_CLOSENESS_THRESHOLD);
    (p2 - p1).hypot() <= min_distance
}
```

madig: `creator` usually means the library that writes the UFO, which is norad here. So, a `new` function that fills in both? :)

A `new`'d UFO object should always default to the latest version. So... `Option<MetaInfo>`?

cmyr: Okay, if `creator` is intended to be the producing library than I think your code is good; and if we had a `new` fn it should accept both arguments,.

PR #16, opened 2019-11-06
WIP: Implement de- and serialization for all fontinfo attributes

madig:
According to https://github.com/serde-rs/serde/issues/1301 and https://github.com/serde-rs/serde/issues/1444, easily de/serializing things with custom functions inside containers like Option is not possible, so you have to handle everything around it. Since fontinfo is quite expansive with more custom stuff (and de/serialization) to come, maybe it should move to its own module mid-term.


madig:
Todo:

- Implement validation for all fields
- Implement serialization?

src/ufo.rs
```diff
@@ -69,28 +72,352 @@ impl Default for MetaInfo {
...
-    pub ascender: Option<f64>,
-    pub italic_angle: Option<f64>,
-    pub note: Option<String>,
+    pub year: Option<u32>,
+}
+
+// TODO: validate!
+// http://unifiedfontobject.org/versions/ufo3/fontinfo.plist/#guidelines
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct Guideline {
```
cmyr: this type already exists in `src/glyph/mod.rs`. Can we reuse it? We might want to move it to a different file as well, but for now I think it's fine to just leave it there.

madig: Oh, hm, maybe, if validation and de/serialization can be shared.

madig:
I think I have to properly organize de/serializer code now. Also, need to parse guidelines into existing `glyph::Guideline`.

cmyr:
Okay, just ping me when you think this is ready!

madig:
So, now all possible fields actually get deserialized in the test. What's needed next is custom types (e.g. OS2Panose, OS2FamilyClass, the WOFF dir attributes that should be an enum, etc.) plus custom de/serializers to perform validation. The `glyph::Guideline` type should be reused for fontinfo and might need more extensive custom de/serialization. Not sure what to do about standard types that need validation, like the WOFF list attributes where the spec demands they contain at least one item.

madig:
Do Identifiers need their own `::new` and stuff? The UFO spec has an example algorithm.

madig:
Not sure I need all the `Visitor`s? Maybe doing the work in just `deserialize` would be enough.

cmyr:
Identifiers are annoying, I'm not sure how best to handle them. Probably yes, we should have a newtype and store them a s a string, but we should validate them on creation.

madig:
Same for e.g. `openTypeOS2VendorID`: 

> Four character identifier for the creator of the font. Corresponds to the OpenType OS/2 table achVendID field.

[OS/2 spec](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#achvendid), [tag spec](https://docs.microsoft.com/en-us/typography/opentype/spec/otff#data-types):

> Array of four uint8s (length = 32 bits) used to identify a table,  design-variation axis, script, language system, feature, or baseline
> [...]
> A Tag value is a uint8 array. Each byte within the array must have a value in the range 0x20 to 0x7E. This corresponds to the range of values of Unicode Basic Latin characters in UTF-8 encoding, which is the same as the printable ASCII characters. As a result, a Tag value can be re-interpreted as a four-character sequence, which is conventionally how they are referred to. Formally, however, the value is a byte array. When re-interpreted as characters, the Tag value is case sensitive. It must have one to four non-space characters, padded with trailing spaces (byte value 0x20). A space character cannot be followed by a non-space character.

So maybe both Tag and Identifier should be newtypes with a `::new` and/or `::from_string` or something.

madig:
Hrm. There's some small stuff that needs validation, like various `postscript*` list attributes have a length limit. I think it is much less work to implement a `is_valid` function on FontInfo that is run before serialization and after deserialization?

cmyr:
Yea, I think a validation function is totally reasonable.

madig:
Went through the spec again at http://unifiedfontobject.org/versions/ufo3/fontinfo.plist/ and found some type discrepancies. Also changed some Todo comments, the spec does not mandate checks for some attributes. The spec is also somewhat lenient with types, declaring lots of stuff `Integer` when the OT spec specifies int16, etc. I am thinking about introducing some type aliases like Integer = i32, NoneNegativeInteger = u32, just to be very spec-spirit compliant? http://unifiedfontobject.org/versions/ufo3/conventions/ does not define limits.

cmyr:
I'm happy either way with the aliases.

madig:
I was just informed that the UFO spec is going to grow a minor version soon, 3.1, so that the glyph element can grow a verticalOrigin attribute in advance element.

anthrotype:
https://github.com/unified-font-object/ufo-spec/issues/78 and https://github.com/unified-font-object/ufo-spec/issues/95

anthrotype:
> the UFO spec is going to grow a minor version soon, 3.1

also the GLIF `<glyph>` element will have a corresponding additional `formatMinor` attribute.

madig:
note to self:

- Complete validation
- Test FontInfo serialization?
- Error enums. Would they benefit any process other than loading? if Runebender ever generates invalid data, it probably can't recover easily?
- Document more. See how it looks in cargo doc.

src/shared_types.rs
```diff
@@ -0,0 +1,283 @@
...
+/// These identifiers are required to be unique within certain contexts
+/// as defined on a per object basis throughout this specification.
+/// Identifiers are specified as a string between one and 100 characters long.
+/// All characters must be in the printable ASCII range, 0x20 to 0x7E.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Identifier(pub(crate) String);
+
+impl Identifier {
+    fn is_valid(&self) -> bool {
+        if self.0.len() > 100 {
```
cmyr: I would probably write this as,

```rust
fn is_valid(&self) -> bool {
    self.0.len() <= 100
        && self.0.as_bytes().all(|b| b >= 0x20 && b < 0x7F)
}

File: src/shared_types.rs [hunk 3]
```diff
@@ -0,0 +1,283 @@
...
+            return false;
+        }
+        for c in self.0.chars() {
+            if !(0x20..=0x7F).contains(&(c as u8)) {
```
cmyr: `..=` is an *inclusive* range, that is it includes the last item; `0..=2` is the values `[0, 1, 2]`. An *exclusive* range is written `0..2`. 

This is a little rust thing to keep in mind, I've definitely gotten it confused in the past.

src/shared_types.rs
```diff
@@ -0,0 +1,283 @@
...
+    }
+}
+
+impl Serialize for Identifier {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        if !self.is_valid() {
+            return Err(ser::Error::custom("Identifier must be at most 100 characters long and contain only ASCII characters in the range 0x20 to 0x7F."));
```
cmyr: Another way to express this is to skip the `return` and have an `else` block, so:

```rust
if self.is_valid() {
    serializer.serialize_str(&self.0)
} else {
    Err(MyErr)
}
```

src/fontinfo.rs
```diff
@@ -0,0 +1,940 @@
...
+use crate::Error;
+
+// The specification is vague about data type limits, usually implicitly meaning
+// Python types. Since Python is dynamic, the spec does not nail down the exact type
+// in several locations and we have to assume the biggest type that can hold it.
+type Integer = i32;
+type NonNegativeInteger = u32;
+type IntegerOrFloat = f64;
+type Float = f64;
+type NonNegativeIntegerOrFloat = f64; // Must be validated to be non-negative.
```
cmyr: I would do this by making a newtype struct, `NonNegativeIntegerOrFloat(f64)`, which can only be constructed in a way that validates its input.

src/fontinfo.rs
```diff
@@ -0,0 +1,940 @@
...
+use serde::ser::{SerializeSeq, Serializer};
+use serde::{Deserialize, Serialize};
+
+use crate::shared_types::Guideline;
+use crate::Error;
+
+// The specification is vague about data type limits, usually implicitly meaning
+// Python types. Since Python is dynamic, the spec does not nail down the exact type
+// in several locations and we have to assume the biggest type that can hold it.
+type Integer = i32;
```
cmyr: is it worth making these 64 bit types? my gut says no, but it's strictly 'more' correct. (being fully correct would probably require some fancy real number type with arbitrary precision)

src/fontinfo.rs
```diff
@@ -0,0 +1,940 @@
...
+type NonNegativeInteger = u32;
+type IntegerOrFloat = f64;
```
cmyr: again the 'strictly most correct' way to deal with this would be to make it an enum with `Float` and `Integer` variants; but I suspect in practice the integers we're dealing with are in the range that can be fully represented by an `f64`. 🤷‍♀

src/fontinfo.rs
```diff
@@ -0,0 +1,940 @@
...
+        if let Some(v) = &self.open_type_head_created {
+            const DATE_LENGTH: usize = 19;
+            if v.len() != DATE_LENGTH {
+                return Err(Error::FontInfoError);
+            }
+            if !v.chars().all(|b| b.is_ascii_digit() || b == ' ' || b == '/' || b == ':') {
+                return Err(Error::FontInfoError);
+            }
+
+            if !(v[0..4].parse::<u16>().is_ok()
```
cmyr: for all these parse errors we _could_ do something like, `x.parse::<u8>().map_err(|_| Error::FontInfoError)?`

src/shared_types.rs
madig: Yes, I mean to include `0x7F` here :)

src/fontinfo.rs
madig: I stayed with 32 bits because sane values should stay inside the possible range.

src/shared_types.rs
cmyr: Oops, my mistake. I wouldn't normally consider 0x7F a printable character, but whatever the spec says.

madig: Ooooh! Actually you are right, spec says 0x20 to 0x7E. 😯

cmyr: 😅 💦

src/shared_types.rs
madig: Agreed, but boy do I find Rust's terseness alien coming from Python.

src/fontinfo.rs
madig: I think so. There aren't even any 64 bit values in the spec that I know of.

src/fontinfo.rs
madig: Hm. Is it worth it? It's only used for unitsPerEm which is either an int or a float but must be positive. We can also remove the `type` and just check for this in `validate`.

cmyr: either way is fine.

cmyr:
oh, one other suggestion here: run `cargo clippy` on this patch, and see what suggestions it comes up with.

madig:
That was a lot of lints. Fixed them all.

madig:
1. I also hope they make sense
2. I see no merge button

:astonished:

PR #17, opened 2019-11-27
Include a default layer with a _notdef glyph by default

cmyr:
This changse the behaviour of Ufo::default() to always create
the default layer, and populate it with the '.notdef' glyph.

I'm not 100% sure that this makes sense; it's motivated by wanting
to have a glyph defined when starting a new project in runebender.

### DIFF
--- a/src/ufo.rs
+++ b/src/ufo.rs
@@ -20,7 +20,6 @@ static DEFAULT_GLYPHS_DIRNAME: &str = "glyphs";
 static DEFAULT_METAINFO_CREATOR: &str = "org.linebender.norad";
 
 /// A Unified Font Object.
-#[derive(Default)]
 pub struct Ufo {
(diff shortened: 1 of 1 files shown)


madig:
I wouldn't do this. In TruFont, you can set the glyph set for new fonts. It's something that you set a default for (e.g. `A-Za-z0-9.,/space`) but not hardcode into norad.

cmyr:
okay, good info. I'll find a more private way to do this.

madig:
Revisiting this, I would however create a default layer in `Ufo::default()`. The population of it should be done elsewhere.

cmyr:
yea, I noticed that issue when working on this and on #18, and I agree.

madig:
I think all you need is a metainfo.plist, a layercontents.plist and a glyphs/ dir with at least contents.plist.

PR #18, opened 2019-11-28
Don't store glyph names

cmyr:
This moves to computing glyph names as needed, using only the
default layer. The previous 'design' didn't make any sense once
we can actually add glyphs to a layer, and was blocking
some runebender work.

### DIFF
--- a/src/ufo.rs
+++ b/src/ufo.rs
@@ -3,7 +3,6 @@
 #![deny(intra_doc_link_resolution_failure)]
 
 use std::borrow::Borrow;
-use std::collections::BTreeSet;
 use std::ffi::OsStr;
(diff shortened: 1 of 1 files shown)


madig:
Slight aside: The UFO spec allows layers to be completely independent from one another, i.e. any layer can have a disjoint set of glyph names in it. For the purpose of Runebender, it is however reasonable to assume that the default layer is the most important and other layers are subsets of it. Loading and saving a UFO should however not lose anything.

cmyr:
noted, thanks. I'm letting the API for this crate be guided, in the short-term, by just wanting to get runebender off the ground, but will definitely revisit this at some point in the future.

PR #21, opened 2019-12-04
Wire up basic object saving

cmyr:
A bunch of the file saving code was already in place, but just
wasn't really exposed or used at all.

This attempts to cover the bare minimum case; for the time being
we will only save files that are marked as having been created in norad.

### DIFF
--- a/src/ufo.rs
+++ b/src/ufo.rs
@@ -4,6 +4,7 @@
 
 use std::borrow::Borrow;
 use std::ffi::OsStr;
+use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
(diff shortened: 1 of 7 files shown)


src/ufo.rs
```diff
@@ -124,6 +126,42 @@ impl Ufo {
...
+        let path = path.as_ref();
+        self.save_impl(path)
+    }
+
+    fn save_impl(&self, path: &Path) -> Result<(), Error> {
+        if self.meta.creator.as_str() != DEFAULT_METAINFO_CREATOR {
+            return Err(Error::NotCreatedHere);
+        }
+
+        fs::remove_dir_all(path)?;
```
madig: Fails when dir doesn't already exist. Should be `if path.exists(): remove it`.

cmyr: ah yea I bumped into that when trying to use this too, have it patched locally but forgot to update. :)

madig: You may want an `overwrite: bool` parameter on the function.

PR #22, opened 2019-12-09
Intern glyph names

cmyr:
This began with me just wanting to change Component to use
GlyphName instead of String, but it felt bad that we were going to
be storing multiple copies of the names, so I also implemented
a basic interning scheme; while parsing we pass along a HashSet of
names; before using a name, we check that it doesn't already exist in
the map.

When we implement parallel loading we will need to put this HashSet
behind a RwLock or similar, but that shouldn't be too difficult.

### DIFF
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -5,6 +5,7 @@ mod serialize;
 #[cfg(test)]
 mod tests;
 
+use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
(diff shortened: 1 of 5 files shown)


madig:
(Glyph names are per-layer.)

cmyr:
only in the sense that components don't cross layer boundaries? Certainly multiple glyphs on different layers can have the same name, and this saves us from duplicating data in that case? 🤷‍♀

madig:
(Yes, but layer A can have x, y z while layer B has a, b, c. Layers are independent of each other.)

cmyr:
Okay, makes sense; we'll reuse names if they're the same, not where they aren't.

PR #23, opened 2020-01-18
Implement more of the UFO spec

madig:
1. When serializing, order glif attributes like in the specification: http://unifiedfontobject.org/versions/ufo3/glyphs/glif/ -- that is what https://github.com/unified-font-object/ufoNormalizer/ is doing.
2. Do not serialize default glif attribute values.
3. Implement loading and serializing features.fea, groups.plist, kerning.plist and lib.plist.

### DIFF
--- a/src/glyph/serialize.rs
+++ b/src/glyph/serialize.rs
@@ -1,18 +1,18 @@
 //! Writing out .glif files
 
+use std::io::{Cursor, Write};
+
+use super::{
+    Advance, AffineTransform, Anchor, Color, Component, Contour, ContourPoint, GlifVersion, Glyph,
+    Guideline, Identifier, Image, Line, PointType,
+};
+
+use crate::error::GlifWriteError;
+
 use quick_xml::{
     events::{BytesDecl, BytesEnd, BytesStart, BytesText, Event},
     Error as XmlError, Writer,
 };
-use std::f32;
-use std::io::{Cursor, Write};
-
-use super::{
-    Advance, Anchor, Color, Component, Contour, ContourPoint, GlifVersion, Glyph, Guideline,
-    Identifier, Image, Line, PointType,
-};
-
-use crate::error::GlifWriteError;
 
 impl Glyph {
     pub(crate) fn encode_xml(&self) -> Result<Vec<u8>, GlifWriteError> {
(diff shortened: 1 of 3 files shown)


src/glyph/serialize.rs
```diff
@@ -1,11 +1,11 @@
 //! Writing out .glif files
 
-use std::io::{Cursor, Write};
-
 use quick_xml::{
     events::{BytesDecl, BytesEnd, BytesStart, BytesText, Event},
     Error as XmlError, Writer,
 };
+use std::f32;
```
cmyr: hmm, did cargo fmt do this? The 'standard' ordering of imports is std -> extern crates -> internal types; although this is just a convention.

src/ufo.rs
```diff
@@ -97,15 +117,55 @@ impl Ufo {
...
             } else {
                 None
             };
 
+            let lib_path = path.join(LIB_FILE);
+            let lib = if lib_path.exists() {
+                // Value::as_dictionary(_mut) will only borrow the data, but we want to own it.
+                match plist::Value::from_file(lib_path)? {
+                    plist::Value::Dictionary(dict) => Some(dict),
+                    _ => None, // XXX: Return error
```
cmyr: I've opened https://github.com/ebarnard/rust-plist/pull/48 to add `into_dictionary`; maybe reference that issue here, and we can change this if that patch is merged?

src/ufo.rs
```diff
@@ -243,6 +331,42 @@ impl Ufo {
...
+            for glyph_name in group_glyph_names {
+                if !kern2_set.insert(glyph_name) {
+                    return Err(Error::GroupsError);
+                }
+            }
+        }
+    }
+
+    Ok(())
+}
```
cmyr: I would suggest adding a separate `GroupsError` type, that might look something like,

```rust
enum GroupsError {
    InvalidName,
    DuplicateGlyph {
        glyph_name: String,
        group_name: String,
    },
    OverlappingGroups {
        overlapping_glyphs: Vec<String>,
    },
}
```

and then you can implement `Display` for this error in a way that conveys more specific information about the error.

src/glyph/serialize.rs
madig: `cargo fmt` does not change this, so uh?

src/glyph/serialize.rs
```diff
@@ -2,18 +2,18 @@
...
 use super::{
-    Advance, Anchor, Color, Component, Contour, ContourPoint, GlifVersion, Glyph, Guideline,
-    Identifier, Image, Line, PointType,
+    Advance, AffineTransform, Anchor, Color, Component, Contour, ContourPoint, GlifVersion, Glyph,
+    Guideline, Identifier, Image, Line, PointType,
 };
 
 use crate::error::GlifWriteError;
 
+use quick_xml::{
```
cmyr: import ordering is generally std > crates > internal stuff

src/glyph/serialize.rs
```diff
@@ -236,12 +238,8 @@ impl Image {
...
         let mut start = BytesStart::borrowed_name(b"image");
         start.push_attribute(("fileName", self.file_name.to_str().unwrap_or("missing path")));
-        start.push_attribute(("xScale", self.transform.x_scale.to_string().as_str()));
-        start.push_attribute(("yScale", self.transform.y_scale.to_string().as_str()));
-        start.push_attribute(("xyScale", self.transform.xy_scale.to_string().as_str()));
-        start.push_attribute(("yxScale", self.transform.yx_scale.to_string().as_str()));
-        start.push_attribute(("xOffset", self.transform.x_offset.to_string().as_str()));
-        start.push_attribute(("yOffset", self.transform.y_offset.to_string().as_str()));
+
+        transformation_to_attributes(&mut start, &self.transform);
```
cmyr: I would call this function something like `write_transform_attributes`; the `to` makes it sound like this is returning something?

Issue #31, opened 2020-02-19
UFO implementation checklist


madig:
What's missing until Norad is basically feature complete:

- [x] De/serialization of a glyph's `<lib>` element. The difficulty is making `plist` reading from and writing to an existing XML stream up until a certain point and the returning control.
- [x] Reading and writing a layer's layerinfo.plist. Layers have a `lib`, too.
- [x] Upgrading of UFO v1 and v2 formats on load according to specification recommendations. I don't think we need to support downgrading. 
  - [x] Upgrade fontinfo (and feature) data from v1 and v2.
  - [x] Upgrade GLIF v1, which handled anchors by informal convention (emulate what ufoLib does).
- [x] Support the `data` folder. It should lazily load any referenced data, as it could be megabytes of binary gunk. This probably implies a split between saving in-place (where you only write data objects that are loaded into memory and could therefore be changed) and saving into a fresh location/overwriting (where you load and then write out everything).
- [x] Support the `images` folder: http://unifiedfontobject.org/versions/ufo3/images/
- [x] Verify we have all validators from ufoLib on board. Todo: 
  - [x] identifiers! All identifiers _within a glyph_ and separately _within fontinfo's guideline list_ must be unique.
    - [x] glyph identifiers.
    - [x] fontinfo identifiers
  - [x] guideline validators, others?
  - [x] Point data combination validation?
  - [x] Glif reader validation (do we cover what e.g. ufoLib's `_validateAndMassagePointStructures` does?)
- [x] Track new additions to the spec 
   - [x] public objects and automatic pruning
   - [x] Enforce use of name Strings without control characters: https://unifiedfontobject.org/versions/ufo3/conventions/#controls (#191)
- [x] More complete user name to filename conversion (#85)

Very nice to have:

- [x] When serializing property lists like ~~fontinfo~~ and kerning, make an effort to write out integers when we have integers instead of `<real>` elements everywhere. The values are the same, but other tools make the same effort, and we want to keep down on diff noise.
- [x] Better error types/API so that e.g. loading a faulty UFO produces a legible error message? Need to check all possible errors.
    - [x] More detailed fontinfo errors
    - [x] If missing e.g. a glif file listed in contents.plist, note the name, not just the OsError.

Deferred to later:

- https://github.com/linebender/norad/issues/262
- https://github.com/linebender/norad/issues/263
- https://github.com/linebender/norad/issues/198
[...]

PR #32, opened 2020-03-04
Upconvert UFO v1 and v2 kerning data

madig:
Part of https://github.com/linebender/norad/issues/31.

- Spec: http://unifiedfontobject.org/versions/ufo3/kerning.plist/#converting-to-ufo-3-formatted-kerning
- Reference implementation: https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/ufoLib/converters.py

TODO:
- [x] Make it work, then make it not copy everything
- [x] Run kerning group validation after conversion
- [x] Implement all reference implementation tests (`Tests/ufoLib/UFOConversion_test.py`!)
- [x] Move into own utility/conversion module?
- [x] ~~What if the target kerning group exists already? Making the names unique won't save you from then potentially having glyphs in more than one kerning group. The reference implementation already fails here.~~ Error out
- [x] ~~What if either groups.plist or kerning.plist exist but not both?~~ The algorithm needs groups.plist at minimum, kerning.plist can be empty. Pass in empty container if kerning is `None`.

### DIFF
--- a/src/ufo.rs
+++ b/src/ufo.rs
@@ -28,6 +28,14 @@ static DEFAULT_LAYER_NAME: &str = "public.default";
 static DEFAULT_GLYPHS_DIRNAME: &str = "glyphs";
 static DEFAULT_METAINFO_CREATOR: &str = "org.linebender.norad";
 
+/// Groups is a map of group name to a list of glyph names. It's a BTreeMap because we need sorting
+/// for serialization.
+pub type Groups = BTreeMap<String, Vec<GlyphName>>;
+/// Kerning is a map of first half of a kerning pair (glyph name or group name) to the second half
+/// of a pair (glyph name or group name), which maps to the kerning value (high-level view:
+/// (first, second) => value). It's a BTreeMap because we need sorting for serialization.
+pub type Kerning = BTreeMap<String, BTreeMap<String, f32>>;
+
 /// A Unified Font Object.
 #[derive(Clone)]
 pub struct Ufo {
@@ -35,9 +43,8 @@ pub struct Ufo {
     pub font_info: Option<FontInfo>,
     pub layers: Vec<LayerInfo>,
     pub lib: Option<plist::Dictionary>,
-    // groups and kerning: BTreeMap because we need sorting for deserialization.
-    pub groups: Option<BTreeMap<String, Vec<String>>>,
-    pub kerning: Option<BTreeMap<String, BTreeMap<String, f32>>>,
+    pub groups: Option<Groups>,
+    pub kerning: Option<Kerning>,
     pub features: Option<String>,
     __non_exhaustive: (),
 }
@@ -147,9 +154,8 @@ impl Ufo {
 
             let groups_path = path.join(GROUPS_FILE);
             let groups = if groups_path.exists() {
-                let groups: BTreeMap<String, Vec<String>> = plist::from_file(groups_path)?;
+                let groups: Groups = plist::from_file(groups_path)?;
                 validate_groups(&groups).map_err(Error::GroupsError)?;
-
                 Some(groups)
             } else {
                 None
@@ -157,9 +163,7 @@ impl Ufo {
 
             let kerning_path = path.join(KERNING_FILE);
             let kerning = if kerning_path.exists() {
-                let kerning: BTreeMap<String, BTreeMap<String, f32>> =
-                    plist::from_file(kerning_path)?;
-
+                let kerning: Kerning = plist::from_file(kerning_path)?;
                 Some(kerning)
             } else {
                 None
@@ -168,7 +172,6 @@ impl Ufo {
             let features_path = path.join(FEATURES_FILE);
             let features = if features_path.exists() {
                 let features = fs::read_to_string(features_path)?;
-
                 Some(features)
             } else {
                 None
@@ -341,9 +344,7 @@ impl Ufo {
 
 /// Validate the contents of the groups.plist file according to the rules in the
 /// [Unified Font Object v3 specification for groups.plist](http://unifiedfontobject.org/versions/ufo3/groups.plist/#specification).
-fn validate_groups(
-    groups_map: &BTreeMap<String, Vec<String>>,
-) -> Result<(), GroupsValidationError> {
+fn validate_groups(groups_map: &Groups) -> Result<(), GroupsValidationError> {
     let mut kern1_set = HashSet::new();
     let mut kern2_set = HashSet::new();
     for (group_name, group_glyph_names) in groups_map {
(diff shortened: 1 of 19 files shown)


madig:
ufoLib fails itself when you up/downconvert UFOs from v2 -> v3 -> v2 and load the last UFO again. Since it leaves all groups intact, upconversion will hit duplicate group names and make new ones with the same glyph list. Validation will subsequently fail.

```
In [3]: u = defcon.Font("sourcesanspro_v2.ufo")
In [4]: u.save("/tmp/sourcesanspro_v3.ufo", formatVersion=3)
In [5]: u2 = defcon.Font("sourcesanspro_v3.ufo")
In [6]: u2.save("/tmp/sourcesanspro_v2_2.ufo", formatVersion=2)
In [7]: u3 = defcon.Font("/tmp/sourcesanspro_v2_2.ufo")
[...]
UFOLibError: The glyph "nu" occurs in too many kerning groups.
```

src/ufo.rs
```diff
@@ -28,16 +29,19 @@ static DEFAULT_LAYER_NAME: &str = "public.default";
 static DEFAULT_GLYPHS_DIRNAME: &str = "glyphs";
 static DEFAULT_METAINFO_CREATOR: &str = "org.linebender.norad";
 
+// groups and kerning: BTreeMap because we need sorting for serialization.
+pub type Groups = BTreeMap<String, Vec<String>>;
```
cmyr, line 33: I think it's worth being explicit about the structure of each map is.

src/ufo.rs
```diff
@@ -194,6 +193,39 @@ impl Ufo {
...
             let layers = layers?;
 
+            match meta.format_version {
+                FormatVersion::V1 | FormatVersion::V2 => {
+                    // Upconvert UFO v1 or v2 kerning if necessary. To upconvert, we need at least
+                    // a groups.plist file...
+                    if let Some(groups_to_upconvert) = &groups {
+                        // ... while a kerning.plist is optional. If there isn't one, pass an empty
+                        // map to the converter to keep the code simple.
+                        let kerning_to_upconvert;
```
cmyr, line 203: I would just write this as,

```rust
// maybe just 'let kerning = ..'?
let kerning_to_upconvert = kerning.unwrap_or_default();
```

(if default is unavailable you can do `kerning.unwrap_or_else(|| BTreeMap::new()`)

src/ufo.rs
```diff
@@ -194,6 +193,39 @@ impl Ufo {
...
+
+                        let (groups_new, kerning_new) = upconvert_kerning(
+                            &groups_to_upconvert,
+                            &kerning_to_upconvert,
+                            &glyph_names,
+                        );
+                        match validate_groups(&groups_new) {
+                            Err(Error::Groups(e)) => Err(Error::GroupsUpconversionError(e))?,
+                            _ => {}
+                        }
```
cmyr, line 219: Hmm, this is a bit funny.

First: if `validate_groups` only ever returns a `GroupsError`, it should return that directly as its error type, instead of wrapping it in `Error`; the caller can then convert to `Error` as needed.

The normal way to do this is to implement `From` for your inner error type to your outer error type; then the caller can just use `?` and it will convert the error types as needed.

That no longer makes sense here, since there are two different outer error types that use the same inner type, and so you'll have to be explicit. In any case, what I *think* this should look like is,

```rust
validate_groups(&groups_new).map_err(|e| Error::GroupsUpConversionError(e))?;
```

If you wanted to get _extra_ fancy, you could have some simple methods on`GroupsError` that would convert it to the different variants of the outer error; then you could use those in your `map_err`, for instance:

```rust
impl GroupsError {
    fn into_upconversion_error(self) -> Error {
        Error::GroupsUpConversionError(self)
    }
}

// and then
validate_groups(&groups_new).map_err(GroupsError::into_upconversion_error)?;
```

src/upconversion.rs
```diff
@@ -0,0 +1,252 @@
+use std::collections::{BTreeMap, HashMap, HashSet};
+
+use crate::glyph::GlyphName;
+use crate::ufo::{Groups, Kerning};
+
+pub fn upconvert_kerning(
```
cmyr, line 6: can we get a brief docstring here?


Reading through this I do feel that we're doing more cloning than we need to, but I also don't think it's *really* worth the hassle to optimize right now.

madig: I'd do it for cleaner code primarily :)

src/ufo.rs
madig: What do you mean?

cmyr: just a doc string I think? "A map of group names to the names of the glyphs in that group".

src/ufo.rs
```diff
@@ -194,6 +197,30 @@ impl Ufo {
                 .collect();
             let layers = layers?;
 
+            match meta.format_version {
```
cmyr, line 200: another little rust style thing, but is there a way we could write this that uses the match to avoid the `mut` annotations? For instance,

```rust
let (groups, kerning) = match (meta.format_version, groups, kerning) {
    // for v3, we do nothing
    (FormatVersion::V3, g, k) => (g, k),
    (_, None, k) => (None, k),
    (_, Some(g), k) => {
        let (groups, kerning) = upconvert_kerning(g, &k.unwrap_or_default());
        validate_groups(&groups)..maperr(Error::GroupsUpconversionError)?;
        (Some(groups), Some(kerning))
    }
}
```

src/upconversion.rs
```diff
@@ -0,0 +1,408 @@
...
+        let second_new = make_unique_group_name(
+            format!("public.kern2.{}", second.replace("@MMK_R_", "")),
+            &groups_new,
+        );
+        groups_second_old_to_new.insert(second, second_new.to_string());
+        groups_new.insert(second_new, groups_new.get(second).unwrap().clone());
+    }
+
+    // Update all kerning pairs that have an old kerning group in them with the new name.
+    let mut kerning_new: Kerning = BTreeMap::new();
```
cmyr, line 62: or just: `let mut kerning_new = Kerning::new()`;

src/ufo.rs
```diff
@@ -28,16 +29,23 @@ static DEFAULT_LAYER_NAME: &str = "public.default";
 static DEFAULT_GLYPHS_DIRNAME: &str = "glyphs";
 static DEFAULT_METAINFO_CREATOR: &str = "org.linebender.norad";
 
+/// Groups is a map of group name to a list of glyph names. It's a BTreeMap because we need sorting
+/// for serialization.
+pub type Groups = BTreeMap<String, Vec<String>>;
```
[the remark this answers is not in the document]
madig: Yes :smile:

PR #38, opened 2020-04-05
Upconvert UFO v1 and v2 fontinfo and feature data

madig:
This does a bit of work.

- UFO v1 stored feature code and the Postscript info attributes in lib.plist, which need to be sourced when upconverting: https://github.com/robotools/defcon/blob/a1e9d2852bb9d531ae5ec5b707fa0ab51aa202f1/Lib/defcon/objects/font.py#L1531-L1589
- UFO v1 had different names for the fontinfo attributes, some of which  need conversion: https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/ufoLib/__init__.py#L1934-L1962 (and possibly more code)
- UFO v2 and v3 have the same attribute names (v3 just has more), but v3 also updates the data types for some attributes (e.g. unitsPerEm are made positive)

ufoLib supports converting the versions up and down, but I think forcibly one-way upgrading the data to the latest supported version is the way to go for Norad.

The fontinfo error messages remain... rudimentary. I think they need to be revisited to return more useful error messages at a later point. I don't know how to get serde to include the field name in error messages if a field contains invalid data.

### DIFF
--- a/src/ufo.rs
+++ b/src/ufo.rs
@@ -132,10 +132,10 @@ impl Ufo {
         fn load_impl(path: &Path) -> Result<Ufo, Error> {
             let meta_path = path.join(METAINFO_FILE);
             let mut meta: MetaInfo = plist::from_file(meta_path)?;
+
             let fontinfo_path = path.join(FONTINFO_FILE);
-            let font_info = if fontinfo_path.exists() {
-                let font_info: FontInfo = plist::from_file(fontinfo_path)?;
-                font_info.validate()?;
+            let mut font_info = if fontinfo_path.exists() {
+                let font_info: FontInfo = FontInfo::from_file(fontinfo_path, meta.format_version)?;
                 Some(font_info)
             } else {
                 None
--- a/src/ufo.rs
+++ b/src/ufo.rs
@@ -15,7 +15,7 @@ use crate::fontinfo::FontInfo;
 use crate::glyph::{Glyph, GlyphName};
 use crate::layer::Layer;
 use crate::names::NameList;
-use crate::upconversion::upconvert_kerning;
+use crate::upconversion;
 use crate::Error;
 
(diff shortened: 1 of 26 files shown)


src/ufo.rs
```diff
@@ -205,12 +205,34 @@ impl Ufo {
...
+            if meta.format_version == FormatVersion::V1 {
+                let mut feature_text = String::new();
+                let mut fontinfo =
+                    if let Some(fontinfo) = font_info { fontinfo } else { FontInfo::default() };
+
+                if let Some(lib_data) = &mut lib {
+                    upconversion::upconvert_ufov1_robofab_data(
+                        &lib_path,
+                        lib_data,
+                        &mut feature_text,
```
cmyr, line 225: I don't think it makes sense to have this be an inout argument; I'd prefer to just return a tuple from this function.

madig:
This works, but strikes me as somehow unrusty.

src/ufo.rs
```diff
@@ -214,21 +214,20 @@ impl Ufo {
...
                         &lib_path,
                         lib_data,
-                        &mut feature_text,
                         &mut fontinfo,
                     )?;
                 }
 
-                if !feature_text.is_empty() {
-                    features = Some(feature_text);
+                if features_upgraded.is_some() && !features_upgraded.as_ref().unwrap().is_empty() {
```
cmyr, line 229: I would write this as something like (and this may not compile, I haven't actually tried it):

```suggestion
                if features_upgraded.as_ref().map(String::is_empty).unwrap_or(false) {
```

but basically doing a `map` to get the bool, and then using `unwrap_or` to choose what happens in the `None` case.

madig: Hrm. I had to read that multiple times. It also compiles but makes the tests fail, so it has to be `!features_upgraded.as_ref().map(String::is_empty).unwrap_or(true)`. Uh... I'm going to go ahead and merge as is...

PR #40, opened 2020-04-12
Fontinfo: Opportunistically convert IntegerOrFloat to int

madig:
Turn `IntegerOrFloat` from a type alias to a newtype.

One additional detail: `units_per_em` was a bare f64 before, is a `IntegerOrFloat` now. The v3 spec says it must be positive, which the `validate` continues to ensure. I think once the library has matured, _someone_ should think about how to do this more elegantly maybe, like ensuring that the data can never be negative by construction.

### DIFF
--- a/src/fontinfo.rs
+++ b/src/fontinfo.rs
@@ -1,6 +1,3 @@
-use std::convert::TryFrom;
-use std::ops::Deref;
-
 use serde::de::Deserializer;
 use serde::ser::{SerializeSeq, Serializer};
 use serde::{Deserialize, Serialize};
@@ -178,7 +175,9 @@ mod types {
     }
 }
 
-use types::{Integer, IntegerOrFloat, Float, NonNegativeInteger, NonNegativeIntegerOrFloat, Bitlist};
+use types::{
+    Bitlist, Float, Integer, IntegerOrFloat, NonNegativeInteger, NonNegativeIntegerOrFloat,
+};
 
 /// The contents of the [`fontinfo.plist`][] file. This structure is hard-wired to the
 /// available attributes in UFO version 3.
@@ -864,6 +863,7 @@ impl<'de> Deserialize<'de> for StyleMapStyle {
 
 #[cfg(test)]
 mod tests {
+    use std::convert::TryFrom;
     use super::*;
     use serde_test::{assert_tokens, Token};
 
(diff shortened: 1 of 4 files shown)


src/fontinfo.rs
```diff
@@ -11,10 +13,51 @@ use crate::Error;
...
+/// IntegerOrFloat represents a number that can be an integer or float. It should
+/// serialize to an integer if it effectively represents one.
+#[derive(Debug, Clone, PartialEq)]
+pub struct IntegerOrFloat(f64);
+
+impl IntegerOrFloat {
+    fn is_integer(&self) -> bool {
+        (self.0 - self.round()).abs() < std::f64::EPSILON
+    }
+}
```
cmyr, line 28: we're going to need some other methods here, like something for getting the actual value. Maybe just public `get` and `set` methods?

madig:
One thing I'd like to solve somehow is that if deserialization fails, it should tell which attribute was involved.

src/fontinfo.rs
```diff
@@ -5,15 +5,178 @@ use serde::{Deserialize, Serialize};
...
+        pub fn set(&mut self, value: f64) {
+            self.0 = value
+        }
+
+        pub fn is_integer(&self) -> bool {
+            (self.0 - self.round()).abs() < std::f64::EPSILON
+        }
+    }
+
+    impl Deref for IntegerOrFloat {
```
cmyr, line 48: Do you have a specific motivation for this?

src/fontinfo.rs
```diff
@@ -5,15 +5,178 @@ use serde::{Deserialize, Serialize};
...
-// in several locations and we have to assume a bigger type that can hold all sane
-// values.
-type Integer = i32;
-type NonNegativeInteger = u32;
-type IntegerOrFloat = f64;
-type Float = f64;
-type Bitlist = Vec<u8>;
+/// Fontinfo value types are put into a module so that code on this level cannot access the
+/// literal constructors.
+mod types {
```
cmyr, line 10: I might just put this in a separate file? That should give us the same guarantees.

src/fontinfo.rs
madig: nope. just read that this gives the newtype the same methods as the inner type which may be what we want?

src/fontinfo.rs
madig: Moved it into shared_types.

Maybe I need to re-export `*IntegerOrFloat`?

src/fontinfo.rs
madig: Makes it slightly less to type in some other lines of code :)

PR #41, opened 2020-04-17
Add postscript glyph name lookup

cmyr:
This uses a build.rs build script to generate a perfect hashmap for the glyph
names. This is maybe a bit over the top; we could have just put them
in order and done a binary search?

But we can always change this down the road if it feels like we're
definitely being too fancy.

Also: this isn't strictly related to UFO, and it's possible it should live elsewhere, or be behind a feature. I don't have strong opinions about this.


madig:
(I'd put it into Runebender.)

cmyr:
yea that maybe makes more sense

PR #48, opened 2020-05-15
Fix default file name guessing

cmyr:
It's perhaps arguable that this should even be part of norad? I think
it makes the most sense here.

This still likely needs some work; in particular we might want to
add in an ability to handle duplicate names.


madig:
Guessing the file name from a glyph name can be part of norad, it's also part of ufoLib.

PR #51, opened 2020-06-21
Implement Glif v1 upconversion and minor fixups

madig:
This PR implements:
- A few miscellaneous fixes to make output more diff-less compared to the reference libraries.
- Ensure a glifs' note and lib elements are present once and that a v1 contour has no attributes.
- Loading and upconverting Glif v1 files, with single-point move-type contours being upgraded to anchors, as recommended in http://unifiedfontobject.org/versions/ufo3/glyphs/glif/#converting-implied-anchors-in-glif-1-to-glif-2-anchor-elements.

### DIFF
--- a/src/glyph/parse.rs
+++ b/src/glyph/parse.rs
@@ -249,6 +249,10 @@ impl<'names> GlifParser<'names> {
         reader: &Reader<&[u8]>,
         data: BytesStart<'a>,
     ) -> Result<(), Error> {
+        if self.glyph.advance.is_some() {
+            return Err(err!(reader, ErrorKind::UnexpectedDuplicate));
+        }
+
         let mut advance = Advance::default();
         for attr in data.attributes() {
             let attr = attr?;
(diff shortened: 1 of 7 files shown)


src/glyph/parse.rs
```diff
@@ -122,7 +135,32 @@ impl<'names> GlifParser<'names> {
...
+            let anchor_point = points.remove(0);
+            let anchor = Anchor {
+                name: anchor_point.name,
+                x: anchor_point.x,
+                y: anchor_point.y,
+                identifier: None,
+                color: None,
+            };
+            // Create anchors vector unless already present.
+            if let Some(anchors) = self.glyph.anchors.as_mut() {
```
cmyr, line 155: I think this could just be,

```rust
self.glyph.anchors.get_or_insert(Vec::new()).push(anchor);
```

src/glyph/parse.rs
```diff
@@ -122,7 +135,32 @@ impl<'names> GlifParser<'names> {
                 _other => return Err(err!(reader, ErrorKind::UnexpectedElement)),
             }
         }
-        self.glyph.outline.as_mut().unwrap().contours.push(Contour { identifier, points });
+
+        // In the Glif v1 spec, single-point contours that have a "move" type and a name should
+        // be treated as anchors and upgraded.
+        if self.glyph.format == GlifVersion::V1
```
cmyr, line 141: I might replace this with a standalone function,

```rust
fn is_v1_anchor(format: GlifVersion, points: &[Point]) -> bool;
```

PR #53, opened 2020-09-09
Add UfoDataRequest

ctrlcctrlv:
This makes it so I can do this in MFEQ's Qmetadata:

```rust
fn main() {
    ..
    let ufo = Ufo::with_fields(UfoDataRequest::from_bool(false)).load_ufo(matches.value_of("UFO").unwrap());
    ..
}
```

That MFEQ module will never need all the glyph and kerning parsing this took does; so all it does is slow things down substantially, especially with the 125MB UFO for [TT2020 Style F](https://fontlibrary.org/fr/font/tt2020-style-f). I know Rust maintainers are rather picky, so if you don't like this idea, just let me know, I'm happy to maintain my own Norad version, I already maintain a personal `.glif` parser because I need Skia-friendly points.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "norad"
-version = "0.2.0"
+version = "0.2.1"
 authors = ["Colin Rofls <colin@cmyr.net>", "Nikolaus Waxweiler <madigens@gmail.com>"]
 license = "MIT/Apache-2.0"
 edition = "2018"
--- a/src/ufo.rs
+++ b/src/ufo.rs
@@ -47,6 +47,7 @@ pub struct Ufo {
     pub groups: Option<Groups>,
     pub kerning: Option<Kerning>,
     pub features: Option<String>,
+    data_request: UfoDataRequest,
     __non_exhaustive: (),
 }
 
@@ -66,8 +67,40 @@ impl Default for Ufo {
             groups: None,
             kerning: None,
             features: None,
+            data_request: Default::default(),
             __non_exhaustive: (),
         }
+    }
+}
+
+/// Which UFO data do you want?
+#[derive(Clone, Debug, PartialEq)]
+pub struct UfoDataRequest {
+    pub layers: bool,
+    pub lib: bool,
+    pub groups: bool,
+    pub kerning: bool,
+    pub features: bool,
+    __non_exhaustive: (),
+}
+
+impl UfoDataRequest {
+    /// Convenience function to create a UfoDataRequest with all fields either true or false.
+    pub fn from_bool(b: bool) -> Self {
+        UfoDataRequest {
+            layers: b,
+            lib: b,
+            groups: b,
+            kerning: b,
+            features: b,
+            __non_exhaustive: (),
+        }
+    }
+}
+
+impl Default for UfoDataRequest {
+    fn default() -> Self {
+        UfoDataRequest::from_bool(true)
     }
 }
 
@@ -114,9 +147,16 @@ impl Default for MetaInfo {
 }
 
 impl Ufo {
-    /// Crate a new `Ufo`.
+    /// Create a new `Ufo`.
     pub fn new() -> Self {
         Ufo::default()
+    }
+
+    /// Create a new `Ufo` only with certain fields
+    pub fn with_fields(data_request: UfoDataRequest) -> Self {
+        let mut ufo = Self::new();
+        ufo.data_request = data_request;
+        ufo
     }
 
     /// Attempt to load a font object from a file. `path` must point to
@@ -125,11 +165,14 @@ impl Ufo {
     ///
     /// [v3]: http://unifiedfontobject.org/versions/ufo3/
     pub fn load<P: AsRef<Path>>(path: P) -> Result<Ufo, Error> {
+        Self::new().load_ufo(path)
+    }
+
+    pub fn load_ufo<P: AsRef<Path>>(&self, path: P) -> Result<Ufo, Error> {
         let path = path.as_ref();
-        return load_impl(path);
 
         // minimize monomorphization
-        fn load_impl(path: &Path) -> Result<Ufo, Error> {
+        let load_impl = |ufo: &Ufo, path: &Path| -> Result<Ufo, Error> {
             let meta_path = path.join(METAINFO_FILE);
             let mut meta: MetaInfo = plist::from_file(meta_path)?;
 
@@ -142,7 +185,7 @@ impl Ufo {
             };
 
             let lib_path = path.join(LIB_FILE);
-            let mut lib = if lib_path.exists() {
+            let mut lib = if lib_path.exists() && self.data_request.lib {
                 // Value::as_dictionary(_mut) will only borrow the data, but we want to own it.
                 // https://github.com/ebarnard/rust-plist/pull/48
                 match plist::Value::from_file(&lib_path)? {
@@ -154,7 +197,7 @@ impl Ufo {
             };
 
             let groups_path = path.join(GROUPS_FILE);
-            let groups = if groups_path.exists() {
+            let groups = if groups_path.exists() && self.data_request.groups {
                 let groups: Groups = plist::from_file(groups_path)?;
                 validate_groups(&groups).map_err(Error::GroupsError)?;
                 Some(groups)
@@ -163,7 +206,7 @@ impl Ufo {
             };
 
             let kerning_path = path.join(KERNING_FILE);
-            let kerning = if kerning_path.exists() {
+            let kerning = if kerning_path.exists() && self.data_request.kerning {
                 let kerning: Kerning = plist::from_file(kerning_path)?;
                 Some(kerning)
             } else {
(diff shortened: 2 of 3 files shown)


madig:
I don't understand what this does from the PR text or the code :O please elaborate.

ctrlcctrlv:
@madig I don't understand why you don't understand 😓 

If you call `Ufo::load("someufo.ufo")`, norad will parse all the glyphs and allocate memory to hold their data.

If you call my above example, it will _only_ parse the font's `fontinfo.plist`.

madig:
Aha! Now I understand the purpose, though I still don't know how the code above arrives at the fontinfo.plist data. It will be some time until I can study this. Maybe @cmyr has a quick take...

ctrlcctrlv:
Right here:

https://github.com/linebender/norad/blob/99b565ede6320efa6e686a70ff8be0a3efbe2ef6/src/ufo.rs#L181

This part has not changed. All I did was not fetch fields of the `Ufo` which the program has said it never plans to read by setting them `false` in the `UfoDataRequest`.

src/ufo.rs
```diff
@@ -66,11 +67,43 @@ impl Default for Ufo {
...
 
+/// Which UFO data do you want?
+#[derive(Clone, Debug, PartialEq)]
+pub struct UfoDataRequest {
+    pub layers: bool,
+    pub lib: bool,
+    pub groups: bool,
+    pub kerning: bool,
+    pub features: bool,
+    __non_exhaustive: (),
```
cmyr, line 84: there's now a `#[non_exhaustive]` annotation we can use on the struct itself.

src/ufo.rs
```diff
@@ -114,22 +147,32 @@ impl Default for MetaInfo {
 }
 
 impl Ufo {
-    /// Crate a new `Ufo`.
+    /// Create a new `Ufo`.
```
cmyr, line 150: 😅

src/ufo.rs
```diff
@@ -66,11 +67,43 @@ impl Default for Ufo {
...
             kerning: None,
             features: None,
+            data_request: Default::default(),
             __non_exhaustive: (),
         }
     }
 }
 
+/// Which UFO data do you want?
+#[derive(Clone, Debug, PartialEq)]
```
cmyr, line 77: Might as well make this `copy`?

src/ufo.rs
```diff
@@ -66,11 +67,43 @@ impl Default for Ufo {
...
+        UfoDataRequest {
+            layers: b,
+            lib: b,
+            groups: b,
+            kerning: b,
+            features: b,
+            __non_exhaustive: (),
+        }
+    }
+}
```
cmyr, line 99: thinking about this purely in terms of API: instead of `from_bool` I might just have two fns, `all` or `none`? Alternatively I might consider a builder pattern, where you do like,

```rust
let request = UfoDataRequest::new().layers(true).features(true);
```

or similar.

src/ufo.rs
```diff
@@ -66,11 +67,43 @@ impl Default for Ufo {
             groups: None,
             kerning: None,
             features: None,
+            data_request: Default::default(),
             __non_exhaustive: (),
         }
     }
 }
 
+/// Which UFO data do you want?
```
cmyr, line 76: I would be more explicit here; something like,

```rust
/// A type that describes which components of a UFO should be loaded.
///
/// By default, we load all components of the UFO file; however if you only
/// need some subset of these, you can pass this struct to [`Ufo::with_fields`]
/// in order to only load the fields specified in this object.
///
/// [`Ufo::with_fields`]: struct.Ufo.html#method.with_fields
```

src/ufo.rs
```diff
@@ -114,22 +147,32 @@ impl Default for MetaInfo {
...
 
 impl Ufo {
-    /// Crate a new `Ufo`.
+    /// Create a new `Ufo`.
     pub fn new() -> Self {
         Ufo::default()
     }
 
+    /// Create a new `Ufo` only with certain fields
+    pub fn with_fields(data_request: UfoDataRequest) -> Self {
```
cmyr, line 156: I might just have a `set_data_request` method? Alternatively there could be a builder-style method, like:

```rust
impl Ufo {
    // fn new etc
    pub fn with_data_request(mut self, data_request: UfoDataRequest) -> Self {
        self.data_request = data_request;
        self
    }
}

// and we use this like,

let my_ufo = Ufo::new().with_data_request(UfoDataRequest::default());
```

src/ufo.rs
```diff
@@ -66,11 +67,43 @@ impl Default for Ufo {
...
             features: None,
+            data_request: Default::default(),
             __non_exhaustive: (),
         }
     }
 }
 
+/// Which UFO data do you want?
+#[derive(Clone, Debug, PartialEq)]
+pub struct UfoDataRequest {
```
cmyr, line 78: I'd just call this `DataRequest`, the `Ufo` bit is implied I think. 🤷

cmyr (changes requested):
I think this is a reasonable addition? We may want to revisit at some point in the future, but this seems pretty straight-forward.

I have some naming and API suggestions, but overall I have no problem with this being added. Thanks!

madig:
This suddenly reminds me of `ufoLib`'s API. See all the `read*` methods for https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/ufoLib/__init__.py#L197. Might be useful for making a rustified Python `ufoLib` API.

src/ufo.rs
```diff
@@ -66,11 +67,73 @@ impl Default for Ufo {
...
+            data_request: Default::default(),
         }
     }
 }
 
+/// A type that describes which components of a UFO should be loaded.
+///
+/// By default, we load all components of the UFO file; however if you only
+/// need some subset of these, you can pass this struct to [`Ufo::with_fields`]
+/// in order to only load the fields specified in this object.
```
cmyr, line 79: great, thanks 👌

src/ufo.rs
```diff
@@ -114,22 +177,32 @@ impl Default for MetaInfo {
...
     /// Attempt to load a font object from a file. `path` must point to
     /// a directory with the structure described in [v3 of the Unified Font Object][v3]
     /// spec.
     ///
     /// [v3]: http://unifiedfontobject.org/versions/ufo3/
     pub fn load<P: AsRef<Path>>(path: P) -> Result<Ufo, Error> {
+        Self::new().load_ufo(path)
+    }
+
+    pub fn load_ufo<P: AsRef<Path>>(&self, path: P) -> Result<Ufo, Error> {
```
cmyr, line 201: I'm feeling a bit thick, can you help me understand this change? it seems like we now have two differently named functions that do the same thing?

In any case, if we're adding new public API I'd like at least a short doc comment. :)

src/ufo.rs
```diff
@@ -163,40 +236,45 @@ impl Ufo {
...
+                    .drain(..)
+                    .map(|(name, p)| {
+                        let layer_path = path.join(&p);
+                        let layer = Layer::load_impl(&layer_path, &glyph_names)?;
+                        Ok(LayerInfo { name, path: p, layer })
+                    })
+                    .collect();
+                layers = layers_r?;
+            } else {
+                layers = Vec::new();
```
cmyr, line 276: The idiomatic way to do this would be not to assign in each block, but to use the fact that blocks in rust are all expressions; that is, I would do:

```rust
let layers = if self.data_request.layers {
    // load layers
} else {
    Vec::new()
}
```

no big deal of course but nice to be consistent.

src/ufo.rs
```diff
@@ -66,11 +67,73 @@ impl Default for Ufo {
...
+    pub kerning: bool,
+    pub features: bool,
+}
+
+impl DataRequest {
+    fn from_bool(b: bool) -> Self {
+        DataRequest { layers: b, lib: b, groups: b, kerning: b, features: b }
+    }
+
+    pub fn all() -> Self {
```
cmyr, line 97: Thanks for this, I think it's a nice way of doing the API. It would be nice if these public items had a short doc comment, though?

cmyr (commented):
Okay a few more little notes but this is basically good to go I think.

src/ufo.rs
ctrlcctrlv: It assures that old code works unchanged, as we previously did not require Ufo type be initialized before calling `load()`. So, that is maintained, with us now doing the initialization so API consumers won't notice anything has changed.

Cargo.toml
```diff
@@ -1,12 +1,12 @@
...
-version = "0.2.0"
+version = "0.2.1"
 authors = ["Colin Rofls <colin@cmyr.net>", "Nikolaus Waxweiler <madigens@gmail.com>"]
 license = "MIT/Apache-2.0"
 edition = "2018"
 keywords = ["font", "ufo", "fonts"]
-repository = "https://github.com/linebender/norad"
-description = "Read and write Unifed Font Object files."
+repository = "https://github.com/mfeq/norad"
+description = "Read and write Unifed Font Object files. (MFEQ fork)"
```
madig, line 9: Not yet!

ctrlcctrlv: Sorry, it seems that I mixed up the branches. This was unintentional.

src/ufo.rs
ctrlcctrlv: Following up. :-)

cmyr: makes sense!

cmyr:
@ctrlcctrlv what's the status of this? It felt like we were very close to merging, would still be happy to get it in...

ctrlcctrlv:
@cmyr Hello four months later. Sorry, I moved in November 2020, and have been working on a lot of different stuff. But finally found the time to advance this, as MFEK project (it's K now, not Q, the core team and I decided to change it) is taking up more of my time again. Please see 73840a9.

cmyr:
No worries, happy to take a look whenever it feels ready to for you :)

ctrlcctrlv:
I'd say it's ready. The failing test isn't my doing, the service seems down.

cmyr (approved):
Thanks!

PR #54, opened 2020-09-26
Consolidate Color methods to shared_types

madig:
I noticed that there is a `Color::from_str` already, but the one I wrote for serde is more anal.

### DIFF
--- a/src/shared_types.rs
+++ b/src/shared_types.rs
@@ -1,5 +1,6 @@
 use std::convert::TryFrom;
 use std::ops::Deref;
+use std::str::FromStr;
 
 use serde::de;
 use serde::de::Deserializer;
@@ -10,6 +11,7 @@ use serde::{Deserialize, Serialize};
 #[cfg(feature = "druid")]
 use druid::Data;
 
+use crate::error::ErrorKind;
 use crate::Error;
 
 /// Identifiers are optional attributes of several objects in the UFO.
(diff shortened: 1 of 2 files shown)


src/shared_types.rs
```diff
@@ -56,6 +58,51 @@ pub struct Color {
     pub alpha: f32,
 }
 
+impl FromStr for Color {
+    type Err = ErrorKind;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let colors: Vec<f32> = s.split(',').map(|v| v.parse().unwrap()).collect();
```
cmyr, line 65: I think it's worth skipping this allocation? I'm not sure if you can `collect` to a fixed-size array, but that might be worth it, or you could work with the iterator directly. Here's a quick sketch of how I might write this:

```rust
let mut iter = s.split(',').map(|v| match v.parse::<f32>() {
    Ok(val) if (0.0..=1.0).contains(val) => Ok(val),
    _ => ErrorKind::BadColor,
});
let red = iter.next()?;
let green = iter.next()?;
let blue = iter.next()?;
let alpha = iter.next()?;
if iter.next().is_some() {
    Err(ErrorKind::BadColor)
} else {
    Ok(Color { red, green, blue, alpha })
}
```

cmyr (approved):
This looks good, but I might rewrite the parsing code to skip the allocation?

cmyr (approved):
👍

PR #56, opened 2020-10-03
Add sanity check for outlines, add point pen

madig:
As detailed on https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#point.

I'm not quite sure about off-curves follows by a `move`, so I opened https://github.com/unified-font-object/ufo-spec/issues/148.

Also added a GlyphPointPen for programmatically drawing glyphs, like https://fonttools.readthedocs.io/en/latest/pens/pointPen.html.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -79,9 +79,23 @@ pub enum ErrorKind {
     BadImage,
     BadIdentifier,
     UnexpectedDuplicate,
+    UnexpectedMove,
+    UnexpectedSmooth,
     UnexpectedElement,
     UnexpectedAttribute,
     UnexpectedEof,
+    UnexpectedPointAfterOffCurve,
+    TooManyOffCurves,
+    PenPathNotStarted,
+    TrailingOffCurves,
+    DuplicateIdentifier,
+    UnexpectedDrawing,
+    UnfinishedDrawing,
+    UnexpectedPointField,
+    UnexpectedComponentField,
+    UnexpectedAnchorField,
+    UnexpectedGuidelineField,
+    UnexpectedImageField,
 }
 
 impl std::fmt::Display for Error {
@@ -139,9 +153,35 @@ impl std::fmt::Display for ErrorKind {
             ErrorKind::BadImage => write!(f, "Bad image"),
             ErrorKind::BadIdentifier => write!(f, "Bad identifier"),
             ErrorKind::UnexpectedDuplicate => write!(f, "Unexpected duplicate"),
+            ErrorKind::UnexpectedMove => {
+                write!(f, "Unexpected move point, can only occur at start of contour")
+            }
+            ErrorKind::UnexpectedSmooth => {
+                write!(f, "Unexpected smooth attribute on an off-curve point")
+            }
             ErrorKind::UnexpectedElement => write!(f, "Unexpected element"),
             ErrorKind::UnexpectedAttribute => write!(f, "Unexpected attribute"),
             ErrorKind::UnexpectedEof => write!(f, "Unexpected EOF"),
+            ErrorKind::UnexpectedPointAfterOffCurve => {
+                write!(f, "An off-curve point must be followed by a curve or qcurve")
+            }
+            ErrorKind::TooManyOffCurves => {
+                write!(f, "At most two off-curve points can precede a curve")
+            }
+            ErrorKind::PenPathNotStarted => {
+                write!(f, "Must call begin_path() before calling add_point() or end_path()")
+            }
+            ErrorKind::TrailingOffCurves => {
+                write!(f, "Open contours must not have trailing off-curves")
+            }
+            ErrorKind::DuplicateIdentifier => write!(f, "Duplicate identifier"),
+            ErrorKind::UnexpectedDrawing => write!(f, "Unexpected drawing without an outline"),
+            ErrorKind::UnfinishedDrawing => write!(f, "Unfinished drawing, you must call end_path"),
+            ErrorKind::UnexpectedPointField => write!(f, "Unexpected point field"),
+            ErrorKind::UnexpectedComponentField => write!(f, "Unexpected component field "),
+            ErrorKind::UnexpectedAnchorField => write!(f, "Unexpected anchor field "),
+            ErrorKind::UnexpectedGuidelineField => write!(f, "Unexpected guideline field "),
+            ErrorKind::UnexpectedImageField => write!(f, "Unexpected image field "),
         }
     }
 }
--- a/src/glyph/parse.rs
+++ b/src/glyph/parse.rs
@@ -101,7 +101,7 @@ impl<'names> GlifParser<'names> {
     ) -> Result<(), Error> {
         let mut identifier = None;
         for attr in data.attributes() {
-            if self.pen.format() == &GlifVersion::V1 {
+            if self.pen.get_format() == &GlifVersion::V1 {
                 return Err(err!(reader, ErrorKind::UnexpectedAttribute));
             }
             let attr = attr?;
--- a/src/glyph/parse.rs
+++ b/src/glyph/parse.rs
@@ -95,7 +95,7 @@ impl<'names> GlifParser<'names> {
                     match tag_name.borrow() {
                         "contour" => self.parse_contour(start, reader, &mut new_buf)?,
                         "component" => self.parse_component(reader, start)?,
-                        _other => eprintln!("unexpected tag in outline {}", tag_name),
+                        _other => return Err(err!(reader, ErrorKind::UnexpectedTag)),
                     }
                 }
                 Event::End(ref end) if end.name() == b"outline" => break,
@@ -118,9 +118,12 @@ impl<'names> GlifParser<'names> {
                 return Err(err!(reader, ErrorKind::UnexpectedAttribute));
             }
             let attr = attr?;
-            if attr.key == b"identifier" {
-                let ident = attr.unescape_and_decode_value(reader)?;
-                identifier = Some(Identifier::new(ident).map_err(|kind| err!(reader, kind))?);
+            match attr.key {
+                b"identifier" => {
+                    let ident = attr.unescape_and_decode_value(reader)?;
+                    identifier = Some(Identifier::new(ident).map_err(|kind| err!(reader, kind))?);
+                }
+                _other => return Err(err!(reader, ErrorKind::UnexpectedAttribute)),
             }
         }
 
@@ -190,7 +193,7 @@ impl<'names> GlifParser<'names> {
                 b"identifier" => {
                     identifier = Some(value.parse().map_err(|kind| err!(reader, kind))?);
                 }
-                _other => eprintln!("unexpected component field {}", value),
+                _other => return Err(err!(reader, ErrorKind::UnexpectedComponentField)),
             }
         }
 
@@ -270,7 +273,7 @@ impl<'names> GlifParser<'names> {
                 b"identifier" => {
                     identifier = Some(value.parse().map_err(|kind| err!(reader, kind))?);
                 }
-                _other => eprintln!("unexpected point field {}", String::from_utf8_lossy(_other)),
+                _other => return Err(err!(reader, ErrorKind::UnexpectedPointField)),
             }
         }
         if x.is_none() || y.is_none() {
@@ -355,7 +358,7 @@ impl<'names> GlifParser<'names> {
                 b"identifier" => {
                     identifier = Some(value.parse().map_err(|kind| err!(reader, kind))?);
                 }
-                _other => eprintln!("unexpected anchor field {}", value),
+                _other => return Err(err!(reader, ErrorKind::UnexpectedAnchorField)),
             }
         }
 
@@ -399,7 +402,7 @@ impl<'names> GlifParser<'names> {
                 b"identifier" => {
                     identifier = Some(value.parse().map_err(|kind| err!(reader, kind))?);
                 }
-                _other => eprintln!("unexpected guideline field {}", value),
+                _other => return Err(err!(reader, ErrorKind::UnexpectedGuidelineField)),
             }
         }
 
@@ -444,7 +447,7 @@ impl<'names> GlifParser<'names> {
                 b"yOffset" => transform.y_offset = value.parse().map_err(|_| (kind, pos))?,
                 b"color" => color = Some(value.parse().map_err(|e: ErrorKind| e.to_error(pos))?),
                 b"fileName" => filename = Some(PathBuf::from(value.to_string())),
-                _other => eprintln!("unexpected image field {}", value),
+                _other => return Err(err!(reader, ErrorKind::UnexpectedImageField)),
             }
         }
 
--- a/src/glyph/parse.rs
+++ b/src/glyph/parse.rs
@@ -88,11 +88,20 @@ impl<'names> GlifParser<'names> {
 
         loop {
             match reader.read_event(buf)? {
-                Event::Start(start) | Event::Empty(start) => {
+                Event::Start(start) => {
                     let tag_name = reader.decode(&start.name())?;
                     let mut new_buf = Vec::new(); // borrowck :/
                     match tag_name.borrow() {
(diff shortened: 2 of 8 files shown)


src/error.rs
```diff
@@ -137,9 +142,24 @@ impl std::fmt::Display for ErrorKind {
             ErrorKind::BadComponent => write!(f, "Bad component"),
             ErrorKind::BadImage => write!(f, "Bad image"),
             ErrorKind::UnexpectedDuplicate => write!(f, "Unexpected duplicate"),
+            ErrorKind::UnexpectedMove => {
+                write!(f, "Unexpected move point, can only occur at start of contour")
+            }
+            ErrorKind::UnexpectedSmooth => {
+                write!(f, "Unexpected smooth attribute on an off-curve point")
```
anthrotype, line 149: this looks to me be a bit too harsh. Can't you simply ignore that "smooth" flag on off-curve points?

madig: But the spec says it's illegal :O

anthrotype: oh does it? Well.. ok

src/glyph/parse.rs
```diff
@@ -123,18 +123,66 @@ impl<'names> GlifParser<'names> {
...
                 Event::End(ref end) if end.name() == b"contour" => break,
                 Event::Empty(ref start) if start.name() == b"point" => {
                     let point = self.parse_point(reader, start)?;
+                    match point.typ {
+                        PointType::Move => {
+                            if !points.is_empty() {
+                                return Err(err!(reader, ErrorKind::UnexpectedMove));
+                            }
+                            contour_closed = false;
+                        }
```
cmyr, line 146: I would include the condition in the match arm.
```suggestion
                        PointType::Move if !points.is_empty() => return Err(err!(reader, ErrorKind::UnexpectedMove)),
```

src/glyph/parse.rs
```diff
@@ -123,18 +123,66 @@ impl<'names> GlifParser<'names> {
...
 
+        // UFO contour sanity constraints:
+        // 1. If a "move" occurs, it must be the first point of a contour. ufoLib
+        //    may allow this for format 1, we don't.
+        // 2. The "smooth" attribute must not be set on off-curve points.
+        // 3. An off-curve point must be followed by a curve or qcurve.
+        // 4. A maximum of two offcurves can precede a curve.
         let mut points = Vec::new();
+        let mut number_of_offcurves: u32 = 0;
+        let mut contour_closed: bool = true;
```
cmyr, line 134: instead of maintaining a flag for this, I would just wait until we have all the points and then do,

```rust
let is_closed = points.first().map(|p| p.typ) == Some(PointType::Move);
```

src/glyph/parse.rs
```diff
@@ -123,18 +123,66 @@ impl<'names> GlifParser<'names> {
...
                 Event::Eof => return Err(err!(reader, ErrorKind::UnexpectedEof)),
                 _other => return Err(err!(reader, ErrorKind::UnexpectedElement)),
             }
         }
+        // If we are a closed contour and ended with off-curves, loop around and continue checking from the start.
+        if contour_closed && number_of_offcurves > 0 {
+            for point in &points {
+                match point.typ {
+                    PointType::OffCurve => number_of_offcurves += 1,
+                    PointType::QCurve => break,
```
cmyr: What if we have two off_curves here?

src/glyph/parse.rs
```diff
@@ -156,6 +204,50 @@ impl<'names> GlifParser<'names> {
         Ok(())
     }
 
+    fn parse_point<'a>(
```
cmyr, line 211: was this just moved, or was the code changed?

As patches get bigger it's nice sometimes to avoid making non-functional changes (like moving code) since it adds to the diff and can make the patch harder to reason about; just something I try to keep in mind.

src/glyph/pen.rs
```diff
@@ -0,0 +1,371 @@
...
+            None => Err(ErrorKind::PenPathNotStarted),
+        }
+    }
+
+    fn end_path(&mut self) -> Result<(), ErrorKind> {
+        // If we are a closed contour and ended with off-curves, loop around and continue checking from the start.
+        let contour = self.contour.take();
+        match contour {
+            Some(c) => {
+                let contour_closed = c.points.first().map_or(true, |v| v.typ != PointType::Move);
```
cmyr, line 104: might be worth just adding an `is_closed()` method to `Contour` 🤷

cmyr (commented):
This looks reasonable to me, a few comments inline!

src/glyph/parse.rs
madig: I just moved it! Ok, I can do a clean-up run after I fleshed this out. See it as a draft.

src/glyph/parse.rs
madig: I'll try to use the Pen thing instead! Maybe that works.

madig:
There are a bunch of `eprintln` in use in the parser code. Shouldn't they be the usual ErrorKinds instead?

src/glyph/parse.rs
```diff
@@ -348,15 +328,16 @@ impl<'names> GlifParser<'names> {
                     color = Some(value.parse().map_err(|_| err!(reader, ErrorKind::BadColor))?)
                 }
                 b"identifier" => identifier = Some(Identifier(value.to_string())),
-                _other => eprintln!("unexpected anchor field {}", value),
+                _other => return Err(err!(reader, ErrorKind::UnexpectedAnchorField)),
```
cmyr, line 331: yep, this is reasonable. I was probably just lazy.

src/glyph/parse.rs
```diff
@@ -460,40 +438,31 @@ fn start(reader: &mut Reader<&[u8]>, buf: &mut Vec<u8>) -> Result<Glyph, Error>
...
-                        name = attr.unescape_and_decode_value(&reader)?;
-                    } else if attr.key == b"format" {
-                        let value = attr.unescaped_value()?;
-                        let value = reader.decode(&value)?;
-                        format = Some(
-                            value
-                                .parse()
-                                .map_err(|e: ErrorKind| e.to_error(reader.buffer_position()))?,
-                        );
+                    // XXX: support `formatMinor`
```
cmyr, line 441: what does this mean?

src/glyph/pen.rs
```diff
@@ -0,0 +1,532 @@
...
+
+use crate::error::ErrorKind;
+use crate::glyph::{
+    Advance, AffineTransform, Anchor, Component, Contour, ContourPoint, GlifVersion, Glyph,
+    GlyphName, Guideline, Image, Outline, Plist, PointType,
+};
+use crate::shared_types::Identifier;
+
+#[derive(Debug)]
+pub struct Pen {
```
cmyr, line 13: I want docs! What is this, when should you use it, etc. Does this even need to be public?

src/glyph/pen.rs
```diff
@@ -0,0 +1,532 @@
...
+    glyph: Glyph,
+    height: Option<f32>,
+    width: Option<f32>,
+    identifiers: HashSet<u64>, // All identifiers within a glyph must be unique.
+    scratch_contour: Option<Contour>,
+    number_of_offcurves: u32,
+}
+
+impl Pen {
+    pub fn new(name: GlyphName, format: GlifVersion) -> Self {
```
cmyr, line 23: docs on all these `pub` fns please. 👼

src/glyph/pen.rs
```diff
@@ -0,0 +1,532 @@
...
+                width: self.width.unwrap_or(0.0),
+                height: self.height.unwrap_or(0.0),
+            })
+        }
+
+        self.glyph.format = GlifVersion::V2;
+        Ok(self.glyph)
+    }
+
+    // XXX: error when beginning a path without having ended the last?
```
cmyr, line 137: I agree that we probably should, unless there's some reason its difficult?

src/glyph/pen.rs
```diff
@@ -0,0 +1,532 @@
...
+                if Self::contour_is_v1_anchor(&self.glyph.format, &c.points) {
+                    let anchor_point = c.points.remove(0);
+                    let anchor = Anchor {
+                        name: anchor_point.name,
+                        x: anchor_point.x,
+                        y: anchor_point.y,
+                        identifier: None,
+                        color: None,
+                    };
+                    self.glyph.anchors.get_or_insert(Vec::new()).push(anchor);
```
cmyr, line 228: I would just write this as `vec![anchor]`.

src/glyph/pen.rs
```diff
@@ -0,0 +1,532 @@
...
+                    Ok(self)
+                } else {
+                    if self.number_of_offcurves > 0 {
+                        if c.is_closed() {
+                            for point in &c.points {
+                                match point.typ {
+                                    PointType::OffCurve => self.number_of_offcurves += 1,
+                                    PointType::QCurve => break,
```
cmyr, line 236: should we not also check for the number of offcurves?

src/lib.rs
```diff
@@ -30,7 +30,7 @@ mod upconversion;
 
 pub use error::Error;
 pub use fontinfo::FontInfo;
-pub use glyph::{Glyph, GlyphName};
+pub use glyph::{pen::Pen, Glyph, GlyphName};
```
cmyr, line 33: Is this something that consumers of norad need to access directly?

cmyr (commented):
Honestly this all looks great to me! A few questions and comments inline, but no major concerns.

src/glyph/parse.rs
madig: A new `formatMinor` attribute was added to the spec: https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#glyph. SO we can have UFO v3.1 at some point.

src/glyph/pen.rs
madig: But how to handle `anchors` being `None` still?

src/lib.rs
madig: My idea is that Pen is the official thing you build Glyphs with programmatically (and have some correctness checks), instead of constructing things manually.

src/glyph/pen.rs
madig: You can have as many off-curves in front of a qcurve as you want I think.

src/glyph/pen.rs
cmyr: oops I'd misread. :)

src/glyph/pen.rs
cmyr: ah I'd been thinking of `qcurve` is just a quadratic. The docs aren't clear on this point, but what would it mean to have a qcurve after, say, 4 off-curve points?

src/shared_types.rs
```diff
@@ -115,6 +115,19 @@ impl Identifier {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let id = Self(s.into());
+        if id.is_valid() {
+            Ok(id)
+        } else {
+            Err(ErrorKind::BadIdentifier)
+        }
+    }
+}
```
cmyr, line 129: I think this is the right approach, but we should be more rigorous.

- The inner string on `Identifier` should be private; you should only be able to construct an identifier via some function that performs validation
- `is_valid` should move out of the `Identifier` impl and just be part of `from_str`?
- `Identifier` could get a constructor, `fn new(&str) -> Result<Self, ErrorKind>` that just calls this `from_str` impl under the hood?
- Everywhere that we currently construct an identifier directly should be changed to use this constructor

This can be a follow-up PR (or I can do it) if that makes the most sense?

cmyr (commented):
Had an opportunity to give this a bit more thought; some comments inline.

I had an idea for an adjustment to the API, which I am going to call the 'meta-builder pattern'.  The basic idea is that some methods on the builder take as arguments closures that are passed another builder that can be used to build the component in question, like a guideline or an anchor. In use this would look something like,

```rust
let mut pen = Pen::new("test", GlifVersion::V2);
pen.unicode("x")
    .guideline(|builder| builder.identifier("accent-2").line(Line::Horizontal(840.0)))?
    .outline(|builder| builder.add_point((173.0, 422.0), PointType::Line, false, None, None)
        .add_point((193.0, 300.0), PointType::Line, false, None, None)
        .end_path()
        .add_component("hallo", None, "some-ident"))?;
```

The signature for `guideline` would then look something like,

```rust
fn guideline(&mut self, build_fn: impl FnMut(&mut GuidelineBuilder)) -> Result<&mut Self, Error> {
    let mut builder = GuidelineBuilder::new();
    build_fn(&mut builder);
    let guildeline = builder.finish()?;
    // insert guideline
    Ok(self)
}
```

It's very possible this doesn't make sense; in particular there's a sort of funny mismatch between the old API we use for parsing and this new Pen stuff. We could definitely consider rethinking that; one option would be that `GlifParser` goes away, and all those parse methods become just free functions; instead of mutating the pen inside the `GlifParser` they would just return the appropriate type, so we'd have things like,

```rust
fn parse_outline(reader: &mut Reader<[u8]>, buf: &mut Vec<u8>) -> Result<Outline, Error>;
fn parse_contour(reader: &mut Reader<[u8]>, buf: &mut Vec<u8>) -> Result<Contour, Error>;
// etc
```

Let me know what you think? This is all just brainstorming, but it feels like you've been fighting a bit to cram the pen into the existing parsing, and maybe it makes sense for it to live outside of that?

src/glyph/pen.rs
madig: I think you can have infinite off-curves before a qcurve, they are basically compressed qcurves with an implicit qcurve between every off-curve.

src/shared_types.rs
madig: Sounds good, yes :thinking: I think I used is_valid for something but it can be a freestanding function in the Impl. I'll mull over Pens some more, so fel free to toss it in via PR :)

madig:
I like the sub-builder approach and will try to come up with something! I think it may work well enough in GlifParser? I want the best API first, then I can make it work with GlifParser somehow.

What do you mean by funny mismatch? How would free standing parse functions work together?

src/glyph/pen.rs
```diff
@@ -31,10 +92,14 @@ impl Pen {
         }
     }
 
+    /// Return the format version currently set.
```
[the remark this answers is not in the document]
madig: Hm! Yes. It is only used for switching on the format in GlifParser though, maybe there's an even better way?

madig:
So, I tried the sub-builder approach. It's ok for constructing things directly but clashes with GlifParser. E.g. to have an outline builder, closures are awkward because you ideally do everything at once in the closure, but outline parsing is strewn across 4 functions. I think this needs more thinking and more use-cases to flesh out. Might revisit when I know more.

madig:
It was suggested to me to call the pen `GlyphBuilder` instead to avoid confusion with fontTools' Pen concept.

src/shared_types.rs
```diff
@@ -177,7 +178,12 @@ impl Serialize for Guideline {
         let (x, y, angle) = match self.line {
             Line::Vertical(x) => (Some(x), None, None),
             Line::Horizontal(y) => (None, Some(y), None),
-            Line::Angle { x, y, degrees } => (Some(x), Some(y), Some(degrees)),
+            Line::Angle { x, y, degrees } => {
+                if !(0.0..=360.0).contains(&degrees) {
+                    return Err(ser::Error::custom("angle must be between 0 and 360 degrees."));
```
cmyr, line 183: as with `Identifier`, my preference here would be to make illegal values impossible to construct. This could be follow-up work, though.

cmyr (commented):
Did a quick pass, and this seems reasonable. There are definitely some ways I might consider doing this differently, as discussed, but it's hard for me to be helpful without actually trying to play around with the implementation itself, more. Happy to discuss in any case, though!

src/shared_types.rs
madig: That was my thinking. At which level though? making a special `Degrees` newtype?

cmyr (approved):
Happy to get this in as a checkpoint, on the understanding that we're going to keep iterating. :)

PR #58, opened 2020-11-04
More rigorous validation of Identifiers

cmyr:
This came up in the discussion of #56. It should no longer be
possible to create invalid identifiers.


src/shared_types.rs
```diff
@@ -109,22 +138,17 @@ struct RawGuideline {
...
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
     {
-        if self.is_valid() {
-            serializer.serialize_str(&self.0)
-        } else {
-            Err(ser::Error::custom("Identifier must be at most 100 characters long and contain only ASCII characters in the range 0x20 to 0x7E."))
-        }
+        debug_assert!(is_valid_identifier(&self.0));
```
madig, line 150: Not a `ser::Error`?

cmyr: Because we now validate on construction, it should actually be impossible to have an invalid identifier; this is just a sanity check to that effect. I should have added a comment though. :)

madig: Ah, right.

PR #59, opened 2020-11-12
Move Identifier hashing into Self's Impl

madig:


src/shared_types.rs
```diff
@@ -76,6 +78,13 @@ impl Identifier {
     pub fn as_str(&self) -> &str {
         &self.0
     }
+
+    /// Return the u64 hash value of self.
+    pub fn hash_self(&self) -> u64 {
```
cmyr, line 83: I might just call this method `hash`?

cmyr (approved):
Looks good! `hash_self` feels a bit funky as a method name to me, and I think just `hash` should be fine?

madig:
Wait, I'm not done yet! Haven't gotten around to the actual splitting out part.

src/shared_types.rs
madig: I think this is already taken by the Hash trait :D

cmyr:
hahaha but this is such a nice little PR, so easy to merge!

src/shared_types.rs
cmyr: the same name can exist on a type and a trait, that's no problem.

cmyr: ... I think?

madig: Only one  way to find out

madig: Hm, nope. It would result in a recursive call, see the body of the function.

cmyr: try:

```rust
    pub fn hash(&self) -> u64 {
        let mut hasher = DefaultHasher::new();
        Hash::hash(self, &mut hasher);
        hasher.finish()
    }
```

PR #60, opened 2020-11-22
Split OutlineBuilder from GlyphBuilder

madig:
This complicates the process slightly because now there's a hand-off between both builders, but is conceptually nicer. I think.

### DIFF
--- a/src/glyph/builder.rs
+++ b/src/glyph/builder.rs
@@ -45,18 +45,18 @@ use crate::shared_types::Identifier;
 ///             line: Line::Horizontal(10.0),
 ///             name: None,
 ///             color: None,
-///             identifier: Some(Identifier::from_str("test1")?),
+///             identifier: Some(Identifier::new("test1")?),
 ///         })?
 ///         .anchor(Anchor {
 ///             x: 1.0,
 ///             y: 2.0,
 ///             name: Some("anchor1".into()),
 ///             color: None,
-///             identifier: Some(Identifier::from_str("test3")?),
+///             identifier: Some(Identifier::new("test3")?),
 ///         })?;
 ///     let mut outline_builder = OutlineBuilder::new();
 ///     outline_builder
-///         .begin_path(Some(Identifier::from_str("abc")?))?
+///         .begin_path(Some(Identifier::new("abc")?))?
 ///         .add_point((173.0, 536.0), PointType::Line, false, None, None)?
 ///         .add_point((85.0, 536.0), PointType::Line, false, None, None)?
 ///         .add_point((85.0, 0.0), PointType::Line, false, None, None)?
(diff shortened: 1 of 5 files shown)


cmyr (commented):
This seems okay on a fairly quick pass. One concern I've just noticed now is that we're passing around the *hashes* of the identifiers, instead of passing the identifiers themselves; is there a specific rationale for this? It seems like it could possibly cause problems.

madig:
Inserting them into a set consumes them :D I figured we don't want that? And also avoid cloning.

cmyr:
Okay, I think if we're worried about cloning the answer is to make sure we're using something like `Arc<str>` that is cheaper to clone. Hash equality is not item equality, and hash collisions are definitely not impossible.

madig:
It does strike me as somewhat intrusive to make all identifiers `Arc` just so we can hand off easier from `OutlineBuilder` to `GlyphBuilder`. The recent UFO spec addition of [`public.objectLibs`](https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#publicobjectlibs), identifiers will get a more prominent role though in enabling setting libs on various objects identifiable with an `Identifier` :o

cmyr:
Yea sorry! I didn't mean that we should use `Arc<Identifier>`, but rather `Identifier should be declared as `Identifier<Arc<str>>`, instead of as `Identifier<String>`.

src/shared_types.rs
```diff
@@ -91,7 +92,7 @@ impl FromStr for Identifier {
     type Err = ErrorKind;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         if is_valid_identifier(s) {
-            Ok(Identifier(s.to_owned()))
+            Ok(Identifier(s.to_owned().into()))
```
cmyr, line 95: We can just do `s.into()`, as `&str` impls `Into<Arc<str>>`.

src/glyph/builder.rs
```diff
@@ -7,7 +7,7 @@ use crate::glyph::{
 };
 use crate::shared_types::Identifier;
 
-/// A GlyphBuilder is a consuming builder for [`Glyph`](../struct.Glyph.html)s.
+/// A GlyphBuilder is a consuming builder for [`crate::glyph::Glyph`].
```
cmyr, line 10: 🥂

src/glyph/builder.rs
```diff
@@ -316,125 +356,100 @@ impl GlyphBuilder {
...
-                                    }
-                                    PointType::Line => {
-                                        return Err(ErrorKind::UnexpectedPointAfterOffCurve);
+                if self.number_of_offcurves > 0 {
+                    if c.is_closed() {
+                        for point in &c.points {
+                            match point.typ {
+                                PointType::OffCurve => {
+                                    self.number_of_offcurves =
+                                        self.number_of_offcurves.saturating_add(1)
```
cmyr, line 376: so diligent!

realistically, it seems unlikely that we have more that `u32::MAX` preceding off-curve points, but honestly I think using the checked arithmetic methods is a good habit. 👍

cmyr (approved):
Cool, only real thing of note is that I think we can save ourselves needing to do all of these `from_str`s, as mentioned in the more detailed comments.

Thanks again for pushing on with this!

src/glyph/builder.rs
madig: Can't stand edge cases!

PR #63, opened 2020-12-01
Testing what happens if I put FontInfo in a box

cmyr:
this is just a test & should not be merged.


cmyr:
this did approximately nothing, which is a bit of a bummer. :/

PR #67, opened 2020-12-11
WIP: implement layerinfo loading

madig:
Implement layerinfo.plist loading: https://unifiedfontobject.org/versions/ufo3/glyphs/layerinfo.plist/

It implements a new `LayerInfo` struct that is attached to every layer and carries data such as a layer color and a layer lib.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -28,7 +28,7 @@ pub enum Error {
     ExpectedPlistDictionaryError,
     ExpectedPlistStringError,
     ExpectedPositiveValue,
-    TypeError(ErrorKind),
+    InvalidDataError(ErrorKind),
 }
 
 /// An error representing a failure to validate UFO groups.
@@ -149,7 +149,7 @@ impl std::fmt::Display for Error {
             Error::ExpectedPositiveValue => {
                 write!(f, "PositiveIntegerOrFloat expects a positive value.")
             }
-            Error::TypeError(e) => {
+            Error::InvalidDataError(e) => {
                 write!(f, "Type parsing error: '{}", e)
             }
         }
--- a/src/error.rs
+++ b/src/error.rs
@@ -26,7 +26,9 @@ pub enum Error {
     GroupsError(GroupsValidationError),
     GroupsUpconversionError(GroupsValidationError),
     ExpectedPlistDictionaryError,
+    ExpectedPlistStringError,
     ExpectedPositiveValue,
+    TypeError(ErrorKind),
 }
 
 /// An error representing a failure to validate UFO groups.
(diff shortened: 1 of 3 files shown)


src/layer.rs
```diff
@@ -62,7 +65,15 @@ impl Layer {
             })
             //FIXME: come up with a better way of reporting errors than just aborting at first failure
             .collect::<Result<_, _>>()?;
-        Ok(Layer { contents, glyphs })
+
+        let layerinfo_path = path.join(LAYER_INFO_FILE);
+        let info = if layerinfo_path.exists() {
+            LayerInfoData::from_file(&layerinfo_path)?
+        } else {
+            LayerInfoData::default()
```
cmyr, line 73: does this mean that if we load a file that has no layer info, and then save, we will save out an empty `layerinfo.plist`?

src/layer.rs
```diff
@@ -140,6 +154,62 @@ impl Layer {
     }
 }
 
+/// The contents of the [`layerinfo.plist`] file.
+///
+/// [`layerinfo.plist`]: https://unifiedfontobject.org/versions/ufo3/glyphs/layerinfo.plist/
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct LayerInfoData {
```
cmyr, line 161: I'd just call this `LayerInfo`.

src/error.rs
```diff
@@ -26,7 +26,9 @@ pub enum Error {
     GroupsError(GroupsValidationError),
     GroupsUpconversionError(GroupsValidationError),
     ExpectedPlistDictionaryError,
+    ExpectedPlistStringError,
     ExpectedPositiveValue,
+    TypeError(ErrorKind),
```
cmyr, line 31: I would call this a `ParseError` maybe? Or an `InvalidData` error? `ErrorKind` describes, (usually) issues we encounter while parsing a `.glif`.

src/layer.rs
```diff
@@ -151,12 +221,63 @@ mod tests {
...
         assert_eq!(glyph.codepoints.as_ref().unwrap()[0], 'A');
     }
 
+    #[test]
+    fn load_write_layerinfo() {
+        let layer_path = "testdata/mutatorSans/MutatorSansBoldWide.ufo/glyphs";
+        assert!(Path::new(layer_path).exists(), "missing test data. Did you `git submodule init`?");
+        let mut layer = Layer::load(layer_path).unwrap();
+
+        // FIXME Make layer.info no option
```
cmyr, line 251: what does this mean?

cmyr (commented):
This seems reasonable to me, a few little notes inline. I see what you were annoyed about with the value/dictionary stuff 😒

src/layer.rs
madig: It shouldn't. https://github.com/linebender/norad/pull/67/files/0823056311edf80b71c4b350b6ea4b7cf0ff5f79#diff-5eab58a7f7ef848629a8be1084b12464ced845ebd69087b9cbc1a2263465212aR86-R88

src/layer.rs
madig: There is a `LayerInfo` elsewhere already 🤔

src/layer.rs
madig: Oops. Leftover.

src/layer.rs
```diff
@@ -72,6 +83,10 @@ impl Layer {
         let path = path.as_ref();
         fs::create_dir(&path)?;
         plist::to_file_xml(path.join(CONTENTS_FILE), &self.contents)?;
+        // Avoid writing empty layerinfo.plist file.
+        if self.info.color.is_some() || self.info.lib.as_ref().map_or(false, |v| !v.is_empty()) {
```
cmyr, line 87: totally not a blocker but I would normally add a convenience method to `FontInfo` for this, and then we can just do, `if self.info.has_contents() {` or something similar.

cmyr (approved):
Looks good, thanks!

src/layer.rs
madig: Made a `is_empty`!

PR #68, opened 2020-12-16
Serialize kerning values as integers if possible

madig:
This spares us having to use a wrapper type like `IntegerOrFloat` for kerning values.


cmyr (approved):
Cool, looks good!

PR #69, opened 2020-12-23
Add normalization example

madig:
Adds a real-world script (taken from Cantarell) to test-drive norad.

Depends on https://github.com/ebarnard/rust-plist/pull/56 to be merged.


cmyr (approved):
Looks good, happy to get see this merged when the upstream patch lands.

PR #72, opened 2020-12-29
Protect various objects' identifier and lib

madig:
In preparation of adding object libs handling.

This de-pubs `lib` and `identifier` and adds some getters and setters so we can make sure that

1. Having a lib ensures that we also have an identifier
2. The identifier cannot be removed while a lib is present.

Here, we set a UUID v4 identifier if none exists on `replace`ing the lib. Use UUID because we can generate one and be reasonably sure that we probably don't introduce a duplicate, saving us a lengthy check. No support for removing the identifier because the spec says an id should not be removed without reason and I haven't had a need yet.

### DIFF
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -14,13 +14,10 @@ use druid::{Data, Lens};
 
 use crate::error::{Error, GlifError, GlifErrorInternal};
 use crate::names::NameList;
-use crate::shared_types::{Color, Guideline, Identifier, Line};
+use crate::shared_types::{Color, Guideline, Identifier, Line, Plist};
 
 /// The name of a glyph.
 pub type GlyphName = Arc<str>;
-
-/// A Plist dictionary.
-pub type Plist = plist::Dictionary;
 
 /// A glyph, loaded from a [.glif file][glif].
 ///
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -246,13 +246,13 @@ impl Anchor {
     }
 
     /// Returns an immutable reference to the anchor's lib.
-    pub fn get_lib(&self) -> &Option<Plist> {
-        &self.lib
+    pub fn lib(&self) -> Option<&Plist> {
+        self.lib.as_ref()
     }
 
     /// Returns a mutable reference to the anchor's lib.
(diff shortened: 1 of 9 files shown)


src/glyph/mod.rs
```diff
@@ -220,6 +233,194 @@ pub struct AffineTransform {
...
+        name: Option<String>,
+        color: Option<Color>,
+        identifier: Option<Identifier>,
+        lib: Option<Plist>,
+    ) -> Self {
+        Self { x, y, name, color, identifier, lib }
+    }
+
+    /// Returns an immutable reference to the anchor's lib.
+    pub fn get_lib(&self) -> &Option<Plist> {
```
cmyr, line 249: No need for `get_` in getters; you can just call this `lib`.

Also: we should return `Option<&Plist>` instead of `&Option<Plist>`; you can do this with `self.lib.as_ref()`.

src/glyph/mod.rs
```diff
@@ -220,6 +233,194 @@ pub struct AffineTransform {
...
+        &self.lib
+    }
+
+    /// Returns a mutable reference to the anchor's lib.
+    pub fn get_lib_mut(&mut self) -> &mut Option<Plist> {
```
cmyr, line 254: ditto with `lib_mut` and `Option<&mut Plist>`.

cmyr (commented):
Overall looks good, a couple API notes; I only made these in one place but they apply to all of these types.

src/glyph/mod.rs
madig: This seems to mess with innocuous looking code:

```
   --> src/glyph/builder.rs:155:50
    |
155 |         insert_identifier(&mut self.identifiers, guideline.identifier().clone())?;
    |                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `shared_types::Identifier`, found `&shared_types::Identifier`
    |
    = note: expected enum `std::option::Option<shared_types::Identifier>`
               found enum `std::option::Option<&shared_types::Identifier>`
```

madig: Ah. I have to use `cloned()`.

cmyr (approved):
Looks good!

PR #75, opened 2020-12-31
Handle object libs

madig:
Supersedes https://github.com/linebender/norad/pull/71.

Handle object libs by attaching libs directly to glyph objects and automatically handling `public.objectLibs` on loading and saving.

### DIFF
--- a/src/fontinfo.rs
+++ b/src/fontinfo.rs
@@ -782,7 +782,7 @@ impl FontInfo {
     }
 
     /// Dump guideline libs into a Plist.
-    pub fn dump_object_libs(&self) -> Plist {
+    pub(crate) fn dump_object_libs(&self) -> Plist {
         let mut object_libs = Plist::default();
 
         if let Some(guidelines) = &self.guidelines {
--- a/src/fontinfo.rs
+++ b/src/fontinfo.rs
@@ -333,7 +333,7 @@ impl FontInfo {
     pub fn from_file<P: AsRef<Path>>(
         path: P,
         format_version: FormatVersion,
-        lib: &mut Option<Plist>,
+        lib: Option<&mut Plist>,
     ) -> Result<Self, Error> {
         match format_version {
             FormatVersion::V3 => {
--- a/src/fontinfo.rs
+++ b/src/fontinfo.rs
@@ -340,7 +340,7 @@ impl FontInfo {
                 let mut fontinfo: FontInfo = plist::from_file(path)?;
                 fontinfo.validate()?;
                 if let Some(lib) = lib {
-                    fontinfo.fill_in_libs(lib)?;
+                    fontinfo.load_object_libs(lib)?;
                 }
                 Ok(fontinfo)
             }
@@ -760,22 +760,20 @@ impl FontInfo {
 
     /// Move libs from the font lib's `public.objectLibs` key into the actual objects.
     /// The key will be removed from the font lib.
-    fn fill_in_libs(&mut self, lib: &mut Plist) -> Result<(), Error> {
-        let object_libs = match lib.remove(PUBLIC_OBJECT_LIBS_KEY) {
+    fn load_object_libs(&mut self, lib: &mut Plist) -> Result<(), Error> {
+        let mut object_libs = match lib.remove(PUBLIC_OBJECT_LIBS_KEY) {
             Some(lib) => lib.into_dictionary().ok_or(Error::InvalidDataError(ErrorKind::BadLib))?,
             None => return Ok(()),
         };
 
-        'next_key: for (key, value) in object_libs.into_iter() {
-            let value =
-                value.into_dictionary().ok_or(Error::InvalidDataError(ErrorKind::BadLib))?;
-
-            if let Some(guidelines) = &mut self.guidelines {
-                for guideline in guidelines {
-                    if guideline.identifier().map(Identifier::as_str) == Some(&key) {
-                        guideline.replace_lib(value);
-                        continue 'next_key;
-                    }
+        if let Some(guidelines) = &mut self.guidelines {
+            for guideline in guidelines {
+                if let Some(lib) =
+                    guideline.identifier().and_then(|id| object_libs.remove(id.as_str()))
+                {
+                    let lib =
+                        lib.into_dictionary().ok_or(Error::InvalidDataError(ErrorKind::BadLib))?;
+                    guideline.replace_lib(lib);
                 }
             }
         }
@@ -784,7 +782,7 @@ impl FontInfo {
     }
 
     /// Dump guideline libs into a Plist.
-    pub fn libs_to_object_libs(&self) -> Plist {
+    pub fn dump_object_libs(&self) -> Plist {
         let mut object_libs = Plist::default();
 
         if let Some(guidelines) = &self.guidelines {
--- a/src/fontinfo.rs
+++ b/src/fontinfo.rs
@@ -340,7 +340,7 @@ impl FontInfo {
                 let mut fontinfo: FontInfo = plist::from_file(path)?;
                 fontinfo.validate()?;
                 if let Some(lib) = lib {
-                    fill_in_libs(&mut fontinfo, lib)?;
+                    fontinfo.fill_in_libs(lib)?;
                 }
                 Ok(fontinfo)
             }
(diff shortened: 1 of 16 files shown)


src/glyph/parse.rs
```diff
@@ -532,6 +536,73 @@ fn start(reader: &mut Reader<&[u8]>, buf: &mut Vec<u8>) -> Result<GlyphBuilder,
     }
 }
 
+/// Move libs from the lib's `public.objectLibs` into the actual objects.
+/// They key will be removed from the glyph lib.
+fn fill_in_libs(glyph: &mut Glyph) -> Result<(), Error> {
+    if let Some(glyph_lib) = &mut glyph.lib {
+        if let Some(object_libs) = glyph_lib.remove("public.objectLibs") {
```
cmyr, line 543: I would write this as,

```rust
    let object_libs = match glyph.lib.as_mut().and_then(|lib| lib.remove("public.objectLibes")) {
        Some(lib) => lib,
        None => return Ok(()),
    };
```

And then you save yourself two layers of nesting, below.

src/glyph/parse.rs
```diff
@@ -532,6 +536,73 @@ fn start(reader: &mut Reader<&[u8]>, buf: &mut Vec<u8>) -> Result<GlyphBuilder,
...
+
+            'next_key: for (key, value) in object_libs.into_iter() {
+                let value = value
+                    .into_dictionary()
+                    .ok_or(GlifErrorInternal::Spec { kind: ErrorKind::BadLib, position: 0 })?;
+
+                if let Some(anchors) = &mut glyph.anchors {
+                    for anchor in anchors {
+                        if let Some(id) = anchor.identifier() {
+                            if id == &key {
```
cmyr, line 556: you could also replace this with,

```rust
if anchor.identifier().map(Identifier::as_str) == Some(&key) { ... }
```

if you wanted.

src/glyph/serialize.rs
```diff
@@ -70,7 +70,20 @@ impl Glyph {
...
+        // to store them inline, so they have to be placed into the glyph's lib
+        // under the public.objectLibs parent key. To avoid mutation behind the
+        // client's back, object libs are written out but not stored in
+        // glyph.lib in-memory. If there are object libs to serialize, clone the
+        // existing lib and insert them there for serialization, otherwise avoid
+        // cloning and write out the original.
+        let object_libs = libs_to_object_libs(&self);
+        if !object_libs.is_empty() {
+            let mut new_lib =
+                if let Some(lib) = self.lib.as_ref() { lib.clone() } else { Plist::new() };
```
cmyr, line 83: maybe,

```rust
let mut new_lib = self.lib.clone().unwrap_or_else(|| Plist::new());
```

src/glyph/serialize.rs
```diff
@@ -70,7 +70,20 @@ impl Glyph {
...
+            new_lib.insert("public.objectLibs".into(), plist::Value::Dictionary(object_libs));
+            write_lib_section(&new_lib, &mut writer)?;
+        } else if let Some(lib) = self.lib.as_ref() {
```
cmyr, line 86: we can also combine these next two lines if we wanted, with

```rust
        } else if let Some(lib) = self.lib.as_ref().filter(|lib| !lib.is_empty()) {
            write_lib_section(lib, &mut writer)?;
        }
```

src/glyph/serialize.rs
```diff
@@ -84,6 +97,71 @@ impl Glyph {
     }
 }
 
+fn libs_to_object_libs(glyph: &Glyph) -> Plist {
+    let mut object_libs = Plist::default();
+
```
cmyr, line 102: we could write less code by declaring a little closure here to handle the repeated adding of an id/lib pair:

```rust
let mut add_lib |id: Option<&Identifer>, lib: &Dictionary| {
    let id = id.map(|id| id.as_str().to_string());
    debug_assert!(id.is_some());
    object_libs.insert(id.unwrap(), plist::Value::Dictionary(lib.clone()));
};
```

And then elsewhere we just call this, like:

```rust
for anchor in anchor {
    if let Some(lib) = anchor.lib() {
        add_lib(anchor.identifier(), lib);
    }
}
```


On `debug_assert!`:

`debug_assert!` doesn't make a lot of sense here, since we're going to unwrap anyway. `debug_assert!` would make more sense if we were going to just skip these items in a release build or something like that? What do we actually want to do in this case?

src/ufo.rs
```diff
@@ -269,9 +284,23 @@ impl Ufo {
...
+        // under the public.objectLibs parent key. To avoid mutation behind the
+        // client's back, object libs are written out but not stored in
+        // font.lib in-memory. If there are object libs to serialize, clone the
+        // existing lib and insert them there for serialization, otherwise avoid
+        // cloning and write out the original.
+        let object_libs = libs_to_object_libs(&self.font_info);
+        if !object_libs.is_empty() {
+            let mut new_lib =
+                if let Some(lib) = self.lib.as_ref() { lib.clone() } else { Plist::new() };
+            new_lib.insert("public.objectLibs".into(), plist::Value::Dictionary(object_libs));
```
cmyr, line 298: I think we should declare this string as a `const` somewhere.

src/ufo.rs
```diff
@@ -419,6 +448,27 @@ fn validate_groups(groups_map: &Groups) -> Result<(), GroupsValidationError> {
     Ok(())
 }
 
+fn libs_to_object_libs(fontinfo: &Option<FontInfo>) -> Plist {
```
cmyr, line 451: ditto I think this should be a method on `FontInfo`, and then at the callsite you would do,

`self.font_info.map(FontInfo::libs_to_object_libs).unwrap_or_else(|| Plist::new())` or what-have-you.

src/glyph/parse.rs
```diff
@@ -532,6 +536,73 @@ fn start(reader: &mut Reader<&[u8]>, buf: &mut Vec<u8>) -> Result<GlyphBuilder,
...
+                                component.replace_lib(value);
+                                continue 'next_key;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
```
cmyr, line 601: It's maybe worth noting that this has quadratic runtime; if a glyph contains a large number of items with libs, and a layer contains a large number of such glyphs, we'll end up doing some unnecessary work.

A more robust approach might be to iterate once through each set of items (components, points, etc) and then check if that item's id exists in the object libs map; if it does, take out the lib and assign it to the item. This way we're O(n) over the number of items, instead of O(n*m) over the number of items and the number of object libs.

performance here could be quite bad. If we're worried about worst-case performance, we might want to consider, for each objectlib type, building

cmyr (commented):
Okay a bunch code-style and idiom suggestions but no major structural concerns, outside of maybe the parse performance, which probably isn't a huge issue in practice.

Excited to get this landed; I'm going to need to use it for the spline I think. :)

src/glyph/serialize.rs
madig: I just wanted to have a non-crpytic error message if this invariant somehow fails but I'm going to remove it :smile_cat: need to give error handling some deeper thought anyway.

src/glyph/parse.rs
madig: Hm! Yeah I figured. I'll tinker with this.

src/fontinfo.rs
```diff
@@ -332,11 +333,15 @@ impl FontInfo {
     pub fn from_file<P: AsRef<Path>>(
         path: P,
         format_version: FormatVersion,
+        lib: &mut Option<Plist>,
```
cmyr, line 336: I think we should always pass `Option<&T>` instead of `&Option<T>` unless we *explicilty* want to do something like assign a new value if there's `None`, or replace some value *with* `None`.

src/glyph/mod.rs
```diff
@@ -95,6 +100,113 @@ impl Glyph {
...
+        Ok(())
+    }
+
+    /// Dump guideline libs into a Plist.
+    fn dump_object_libs(&self) -> Plist {
+        let mut object_libs = Plist::default();
+
+        let mut dump_lib = |id: Option<&Identifier>, lib: &Plist| {
+            let id = id.map(|id| id.as_str().to_string());
+            object_libs.insert(id.unwrap(), plist::Value::Dictionary(lib.clone()));
```
cmyr, line 171: Are there any possible circumstances where we don't have an identifier? What would they be?

cmyr (approved):
Two little notes, but nothing blocking. Thanks again for doing so much work on this!

src/glyph/mod.rs
madig: I don't know :O maybe if we decide to allow id removal at some point and mess the logic up.

PR #79, opened 2021-01-30
Use enums to keep state

madig:
After reading https://hoverbear.org/blog/rust-state-machine-pattern/, I wanted to try and be a bit more clever about state management in `OutlineBuilder`.

I wonder if there should be a third state `Failure(ErrorKind)` that is sticky after an error happened? It would complicate the code somewhat though, I think. For example:

1. If `begin_path` fails, you can just carry on drawing the previously started contour. It doesn't change the state.
2. If `add_point` fails, it simply doesn't add what you were trying to add. It doesn't change the state.
3. If `end_path` fails, it just drops the contour you were trying to end. It doesn't change the state.
4. If `add_component` fails, it simply doesn't add the component you were trying to add. It doesn't change the state.
5. if `finish` fails, it simply won't return an outline to you. It doesn't change the state.

I think that's reasonably simple so we don't strictly need a `Failure` state?

### DIFF
--- a/src/glyph/builder.rs
+++ b/src/glyph/builder.rs
@@ -268,8 +268,19 @@ impl GlyphBuilder {
 pub struct OutlineBuilder {
     identifiers: HashSet<Identifier>,
     outline: Outline,
-    scratch_contour: Option<Contour>,
-    number_of_offcurves: u32,
+    scratch_state: OutlineBuilderState,
+}
+
+#[derive(Debug)]
+enum OutlineBuilderState {
+    Idle,
+    Drawing { scratch_contour: Contour, number_of_offcurves: u32 },
+}
+
+impl Default for OutlineBuilderState {
+    fn default() -> Self {
+        OutlineBuilderState::Idle
+    }
 }
 
 impl OutlineBuilder {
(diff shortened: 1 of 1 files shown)


src/glyph/builder.rs
```diff
@@ -268,107 +268,145 @@ impl GlyphBuilder {
...
                 Ok(self)
             }
-            None => Err(ErrorKind::PenPathNotStarted),
         }
     }
 
-    /// Ends the path begun by `begin_path()` and adds the contour it to the glyph's outline.
-    ///
-    /// Discards path in case of error.
+    /// Ends the path begun by [`Self::begin_path`] and adds the contour it to the glyph's outline, unless
```
cmyr, line 385:
```suggestion
    /// Ends the path begun by [`Self::begin_path`] and adds the contour to the glyph's outline, unless
```

src/glyph/builder.rs
```diff
@@ -268,107 +268,145 @@ impl GlyphBuilder {
...
+    ///
+    /// It must be finished with [`Self::end_path`] before the outline can be [`Self::finish`]ed
+    /// and retrieved.
     ///
     /// Errors when:
     /// 1. a path has been begun already but not ended yet.
     /// 2. the identifier is not unique within the glyph.
+    ///
+    /// On error, it won't begin a new path and you can continue drawing the previously started
+    /// path.
```
cmyr, line 301: This feels iffy. Shouldn't calls to `start_path` that are not the *first* path in an outline always be preceded with `end_path`?

cmyr (commented):
I use state machines like this for a lot of stuff in runebender, but mostly for tracking gestures. I think it generally makes sense, although this does feel like a pretty simple case, and it doesn't look like a huge win here, but certainly not bad.


Some of the behaviour on error conditions seems a bit iffy to me, but that's my only real feedback. :)

src/glyph/builder.rs
madig: Yes; the error case is that you do `begin_path` followed by `begin_path`. As in, this errors out if you didn't do `end_path` after the first `begin_path`.

cmyr (approved):
Looks good!

Issue #81, opened 2021-02-04
`new` constructors for Contour, ContourPoint and possibly others need to deal with identifier/lib args


madig:
E.g. https://github.com/linebender/norad/blob/v0.3.1/src/glyph/mod.rs#L397-L403 needs to be 

```rust
    pub fn new(
        points: Vec<ContourPoint>,
        identifier: Option<Identifier>,
        lib: Option<Plist>,
    ) -> Self {
        let mut c = Self { points, identifier: None, lib: None };
        if let Some(id) = identifier {
            c.replace_identifier(id);
        }
        if let Some(lib) = lib {
            c.replace_lib(lib);
        }
        c
    }
```

or something. Same for other structs that have a identifier/lib field.

Then https://github.com/linebender/runebender/pull/221 can be reverted.

PR #82, opened 2021-02-06
Properly account for identifier and lib arguments

madig:
Closes https://github.com/linebender/norad/issues/81.


cmyr (approved):
That works, thanks!

PR #86, opened 2021-02-19
Make kurbo a direct optional dependency

madig:
So I don't have to pull in druid to use kurbo converters.


cmyr (approved):
Yep this looks good!

PR #88, opened 2021-02-23
Remove comment-triggering from bloatcmp

cmyr:


madig:
Thanks! I was wondering why I was getting bloat bot notifications for leaving comments :o

Issue #92, opened 2021-03-24
Glyph: get Vecs out of Options?


madig:
Currently, contours, components, anchors, codepoints and guidelines are Vecs hidden behind an `Option`, which seems to be because they allow for non-existence of attributes in a glif file. This leads to code like 

```rust
fn move_glyph(glyph: &mut Glyph, delta: f32) {
    let Some(outline) = &mut glyph.outline {
        // ...
    }
}
```

which I don't find especially, uh, ergonomic? Same goes for  anchors, etc. I'd rather have a 

```rust
pub struct Glyph {
    pub name: GlyphName,
    pub format: GlifVersion,
    pub advance: Option<Advance>,
    pub codepoints: Vec<char>,
    pub note: Option<String>,
    pub guidelines: Vec<Guideline>,
    pub anchors: Vec<Anchor>,
    pub contours: Vec<Contour>,
    pub components: Vec<Component>,
    pub image: Option<Image>,
    pub lib: Option<Plist>,
}
```

Since empty Vecs can just not be serialized (and don't alloc until they have one member), I don't think making them optional holds semantic value? Same for `lib` maybe?

Related to https://github.com/linebender/norad/issues/27.

cmyr:
Yes, it's gross. my intention was to losslessly adhere to the UFO spec, and represent all fields that were optional as `Option<_>` on the rust code. I'm not sure if this matters nor not? The way that it might matter is for round-tripping, if we wanted to preserve empty collections or something.

I defer to you, here; if you think it makes sense to use an empty vec to represent a missing field, and not serialize empty vecs, that sounds good to me.

madig:
- codepoints: defined by a full element, so can't be an empty collection
- guidelines: same
- anchors: same
- components: same
- contours: can have empty collection, but see https://github.com/unified-font-object/ufo-spec/issues/150

FWIW, ufoLib2 and defcon use empty lists for the above.

cmyr:
Happy to see this change, it will be a nice ergonomics improvement.

madig:
I just noticed: in the same vein, advance width and height are kept behind Option. The spec says https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#advance, i.e. both could be regular fields defaulting to 0.

cmyr:
this also sounds good, the only reason these various things are awkward is because i did not feel knowledgeable enough to take any liberties.

madig:
Good thing you now have extreme prejudice on your team.

PR #93, opened 2021-03-27
Simplify Glyph

madig:
Closes https://github.com/linebender/norad/issues/92.

I'm keeping `Outline` around for only-once checking. Maybe... I can cut down the builders again and get the only-once logic back into the glif parser...

### DIFF
--- a/src/glyph/builder.rs
+++ b/src/glyph/builder.rs
@@ -2,8 +2,8 @@ use std::collections::HashSet;
 
 use crate::error::ErrorKind;
 use crate::glyph::{
-    Advance, AffineTransform, Anchor, Component, Contour, ContourPoint, GlifVersion, Glyph,
-    GlyphName, Guideline, Image, Outline, PointType,
+    AffineTransform, Anchor, Component, Contour, ContourPoint, GlifVersion, Glyph, GlyphName,
+    Guideline, Image, PointType,
 };
 use crate::shared_types::{Identifier, Plist};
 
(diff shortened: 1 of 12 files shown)


madig:
I am now looking at de-Option-ing the libs of Anchor, Guideline, Contour, ContourPoint and Component. There, we have the added headache of identifiers. The `replace_lib` method takes care to assign an identifier.. but what if we don't have an Option to guide us? Auto-assign in `lib_mut` even when nothing is added?

cmyr:
Is the option in these types a big problem, ergonomics-wise? I think that creating an identifier on access is reasonable, it's sort of like 'copy on write' semantics.

madig:
Since there doesn't seem to be much code accessing those libs, I can't say, it _is_ very slightly annoying when you have to, though. Maybe I'll leave it for now...

madig:
I wonder why a release build is bigger now :thinking:

src/glyph/builder.rs
```diff
@@ -271,6 +282,12 @@ pub struct OutlineBuilder {
     scratch_state: OutlineBuilderState,
 }
 
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct Outline {
```
cmyr, line 286: If this is now purely an internal type, we can remove the `pub`.

cmyr (approved):
One little note, otherwise this looks good to me!

src/glyph/builder.rs
madig: Hm, then I can't use it in parse.rs :thinking:

madig: Removed it from the public API.

PR #96, opened 2021-04-01
Move stuff around

madig:
I split various UFO types into their own files to better organize them in my head and reduce scrolling distance in various files. This is similar to how defcon and ufoLib2 are organized.


madig:
Will split this up.

PR #97, opened 2021-04-02
Move Guideline to own file

madig:
I want to split various UFO types into their own files to better organize them in my head and reduce scrolling distance in various files. This is similar to how defcon and ufoLib2 are organized.

### DIFF
--- a/src/shared_types.rs
+++ b/src/shared_types.rs
@@ -479,8 +479,11 @@ impl<'de> Deserialize<'de> for NonNegativeIntegerOrFloat {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use std::convert::TryFrom;
+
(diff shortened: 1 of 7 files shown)


src/shared_types.rs
```diff
@@ -479,9 +316,11 @@ impl<'de> Deserialize<'de> for NonNegativeIntegerOrFloat {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
     use serde_test::{assert_tokens, Token};
 
+    use super::*;
+    use crate::*;
```
cmyr, line 322: does this make `use super::*` redundant?

cmyr (approved):
Looks good!

src/shared_types.rs
madig: Not entirely, but I have expanded the imports now.

PR #99, opened 2021-04-03
I don't think the fontinfo and glyph mods need to be pub?

madig:
Does `error` need to be pub? :thinking:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,8 +20,8 @@ extern crate serde_derive;
 extern crate serde_repr;
 
 pub mod error;
-pub mod fontinfo;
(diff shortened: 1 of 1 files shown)


cmyr:
I have a line in runebender, `use norad::glyph::{Contour, ContourPoint, Glyph, GlyphName, PointType};`. I assume all these types are now exported elsewhere?

madig:
It should all be top-level now.

madig:
As in, `use norad::{Contour, ContourPoint, Glyph, GlyphName, PointType};.`.

cmyr (approved):
sounds good!

PR #100, opened 2021-04-03
Fold LayerInfo into Layer

madig:
I don't think they need to be separate.

I'm not entirely sure about `Layer::new_named` taking a layer _and_ dir name, but I suppose that's hard to avoid until we have a dedicated LayerSet stuct attached to Ufo that is essentially a HashMap and manages name and path name (collisions).

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -22,28 +22,44 @@ static LAYER_INFO_FILE: &str = "layerinfo.plist";
 /// [layer]: http://unifiedfontobject.org/versions/ufo3/glyphs/
 #[derive(Debug, Clone, Default, PartialEq)]
 pub struct Layer {
+    pub name: String,
+    /// The layer dir name inside the UFO.
+    pub path: PathBuf,
     pub(crate) glyphs: BTreeMap<GlyphName, Arc<Glyph>>,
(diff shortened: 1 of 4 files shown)


cmyr:
I'm trying to recall the rationale for keeping these separate. I think the idea is that a `Layer` is really just a bag of glyphs in some directory, but the name and path of that directory live outside of the layer, in `layercontents.plist`?

I don't have strong opinions here, but looking through the code it isn't clear that this patch improves ergonomics in a meaningful way? Having one less type is nice.

Maybe one way to think about `LayerInfo` is that our `Vec<LayerInfo>` *is* our `LayerSet`?

madig:
> Maybe one way to think about LayerInfo is that our Vec<LayerInfo> is our LayerSet?

I think: yes. However, not quite, as e.g. ufoLib2's LayerSet prevents deletion of the default layer, making get_default_layer never fail.

So maybe, instead, I should turn LayerInfo into a 

```rust
struct LayerSet {
    contents: HashMap<LayerName, OsPath>,
    layers: HashMap<LayerName, Layer>
}
```

?

cmyr:
I'm starting work today on the python bindings project, so maybe let's leave this for now and then when I go to implement the python LayerSet I'll take a look then and try to figure out a solution?

PR #102, opened 2021-04-04
Use two spaces instead of a tab per indent level.

### LINKED ISSUE #101: Tabs vs. spaces
defcon and FontForge seem to prefer two-space indentation in UFO XML files, but Norad uses tabs.

I see a PR in the past saying that ufonormalizer used a single tab instead, but everything else uses two spaces so I'm a bit confused. From where I stand, more tools use two spaces than one tab, and ufonormalizer is the odd one out. The UFO specification itself has all of its examples set with two spaces per indent as well, which I'm sure contributes to it being a popular option for tools.

xorgy:
AFAIK:

Most UFO editing/processing tools output XML with two-space indents,
so Norad should do the same so that people don't need to use external
xml formatting tools to merge work between tools using Norad and tools
using defcon or FontForge.

Resolves #101

### DIFF
--- a/src/glyph/serialize.rs
+++ b/src/glyph/serialize.rs
@@ -21,7 +21,7 @@ impl Glyph {
     }
 
     fn encode_xml_impl(&self) -> Result<Vec<u8>, WriteError> {
-        let mut writer = Writer::new_with_indent(Cursor::new(Vec::new()), b'\t', 1);
+        let mut writer = Writer::new_with_indent(Cursor::new(Vec::new()), b' ', 2);
         writer.write_event(Event::Decl(BytesDecl::new(b"1.0", Some(b"UTF-8"), None)))?;
         let mut start = BytesStart::borrowed_name(b"glyph");
         start.push_attribute(("name", &*self.name));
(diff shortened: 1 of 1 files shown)


xorgy:
My motivation for this change was seeing this PR <https://github.com/eliheuer/baby-shark/pull/3> where there is a clear disagreement between tabs and spaces causing almost every line in the font to be modified, rather than just inserting one clean line into each glif file.

src/glyph/serialize.rs
```diff
@@ -106,7 +106,7 @@ fn write_lib_section<T: Write>(lib: &Plist, writer: &mut Writer<T>) -> Result<()
     let as_value: plist::Value = lib.to_owned().into();
     let mut out_buffer = Vec::with_capacity(256); // a reasonable min size?
     as_value.to_writer_xml(&mut out_buffer)?;
-    let lib_xml = String::from_utf8(out_buffer).expect("xml writer writs valid utf8");
+    let lib_xml = String::from_utf8(out_buffer).expect("XML writer writes valid UTF-8");
```
cmyr, line 109: geez was I drunk or what

cmyr (commented):
I'm.... really not sure what to think about this.

UfoNormalizer may be in the majority, but it is still the accepted... normalizer of UFOs.

a single tab also uses 50% fewer bytes! which okay, whatever, but I was just on a call with some folks who have VF projects using UFO where they're regularly manipulating 400MB of UFO, so it ain't nothing.


So I guess currently I'm feeling the pull of inertia on this one?

chrissimpkins:
Here are production Noto UFO source comparisons across other commonly used UFO serializers:

https://github.com/chrissimpkins/ufo-format-diff

Each tool has its own PR

chrissimpkins:
https://github.com/linebender/norad/pull/148

cmyr:
I'm going to close this. As of #148 the user can optionally specify custom whitespace.

PR #106, opened 2021-04-15
LayerSet + other layer-related tweaks

### LINKED ISSUE #84: Add Ufo::new_layer(name) method
Should insert an empty new named layer and return a mutable reference to it. UX problem: what if your chosen layer name exists already? Return `Result<Error, &mut Layer>`?

Maybe even a combination method like `get_or_insert_new(name)`? Or something mirroring https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert. Use case is mutably grabbing e.g. the `public.background` layer or creating it if it doesn't exist yet. For when you just want to add stuff into it.

### LINKED ISSUE #89: Ensure default layer is always present
The layer list is currently just a Vec, maybe it should be a wrapper struct that ensures the default layer is always present. This can also enforce that the name `public.default` is only used for the default layer.

Once this is ensured, `Ufo::get_default_layer` can return `&Layer` instead of an `Option`.

cmyr:
This adds a LayerSet type that enforces various constraints on
layers.

Among other things, this means that:

- Layer names are now also stored as Arc<str>
- There is no more LayerInfo type
- there is now always a default layer
- the default layer cannot be removed
- new layers can be added

This also introduces a few other breaking api changes; for instance
I'm removing `Font::get_default_layer(&self) -> Option<&Layer>`
in favour of `Font::default_layer(&self) -> &Layer`.

In addition it moves the "compute a file name from a user provided name"
logic into a new `util` module.

This was motivated by my experiments with python bindings,
but I think this is generally useful independent of that work
and so I'm PRing it separately.

- supersedes and closes #100
- closes #84 (with an asterisk, this method is on the public `LayerSet` type)
- closes #89 

@madig I caught a few issues here thanks to your tests! 🎉

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -171,7 +171,7 @@ impl Layer {
     pub fn insert_glyph(&mut self, glyph: impl Into<Arc<Glyph>>) {
         let glyph = glyph.into();
         if !self.contents.contains_key(&glyph.name) {
-            let path = crate::glyph::default_file_name_for_glyph_name(&glyph.name);
+            let path = crate::util::default_file_name_for_glyph_name(&glyph.name);
             self.contents.insert(glyph.name.clone(), path.into());
         }
         self.glyphs.insert(glyph.name.clone(), glyph);
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -16,34 +16,203 @@ use crate::{Error, Glyph, Plist};
 static CONTENTS_FILE: &str = "contents.plist";
 static LAYER_INFO_FILE: &str = "layerinfo.plist";
 
+pub(crate) static LAYER_CONTENTS_FILE: &str = "layercontents.plist";
+pub(crate) static DEFAULT_LAYER_NAME: &str = "public.default";
+pub(crate) static DEFAULT_GLYPHS_DIRNAME: &str = "glyphs";
+
+pub type LayerName = Arc<str>;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct LayerSet {
+    // The first layer is always the 'default layer'.
+    layers: Vec<Layer>,
+}
+
+#[allow(clippy::clippy::len_without_is_empty)] // never empty
+impl LayerSet {
+    /// Load the layers from the provided path.
+    ///
+    /// If a `layercontents.plist` file exists, it will be used, otherwise
+    /// we will assume the pre-UFOv3 behaviour, and expect a single glyphs dir.
+    ///
+    /// The `glyph_names` argument allows norad to reuse glyph name strings,
+    /// reducing memory use.
+    pub fn load(base_dir: &Path, glyph_names: &NameList) -> Result<LayerSet, Error> {
+        let layer_contents_path = base_dir.join(LAYER_CONTENTS_FILE);
+        let to_load: Vec<(LayerName, PathBuf)> = if layer_contents_path.exists() {
+            plist::from_file(layer_contents_path)?
+        } else {
+            vec![(Arc::from(DEFAULT_LAYER_NAME), PathBuf::from(DEFAULT_GLYPHS_DIRNAME))]
+        };
+
+        let mut layers: Vec<_> = to_load
+            .into_iter()
+            .map(|(name, path)| {
+                let layer_path = base_dir.join(&path);
+                Layer::load_impl(&layer_path, name, &glyph_names)
+            })
+            .collect::<Result<_, _>>()?;
+
+        // move the default layer to the front
+        let default_idx = layers
+            .iter()
+            .position(|l| l.path.to_str() == Some(DEFAULT_GLYPHS_DIRNAME))
+            .ok_or(Error::MissingDefaultLayer)?;
+        layers.rotate_left(default_idx);
+
+        Ok(LayerSet { layers })
+    }
+
+    /// The number of layers in the set.
+    ///
+    /// This should be non-zero.
+    pub fn len(&self) -> usize {
+        self.layers.len()
+    }
+
+    /// Get a reference to a layer, by name.
+    pub fn get(&self, name: &str) -> Option<&Layer> {
+        self.layers.iter().find(|l| &*l.name == name)
+    }
+
+    /// Get a mutable reference to a layer, by name.
+    pub fn get_mut(&mut self, name: &str) -> Option<&mut Layer> {
+        self.layers.iter_mut().find(|l| &*l.name == name)
+    }
+
+    /// A reference to the default layer.
+    pub fn default_layer(&self) -> &Layer {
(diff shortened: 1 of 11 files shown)


src/ufo.rs
```diff
@@ -406,66 +356,48 @@ impl Font {
...
     pub fn iter_names(&self) -> impl Iterator<Item = GlyphName> + '_ {
-        // this is overly complicated for opaque lifetime reasons, aka 'trust me'
-        self.layers
-            .iter()
-            .filter(|l| l.path.file_name() == Some(OsStr::new(DEFAULT_GLYPHS_DIRNAME)))
-            .flat_map(|l| l.layer.glyphs.keys().cloned())
+        self.layers.default_layer().glyphs.keys().cloned()
     }
 
     //FIXME: support for multiple layers.
```
madig, line 403: ufoLib2 doesn't support multiple layers either, it's ok to make the user access other layers in question directly.

src/upconversion.rs
```diff
@@ -7,12 +7,13 @@ use crate::shared_types::IntegerOrFloat;
...
 
-/// Convert kerning groups and pairs from v1 and v2 informal conventions to v3 formal conventions.
-/// Converted groups are added (duplicated) rather than replacing the old ones to preserve all data
-/// that external entities might rely on. Kerning pairs are updated to reflect the new group names.
+/// Convert kerning groups and pairs from v1 and v2 informal conventions to
+/// v3 formal conventions. Converted groups are added (duplicated) rather than
+/// replacing the old ones to preserve all data that external entities might
+/// rely on. Kerning pairs are updated to reflect the new group names.
 ///
-/// This is an adaptation from the fontTools.ufoLib reference implementation. It will not check if
```
madig, line 14: Is this an accidental reformatting?

src/layer.rs
```diff
@@ -16,34 +16,197 @@ use crate::{Error, Glyph, Plist};
...
+        } else {
+            let layer = Layer::new(name.into(), None);
+            self.layers.push(layer);
+            Ok(())
+        }
+    }
+
+    /// Remove a layer.
+    ///
+    /// The default layer cannot be removed.
```
madig, line 116: I'm not seeing anything stopping client code from doing it anyway?

src/ufo.rs
```diff
@@ -271,27 +233,15 @@ impl Font {
...
-                    .drain(..)
-                    .map(|(name, p)| {
-                        let layer_path = path.join(&p);
-                        let layer = Layer::load_impl(&layer_path, &glyph_names)?;
-                        Ok(LayerInfo { name, path: p, layer })
-                    })
-                    .collect();
-                layers_r?
+            let layers = if self.data_request.layers {
+                if matches!(meta.format_version, FormatVersion::V3)
```
madig, line 237: Why not `if meta.format_version == FormatVersion::V3 && ...`?

src/layer.rs
madig: Maybe needs a `iter().skip(1)` or something?

madig:
The idea of putting the default layer permanently into `[0]` is interesting. I'd love to fuzz the API to see if there is a weird corner case that unhinges it! Maybe needs a `debug_assert` in `default_layer(_mut)` etc. to ensure is never dethroned?

madig:
Also, the extra effort in normalizer to fish out the layers to delete is... unfortunate.

src/layer.rs
```diff
@@ -16,34 +16,197 @@ use crate::{Error, Glyph, Plist};
...
+pub(crate) static LAYER_CONTENTS_FILE: &str = "layercontents.plist";
+pub(crate) static DEFAULT_LAYER_NAME: &str = "public.default";
+pub(crate) static DEFAULT_GLYPHS_DIRNAME: &str = "glyphs";
+
+pub type LayerName = Arc<str>;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct LayerSet {
+    // The first layer is always the 'default layer'.
+    layers: Vec<Layer>,
```
madig, line 28: Careful! Storing the layer paths in `Layer` lets users redefine e.g. the default glyph dir path, which the spec says must always be "glyphs". Only the name can be changed at will.

src/layer.rs
```diff
@@ -16,34 +16,197 @@ use crate::{Error, Glyph, Plist};
...
+
 /// A [layer], corresponding to a 'glyphs' directory. Conceptually, a layer
 /// is just a collection of glyphs.
 ///
 /// [layer]: http://unifiedfontobject.org/versions/ufo3/glyphs/
-#[derive(Debug, Clone, Default, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct Layer {
     pub(crate) glyphs: BTreeMap<GlyphName, Arc<Glyph>>,
+    pub name: LayerName,
```
madig, line 158: Doesn't this also let the user rename a layer outside `rename_layer`'s control?

src/layer.rs
```diff
@@ -16,34 +16,197 @@ use crate::{Error, Glyph, Plist};
...
+    pub path: PathBuf,
```
madig, line 159: Careful! Storing the layer paths publicly in `Layer` lets users redefine e.g. the default glyph dir path, which the spec says must always be "glyphs". Only the name can be changed at will.

https://unifiedfontobject.org/versions/ufo3/layercontents.plist/#specification

src/upconversion.rs
cmyr: heh not exactly. I use vim, and it's set to 80 columns, and longer lines wrap in away that makes them hard to read; I had a failing test case here and was distracted trying to read the docs, so I wrapped them to 80 columns. 🤷

src/layer.rs
cmyr: ah good catch, I changed the implementation (we used to track the default layer name instead of just insisting that it is at the front) and this wasn't updated appropriately. 👍

src/layer.rs
cmyr: oops!

src/layer.rs
cmyr: Yea, this is a funny case. Layer names and paths are annoying; there's a real sense in which they aren't part of the layer itself, which was the motivation for the original design.

In any case, I think you're right, this is better as a private var and a getter.

cmyr:
Thanks for the feedback, I agree with everything.

I also agree it's somewhat a shame that the normalizer example is more verbose now, but I don't really know what an alternative API would be that lets us maintain our guarantees. 🤔

src/ufo.rs
cmyr: no compelling reason!

madig:
We could have a `retain` method that silently skips the default layer 🤔

cmyr:
> We could have a `retain` method that silently skips the default layer 🤔

Let's see what some more real-world use cases are, and then we can try and come up with an API that addresses them?

madig (approved):

PR #113, opened 2021-04-30
Add Layer::clear and ::get_or_create

madig:
For easier layer handling.

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -81,7 +81,7 @@ impl LayerSet {
     }
 
     /// Get a mutable reference to a layer, by name, or create it if it doesn't exist.
-    pub fn get_or_create_mut(&mut self, name: &str) -> &mut Layer {
+    pub fn get_or_create(&mut self, name: &str) -> &mut Layer {
         if let Some(index) = self.layers.iter().position(|l| &*l.name == name) {
             self.layers.get_mut(index).unwrap()
         } else {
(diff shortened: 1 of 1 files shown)


src/layer.rs
```diff
@@ -80,6 +80,17 @@ impl LayerSet {
         self.layers.iter_mut().find(|l| &*l.name == name)
     }
 
+    /// Get a mutable reference to a layer, by name, or create it if it doesn't exist.
+    pub fn get_or_create_mut(&mut self, name: &str) -> &mut Layer {
```
cmyr, line 84: Assuming we aren't going to have a non-mut variant I might just call this `get_or_create`?

src/layer.rs
```diff
@@ -80,6 +80,17 @@ impl LayerSet {
...
+        if let Some(index) = self.layers.iter().position(|l| &*l.name == name) {
+            self.layers.get_mut(index).unwrap()
```
cmyr, line 86: maybe we can just do,
```suggestion
        if let Some(layer) = self.layers.iter_mut().find(|l| &*l.name == name) {
            Some(layer)
```

or else,

```rust
self.layers.iter_mut().find(|l| &*l.name == name)
    .or_else(make_the_layer)
```

But maybe this doesn't work because of that borrowchecker thing we discussed?

cmyr (approved):
Looks good, a few comments inline :)

src/layer.rs
madig: I quickly tried

```rust
        self.layers.iter_mut().find(|l| &*l.name == name).or_else(move || {
            let layer = Layer::new(name.into(), None);
            self.layers.push(layer);
            self.layers.last_mut()
        }).unwrap()
```

but ran into the ownership issue again...

PR #114, opened 2021-05-04
Python bindings

cmyr:
I'm not sure that I want to merge this, but I at least wanted to open the PR?

This is 'pynorad', a python library built on top of norad, that emulates the ufoLib2 API. It has enough functionality to run the fontmake sample project at https://github.com/daltonmaag/pipeline-perf-tracker. Performance on that project seems to be very marginally better than ufoLib2, but not enough to seem especially significant.


alerque:
@davelab6 probably because most of the time is being spent doing other things besides reading and parsing the UFO files. For file operations the disk I/O is usually more significant than what language you're triggering the I/O operation from, and in this case once the underlying filesystem operation are done using Python bindings still means the full overhead of Python and its memory structures. More parts of data handling would have to be ported before significant speed gains are evident.

I'm not suggesting this isn't valuable. I tend to think it is valuable because the more parts of the workflows start getting ported over the bigger and bigger chunks can start skipping Python entirely. I have plenty of beefs with Rust, but one design feature is the ability to introduce it gradually to projects — bit by bit, function by function, module by module. This makes it viable to start chipping away at technical debt without the barrier that full rewrites pose.

cmyr:
Additionally, in fontmake very little time is spent in the ufoLib2 code at all, so even if we were significantly faster it still wouldn't move the needle too too much.

cmyr:
I'm going to close this, although I will keep the branch around for reference.

Issue #110, opened 2021-04-21
Implement transformation for glyph contents when kurbo available


madig:
I'd like to be able to do `kurbo::Affine * Contour` or some such, found at least two instances where that would come in handy. Maybe something for returning a transformed copy and for transforming in place?

cmyr:
mm. I was also today running into decomposition, so I do think it makes sense to have a `to_kurbo` method on `Contour`, behind the `kurbo` feature. Does it make more sense to use the norad affine type in your case? I'm not sure, just thinking aloud...

madig:
Then all the affine transformations would need to be impl'd on norad's affine type, no? I thought we should reuse kurbo for that,

simoncozens:
Can we also have a `to_kurbo` method (or even better some kind of Into trait) on `AffineTransform` for components? Currently I'm doing this:

```Rust
    let transform = [
        component.transform.x_scale as f64,
        component.transform.xy_scale as f64,
        component.transform.y_scale as f64,
        component.transform.yx_scale as f64,
        component.transform.x_offset as f64,
        component.transform.y_offset as f64,
    ];
    Affine::new(transform)
```

cmyr:
@simoncozens we've got this already: https://github.com/linebender/norad/blob/master/src/glyph/mod.rs#L556

simoncozens:
Hah, great. Not sure why I couldn’t see it on docs.rs - maybe it doesn’t pick up stuff behind features.

cmyr:
Yes, I think that's true.

PR #116, opened 2021-05-09
Implement to_kurbo on Contour

simoncozens:
This adds a `to_kurbo` method on a Contour, as discussed in #110.

Most of this has been tested in fonttools-rs, apart from the implied on-curve stuff because I can't find any UFOs that do that.

### DIFF
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -289,11 +289,14 @@ impl Contour {
                     offs.clear();
                 }
                 PointType::QCurve => {
-                    while offs.len() > 1 {
-                        let implied_point = offs[0].midpoint(offs[1]);
-                        path.quad_to(offs.pop_front().unwrap(), implied_point);
+                    while let Some(pt) = offs.pop_front() {
+                        if let Some(next) = offs.front() {
+                            let implied_point = pt.midpoint(*next);
+                            path.quad_to(pt, implied_point);
+                        } else {
+                            path.quad_to(pt, kurbo_point);
+                        }
                     }
-                    path.quad_to(offs[0], kurbo_point);
                     offs.clear();
                 }
             }
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -286,7 +286,7 @@ impl Contour {
                         &mut [p1, p2] => path.curve_to(p1, p2, kurbo_point),
                         _ => return Err(ErrorKind::TooManyOffCurves),
                     };
-                    offs.truncate(0);
+                    offs.clear();
                 }
                 PointType::QCurve => {
                     while offs.len() > 1 {
@@ -294,7 +294,7 @@ impl Contour {
                         path.quad_to(offs.pop_front().unwrap(), implied_point);
                     }
                     path.quad_to(offs[0], kurbo_point);
-                    offs.truncate(0);
+                    offs.clear();
                 }
             }
         }
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -253,7 +253,7 @@ impl Contour {
         self.points.first().map_or(true, |v| v.typ != PointType::Move)
     }
 
-    /// Converts the contour to a kurbo::BezPath
+    /// Converts the `Contour` to a [`kurbo::BezPath`].
     #[cfg(feature = "kurbo")]
     pub fn to_kurbo(&self) -> Result<kurbo::BezPath, ErrorKind> {
         let mut path = kurbo::BezPath::new();
@@ -523,7 +523,7 @@ impl ContourPoint {
         self.identifier.replace(id)
     }
 
-    /// Returns a kurbo::Point with the ContourPoint's coordinates.
+    /// Returns a [`kurbo::Point`] with this `ContourPoint`'s coordinates.
     #[cfg(feature = "kurbo")]
     pub fn to_kurbo(&self) -> kurbo::Point {
         kurbo::Point::new(self.x as f64, self.y as f64)
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -258,6 +258,16 @@ impl Contour {
     pub fn to_kurbo(&self) -> Result<kurbo::BezPath, ErrorKind> {
         let mut path = kurbo::BezPath::new();
         let mut offs: VecDeque<kurbo::Point> = VecDeque::new();
+        // Add end-of-contour offcurves to queue
+        let mut i = self.points.len() - 1;
+        while i > 0 && self.points[i].typ == PointType::OffCurve {
+            let kurbo_point = kurbo::Point::new(self.points[i].x as f64, self.points[i].y as f64);
+            offs.push_front(kurbo_point);
+            i -= 1;
+        }
+        if self.is_closed() {
+            path.move_to(kurbo::Point::new(self.points[i].x as f64, self.points[i].y as f64));
+        }
         for pt in &self.points {
             let kurbo_point = kurbo::Point::new(pt.x as f64, pt.y as f64);
             match pt.typ {
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -273,7 +273,7 @@ impl Contour {
         if let Some(start) = points.next() {
             path.move_to(start.to_kurbo());
         }
-        for pt in &self.points {
+        for pt in points {
             let kurbo_point = pt.to_kurbo();
             match pt.typ {
                 PointType::Move => path.move_to(kurbo_point),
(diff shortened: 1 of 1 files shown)


madig:
I also wrote something: https://github.com/linebender/norad/blob/8c9038b/examples/letterspacer.rs#L523-L680. Assembles PathEls instead of calling drawing methods.

src/glyph/mod.rs
```diff
@@ -251,6 +252,49 @@ impl Contour {
...
+    pub fn to_kurbo(&self) -> Result<kurbo::BezPath, ErrorKind> {
+        let mut path = kurbo::BezPath::new();
+        let mut offs: VecDeque<kurbo::Point> = VecDeque::new();
+        // Add end-of-contour offcurves to queue
+        let mut i = self.points.len() - 1;
+        while i > 0 && self.points[i].typ == PointType::OffCurve {
+            let kurbo_point = kurbo::Point::new(self.points[i].x as f64, self.points[i].y as f64);
+            offs.push_front(kurbo_point);
+            i -= 1;
+        }
```
cmyr, line 267: A few thoughts:

- are we okay with encoding the assumption that the start point in a closed path is always the last on-curve in the point list? I think this is a good convention, but it is not enforced in the spec.
- If we're doing that, I would probably try to write this all in terms of an iterator? I try to avoid indexing into collections where I can avoid it, because it is harder to reason about (any invalid index is a runtime panic) and also can have performance costs (if the compiler doesn't know an index is valid, it has to do a bounds check each time).

With that in mind, I might try to do something like,

```rust
let mut points = if self.is_closed() {
    let rotate = self.points.iter().rev().position(|pt| pt.typ != PointType::OffCurve).map(|idx| self.points.len() - 1 - idx);
    // is this an off-by-one? we want to see the first point twice
    self.points.iter().cycle().skip(rotate.unwrap_or(0)).take(self.points.len() + 1)
} else {
   // we need to return the same concrete iterator type from both branches
   self.points.iter().cycle().skip(0).take(self.points.len())
}
if let Some(start) = points.next() {
    path.move_to(start);
}
for pt in points {
// the usual logic
}
```

In any case this is mostly code golf, and definitely a matter of preference, but it has the nice quality that once we get the points in the right order we just build the path out of a single run through the iterator.

src/glyph/mod.rs
```diff
@@ -251,6 +252,49 @@ impl Contour {
...
+        if self.is_closed() {
+            path.move_to(kurbo::Point::new(self.points[i].x as f64, self.points[i].y as f64));
+        }
+        for pt in &self.points {
+            let kurbo_point = kurbo::Point::new(pt.x as f64, pt.y as f64);
```
cmyr, line 272: I'd probably go ahead and add `CountourPoint::to_kurbo` while we're in here?

src/glyph/mod.rs
```diff
@@ -251,6 +252,49 @@ impl Contour {
...
+                PointType::Move => path.move_to(kurbo_point),
+                PointType::Line => path.line_to(kurbo_point),
+                PointType::OffCurve => offs.push_back(kurbo_point),
+                PointType::Curve => {
+                    match offs.len() {
+                        0 => return Err(ErrorKind::BadPoint),
+                        1 => path.quad_to(offs[0], kurbo_point),
+                        2 => path.curve_to(offs[0], offs[1], kurbo_point),
+                        _ => return Err(ErrorKind::TooManyOffCurves),
+                    };
```
cmyr, line 283: A nice rust trick is that you can actually match on a slice, so in this case:
```rust
                    match offs.as_slice() {
                        &[] => return Err(ErrorKind::BadPoint),
                        &[p1] => path.quad_to(p1, kurbo_point),
                        &[p1, p2] => path.curve_to(p1, p2, kurbo_point),
                        _ => return Err(ErrorKind::TooManyOffCurves),
                    };
```

src/glyph/mod.rs
```diff
@@ -251,6 +252,49 @@ impl Contour {
...
+                    offs.truncate(0);
```
cmyr, line 284: this could also be `offs.clear()`.

src/glyph/mod.rs
```diff
@@ -251,6 +252,49 @@ impl Contour {
...
+                }
+                PointType::QCurve => {
+                    while offs.len() > 1 {
+                        let implied_point = offs[0].midpoint(offs[1]);
+                        path.quad_to(offs.pop_front().unwrap(), implied_point);
+                    }
+                    path.quad_to(offs[0], kurbo_point);
+                    offs.truncate(0);
```
cmyr, line 292: again I've developed an allergy to indexing, and would personally probably do something like:

```rust
while let Some(pt) = offs.pop_front() {
    if let Some(next) = offs.front() {
        let implied_point = pt.midpoint(*next);
        path.quad_to(pt, implied_point);
    } else {
        path.quad_to(pt, kurbo_point);
    }
}
```

src/glyph/mod.rs
```diff
@@ -251,6 +252,49 @@ impl Contour {
     fn is_closed(&self) -> bool {
         self.points.first().map_or(true, |v| v.typ != PointType::Move)
     }
+
+    /// Converts the contour to a kurbo::BezPath
+    #[cfg(feature = "kurbo")]
+    pub fn to_kurbo(&self) -> Result<kurbo::BezPath, ErrorKind> {
```
cmyr, line 258: This doesn't need to be fixed here, but I want to discourage using `ErrorKind` as an actual error type; it's intended more to indicate the cause of some other error type, such as a parsing error or a validation error. I believe we do use it like this elsewhere in the code, but I'd like to rectify that at some point.

cmyr (commented):
Sorry I think my coffee just kicked in, so there's way more feedback here than is realistically necessary. Feel free to ignore most of it at your discretion; I thought it might be helpful to see how I would think through some of these problems in a slightly more idiomatic way. :)


I'll give you a chance to respond, but I'm happy to merge this more-or-less as is.

simoncozens:
This is fantastic feedback, and will really help me think about idiomatic Rust. :-) Thank you! I shall work on addressing the issues above.

src/glyph/mod.rs
simoncozens: Ah! I tried to do something like this but it didn't quite work, so I gave up. The `path.<operator>_to` methods take anything which is `Into<kurbo::Point>`, so I thought it would be much neater to implement that trait for `ContourPoint`. But I guess that consumes the point, and I ended up in a fight with the borrow checker. `to_kurbo`, making a copy, makes sense.

src/glyph/mod.rs
simoncozens: I'm going to leave this part for now, because I don't think there's a better error type right now - things which happen when *using* the curve are not exactly parsing, writing, or spec errors -  and I think I'll leave you to design the error hierarchy you want to see.

cmyr: yep that's fine!

cmyr (approved):
Cool, thanks for revisiting this @simoncozens. I'm going to make a couple of doc tweaks and then merge, and I'll look into the error types before releasing 0.4.0.


Another high-level idiomatic-rust thought, here: I think the ideal implementation of this method would return an `impl Iterator<Item=kurbo::PathEl>`, instead of a `kurbo::BezPath`. A `BezPath` is just a `Vec<PathEl>` under the hood, and by having this return an iterator, you could do fun things like take the elements and pass them directly to some drawing code, or accumulate multiple contours into a single `BezPath`, all without needing any intermediate allocation.

In order to fully avoid allocation, the other thing we would need to get rid of is the `VecDeque`. For this I would probably add some custom helper type, like:

```rust
enum OffCurves {
   None,
   One(Point),
   Two(Point, Point),
}

impl OffCurves {
    fn points(&self) -> &[Point];
    fn push(&mut self, point: Point) -> Result<(), ErrorKind>;
    fn clear(&mut self);
}
```

I may go ahead and play around with this after merging, because I think it would be fun 😬

src/glyph/mod.rs
```diff
@@ -6,6 +6,7 @@ mod serialize;
 #[cfg(test)]
 mod tests;
 
+use std::collections::VecDeque;
```
cmyr, line 9:
```suggestion
```

src/glyph/mod.rs
```diff
@@ -251,6 +252,57 @@ impl Contour {
     fn is_closed(&self) -> bool {
         self.points.first().map_or(true, |v| v.typ != PointType::Move)
     }
+
+    /// Converts the `Contour` to a [`kurbo::BezPath`].
+    #[cfg(feature = "kurbo")]
+    pub fn to_kurbo(&self) -> Result<kurbo::BezPath, ErrorKind> {
+        let mut path = kurbo::BezPath::new();
+        let mut offs: VecDeque<kurbo::Point> = VecDeque::new();
```
cmyr, line 260:
```suggestion
        let mut offs = std::collections::VecDeque::new();
```

PR #121, opened 2021-05-11
Layer API improvements and fixups

cmyr:
- Added LayerSet::new
- Added Layer::rename_glyph
- Fixed bug where get_glyph_mut used Arc::get_mut instead of Arc::make_mut
- Renamed get_glyph to glyph
- Renamed get_glyph_mut to glyph_mut
- Renamed iter_contents & iter_contents_mut -> iter & iter_mut
- Removed deprecated method

A few other things along these lines

----

this is also split off from #114

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -67,6 +67,15 @@ impl LayerSet {
         Ok(LayerSet { layers })
     }
 
+    /// Create a new `LayerSet`.
+    ///
+    /// Will panic if `layers` is empty.
+    pub fn new(mut layers: Vec<Layer>) -> Self {
+        assert!(!layers.is_empty());
+        layers.first_mut().unwrap().path = DEFAULT_GLYPHS_DIRNAME.into();
+        LayerSet { layers }
+    }
+
     /// The number of layers in the set.
     ///
     /// This should be non-zero.
@@ -329,7 +338,7 @@ impl Layer {
         Ok(())
     }
 
-    /// The number of glyphs in this layer.
+    /// The number of [`Glyph`]s in the layer.
     pub fn len(&self) -> usize {
         self.glyphs.len()
     }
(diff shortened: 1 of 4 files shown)


src/layer.rs
```diff
@@ -63,6 +63,15 @@ impl LayerSet {
         Ok(LayerSet { layers })
     }
 
+    /// Create a new `LayerSet`.
+    ///
+    /// Will panic if `layers` is empty.
+    pub fn new(mut layers: Vec<Layer>) -> Self {
+        assert!(!layers.is_empty());
```
madig, line 70: This irks me, but I also find Result here less ergonomic :(

src/layer.rs
```diff
@@ -350,7 +359,7 @@ impl Layer {
     }
 
     /// Returns a reference the glyph with the given name, if it exists.
-    pub fn get_glyph<K>(&self, glyph: &K) -> Option<&Arc<Glyph>>
+    pub fn glyph<K>(&self, glyph: &K) -> Option<&Arc<Glyph>>
```
madig, line 362: Wouldn't `get` be more appropriate? Layers are basically HashMaps with fixed types...

src/layer.rs
```diff
@@ -359,14 +368,36 @@ impl Layer {
...
     where
         GlyphName: Borrow<K>,
         K: Ord + ?Sized,
     {
-        self.glyphs.get_mut(glyph).and_then(|g| Arc::get_mut(g))
+        self.glyph_mut(glyph)
     }
 
+    /// Returns a mutable reference to the glyph with the given name, if it exists.
+
```
madig, line 400: Something's missing here?

src/layer.rs
cmyr: This isn't a method we would really expect people to use except in very special cases, so I'm okay with expecting a higher degree of caution.

src/layer.rs
cmyr: I think there are enough other things hanging of a layer that just 'get' would be unclear; it could be the layer's name, or it's path, or something from the lib...

src/layer.rs
cmyr: Oops yes this was a relic of the deprecations. Nice catch!

src/layer.rs
madig: Then I actually advocate keeping "get_glyph" because it is clear. Otherwise, I'll bite my tongue now.

Issue #125, opened 2021-05-12
Glif: serialize guidelines after anchors


madig:
ufonormalizer does it.

Issue #105, opened 2021-04-08
glif serialization: advance of zero should result in no <advance/> element


madig:
Also, empty outline can be serialized without `<outline></outline>`.

PR #126, opened 2021-05-12
Serialization normalizations

madig:
Move more closely to ufonormalizer output.

I'm a bit mystified by our custom lib writing code. Why does the lib section end up unordered? It should be sorted alphabetically and I thought that's what plist was doing.

Closes https://github.com/linebender/norad/issues/125
Closes https://github.com/linebender/norad/issues/105

### DIFF
--- a/src/glyph/serialize.rs
+++ b/src/glyph/serialize.rs
@@ -32,14 +32,17 @@ impl Glyph {
             writer.write_event(char_to_event(*codepoint))?;
         }
 
-        let mut start = BytesStart::borrowed_name(b"advance");
-        if self.width != 0. {
-            start.push_attribute(("width", self.width.to_string().as_str()));
+        // Skip serializing advance if both values are zero, infinite, subnormal, or NaN.
+        if self.width.is_normal() || self.height.is_normal() {
+            let mut start = BytesStart::borrowed_name(b"advance");
+            if self.width != 0. {
+                start.push_attribute(("width", self.width.to_string().as_str()));
+            }
+            if self.height != 0. {
+                start.push_attribute(("height", self.height.to_string().as_str()));
+            }
+            writer.write_event(Event::Empty(start))?;
         }
-        if self.height != 0. {
-            start.push_attribute(("height", self.height.to_string().as_str()));
-        }
-        writer.write_event(Event::Empty(start))?;
 
         if let Some(ref note) = self.note {
             writer.write_event(Event::Start(BytesStart::borrowed_name(b"note")))?;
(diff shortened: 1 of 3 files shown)


cmyr:
oh no! I hadn't noticed this 😆

src/glyph/serialize.rs
```diff
@@ -33,41 +38,42 @@ impl Glyph {
         }
 
         let mut start = BytesStart::borrowed_name(b"advance");
-        if self.width != 0. {
-            start.push_attribute(("width", self.width.to_string().as_str()));
-        }
+        let mut have_advance = false;
```
cmyr, line 41: see my patch, I think it's simpler to just check both width & height first, and put everything else in a conditional block?

src/error.rs
```diff
@@ -73,6 +73,7 @@ pub enum WriteError {
     /// If for some reason the implementation of that crate changes, we could
     /// be affected, although this is very unlikely.
     InternalLibWriteError,
+    InvalidUtf8Error(FromUtf8Error),
```
cmyr, line 76: This should never happen, and I think a panic is better than an error. We are explicitly serializing UTF-8, and a failure to do that is not recoverable.

src/glyph/serialize.rs
```diff
@@ -33,41 +38,42 @@ impl Glyph {
...
-            writer.write_event(guide.to_event())?;
-        }
-
-        writer.write_event(Event::Start(BytesStart::borrowed_name(b"outline")))?;
-        for contour in &self.contours {
-            contour.write_xml(&mut writer)?;
-        }
-        for component in &self.components {
-            writer.write_event(component.to_event())?;
+        if !self.contours.is_empty() || !self.components.is_empty() {
```
cmyr, line 58: yea, like this 👍

src/glyph/serialize.rs
```diff
@@ -84,6 +90,12 @@ impl Glyph {
             write_lib_section(&self.lib, &mut writer)?;
         }
 
+        if let Some(ref note) = self.note {
```
cmyr, line 93: cool, my patch doesn't include this

src/glyph/serialize.rs
```diff
@@ -106,7 +118,7 @@ fn write_lib_section<T: Write>(lib: &Plist, writer: &mut Writer<T>) -> Result<()
     let as_value: plist::Value = lib.to_owned().into();
     let mut out_buffer = Vec::with_capacity(256); // a reasonable min size?
     as_value.to_writer_xml(&mut out_buffer)?;
-    let lib_xml = String::from_utf8(out_buffer).expect("xml writer writs valid utf8");
+    let lib_xml = String::from_utf8(out_buffer).map_err(WriteError::InvalidUtf8Error)?;
```
cmyr, line 121: echoing above, this really is a firm expectation.

cmyr (commented):
Cool, this all looks sensible. Maybe take a look at my patches and see if there's anything interesting in there, to you?

I'll take a look at this lib-ordering business. Do you know if it's recursive? That is, if the lib contains nested dictionaries, is everything sorted during deserialization?

madig:
> I'll take a look at this lib-ordering business. Do you know if it's recursive? That is, if the lib contains nested dictionaries, is everything sorted during deserialization?

I think dictionary keys are always sorted inside a lib :thinking:

madig:
> Cool, this all looks sensible. Maybe take a look at my patches and see if there's anything interesting in there, to you?

I copied your `is_normal` approach to advance widths. The tests I'll leave for now I think, once the lib order stuff is taken care of, my glif roundtrip test should test the exact order and formatting of all glif elements and attributes.

cmyr (approved):
Cool, yea, let's keep the test because I will work on getting it passing.

PR #130, opened 2021-05-12
Serialize less stuff

### LINKED ISSUE #105: glif serialization: advance of zero should result in no <advance/> element

cmyr:
With this patch we no longer serialize zero advance values, and we don't serialize empty outlines.

- closes #105


madig:
https://github.com/linebender/norad/pull/126 :O

PR #132, opened 2021-05-12
More extensive to_kurbo

madig:
Based on https://github.com/linebender/norad/blob/8c9038b/examples/letterspacer.rs#L523-L680, which is based on the fontTools' point pen to pen adapter pen.

This handles some more corner cases like e.g. converting successive curves without off-curves to lines, and all-off-curve paths.

I'll hold off of implementing this on `Component` and `Glyph` until I have a use-case.

### DIFF
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -411,8 +411,7 @@ impl Contour {
                     [.., cn] => {
                         // Insert a computed on-curve point between each control point.
                         for (c0, c1) in controls.iter().zip(controls.iter().skip(1)) {
-                            let p1 = Point::new(0.5 * (c0.x + c1.x), 0.5 * (c0.y + c1.y));
-                            segments.push(PathEl::QuadTo(*c0, p1));
+                            segments.push(PathEl::QuadTo(*c0, c0.midpoint(*c1)));
                         }
                         segments.push(PathEl::QuadTo(cn, p));
                         controls.clear()
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -896,11 +896,10 @@ impl From<Color> for druid::piet::Color {
     }
 }
 
-#[cfg(test)]
-mod tests2 {
+#[cfg(all(test, feature = "kurbo"))]
+mod kurbo_tests {
     use super::*;
 
-    #[cfg(feature = "kurbo")]
     #[test]
     fn many_control_quads() {
         let c1 = Contour::new(
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -328,17 +328,17 @@ impl Contour {
         let implied_oncurve: ContourPoint;
 
         // Phase 1: Preparation
-        match points.as_slice() {
+        match *points.as_slice() {
             // Empty contours cannot be represented by segments.
-            &[] => return Ok(segments),
+            [] => return Ok(segments),
             // Single points are converted to open MoveTos because closed single points of any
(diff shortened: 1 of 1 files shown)


madig:
The ErrorKinds used here are really UFO specification violations. Maybe they can get their own enum later.

src/glyph/mod.rs
```diff
@@ -258,55 +258,137 @@ impl Contour {
...
-                PointType::OffCurve => offs.push_back(kurbo_point),
-                PointType::Curve => {
-                    match offs.make_contiguous() {
-                        [] => return Err(Error::ConvertContour(ErrorKind::BadPoint)),
-                        [p1] => path.quad_to(*p1, kurbo_point),
-                        [p1, p2] => path.curve_to(*p1, *p2, kurbo_point),
-                        _ => return Err(Error::ConvertContour(ErrorKind::TooManyOffCurves)),
-                    };
-                    offs.clear();
+    pub fn to_kurbo(&self) -> Result<Vec<kurbo::PathEl>, Error> {
```
cmyr, line 263: I would have many of the same comments on this patch as I had on #116, it might be worth reading through those?

madig:
@cmyr i got the following, converting a QCurve with some off-curves:

```rust
let c1 = Contour::new(
    vec![
        ContourPoint::new(0.0, 0.0, PointType::OffCurve, false, None, None, None),
        ContourPoint::new(2.0, 2.0, PointType::OffCurve, false, None, None, None),
        ContourPoint::new(4.0, 4.0, PointType::OffCurve, false, None, None, None),
        ContourPoint::new(100.0, 100.0, PointType::QCurve, false, None, None, None),
    ],
    None,
    None,
);

assert_eq!(
    c1.to_kurbo().unwrap(),
    vec![
        kurbo::PathEl::MoveTo((100.0, 100.0).into()),
        kurbo::PathEl::QuadTo((0.0, 0.0).into(), (1.0, 1.0).into(),),
        kurbo::PathEl::QuadTo((2.0, 2.0).into(), (3.0, 3.0).into(),),
        kurbo::PathEl::QuadTo((4.0, 4.0).into(), (100.0, 100.0).into(),),
        kurbo::PathEl::ClosePath,
    ]
);
```

Does the `PathEl::MoveTo((100.0, 100.0))` as the first element strike you as right or should it be `(0.0, 0.0)`? I copied this behavior from fontTools.

anthrotype:
what's the status of this PR? Are you planning to finish this at some point?

madig:
One day. I think I needed to validate that the decomposer actually does what kurbo expects, i.e. is the code right to put the `MoveTo(100.0, 100.0)` before the `QuadTo((0.0, 0.0), (1.0, 1.0))` in https://github.com/linebender/norad/pull/132/files#diff-19795d9bd46090f88223ed97fc7eecb8758e48ce418c4b2dbed4f57bf7814c15R919-R923 or is that just how fontTools stuff does it?

anthrotype:
I think that's correct, `c1` is a closed path, so the last on-curve point (100,100) becomes the first MoveTo

anthrotype:
(also whatever fonttools does is as normative (if not more) in UFO world as the ufo spec itself)

madig:
Yes, but I don't know if kurbo wants something different. @cmyr?

anthrotype:
why would kurbo ever want anything different? those conventions for converting between GLIF points to segment-oriented APIs like fonttools pens are well in place since ages and work, I don't see why diverging from them now

PR #133, opened 2021-05-13
Serialize height before width

cmyr:
Another little thing we were doing differently form ufo normalizer.


madig:
Argh, I wanted to fix that one but forgot about it :o

madig (approved):

PR #134, opened 2021-05-13
Overwrite meta.creator on save

### LINKED ISSUE #123: Unconditionally overwrite `meta.creator` on writing
Since the `creator` value is meant to represent the writer of the UFO, norad should always write down itself there. I don#t think it is necessary anymore to refuse writing out UFOs not created in norad.

Annoying edge case to think about: `save()` should not modify self, so it would ignore what the user put into meta. This is slightly iffy.

cmyr:
- closes #123

### DIFF
--- a/src/ufo.rs
+++ b/src/ufo.rs
@@ -316,7 +316,14 @@ impl Font {
             fs::remove_dir_all(path)?;
         }
         fs::create_dir(path)?;
-        plist::to_file_xml(path.join(METAINFO_FILE), &self.meta)?;
+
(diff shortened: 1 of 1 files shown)


madig:
Wait, what does the if-else do exactly?

cmyr:
yes it's maybe useless? if we are already set as the creator, then it just uses the existing metadata, otherwise it creates new metadata with us as the creator.

madig (approved):
Oh right, ok.

PR #135, opened 2021-05-16
Add transform() method to ContourPoint

simoncozens:
This applies an `AffineTransform` matrix to a point. This turns out to be very useful when decomposing mixed contour/component glyphs prior to converting to TTF

### DIFF
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -535,6 +535,14 @@ impl ContourPoint {
     pub fn to_kurbo(&self) -> kurbo::Point {
         kurbo::Point::new(self.x as f64, self.y as f64)
     }
+
+    /// Applies a transformation matrix to the point's coordinates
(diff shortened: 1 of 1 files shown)


madig:
Cool, what code do you decompose with? I wrote https://github.com/linebender/norad/blob/add-example-letterspacer/examples/letterspacer.rs#L474-L521 and ctrlcctrlv has something that deals with cycles: https://github.com/MFEK/glifparser.rlib/blob/2f6a6e8ad788ca56f892863db17624943f78df11/src/component.rs#L175.

simoncozens:
Oh, those are much better than mine. I don't flatten nested components.

It would be good to add some kind of decomposition method to `norad::Glyph`.

madig:
I plan on doing that for https://github.com/linebender/norad/pull/132, but haven't had time yet.

cmyr (approved):
Looks good!

PR #136, opened 2021-05-20
get_path is actually pretty handy, don't hide it

simoncozens:
This exposes the `get_path` function to get the path of a glif. This is useful when passing the glif to utilities which work on a .glif object, such as the MFEK stuff.

And in general, I'm a fan of giving the developer access to any information we hold that they claim to want, rather than questioning their motives. ;-)

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -454,6 +454,9 @@ impl Layer {
         self.glyphs.values_mut().map(Arc::make_mut)
     }
 
+    /// Returns the path to the .glif file of a given glyph name.
+    ///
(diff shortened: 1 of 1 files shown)


cmyr:
Yep this makes sense. While you're in here, can we rename it to just `path`?

madig:
Then we have to rename get_glyph, too :O

cmyr:
Oh I thought this was a method on `Glyph`, not on `Layer`, `get_path` is fine on `Layer`. Thanks @madig!

PR #139, opened 2021-06-27
Minor refactoring

madig:
Moving some code around to make files shorter and subdivide functions.

De-anonymizing `load_impl` drops 130 Kb from the release build of open_ufo.

### DIFF
--- a/src/groups.rs
+++ b/src/groups.rs
@@ -3,8 +3,9 @@ use std::collections::{BTreeMap, HashSet};
 use crate::error::GroupsValidationError;
 use crate::GlyphName;
 
-/// Groups is a map of group name to a list of glyph names. It's a BTreeMap because we need sorting
-/// for serialization.
+/// A map of group name to a list of glyph names.
+///
+/// We use a BTreeMap because we need sorting for serialization.
 pub type Groups = BTreeMap<String, Vec<GlyphName>>;
 
 /// Validate the contents of the groups.plist file according to the rules in the
--- a/src/groups.rs
+++ b/src/groups.rs
@@ -1,0 +1,50 @@
+use std::collections::{BTreeMap, HashSet};
+
+use crate::error::GroupsValidationError;
+use crate::GlyphName;
+
+/// Groups is a map of group name to a list of glyph names. It's a BTreeMap because we need sorting
+/// for serialization.
+pub type Groups = BTreeMap<String, Vec<GlyphName>>;
+
+/// Validate the contents of the groups.plist file according to the rules in the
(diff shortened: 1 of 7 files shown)


cmyr:
Ugh sorry I didn't see this until after #141 and #140. Those changes are small, but probably collide with some of this?

madig:
Well, I can rebase :grimacing:

cmyr (approved):
I agree that breaking this up is a good idea, thanks!

src/groups.rs
```diff
@@ -0,0 +1,50 @@
+use std::collections::{BTreeMap, HashSet};
+
+use crate::error::GroupsValidationError;
+use crate::GlyphName;
+
+/// Groups is a map of group name to a list of glyph names. It's a BTreeMap because we need sorting
+/// for serialization.
```
cmyr, line 7:
```suggestion
/// A map of group name to a list of glyph names.
///
/// We use a BTreeMap because we need sorting for serialization.
```

src/kerning.rs
```diff
@@ -0,0 +1,90 @@
+use std::collections::BTreeMap;
+
+use serde::ser::{SerializeMap, Serializer};
+use serde::Serialize;
+
+/// Kerning is a map of first half of a kerning pair (glyph name or group name) to the second half
+/// of a pair (glyph name or group name), which maps to the kerning value (high-level view:
+/// (first, second) => value). It's a BTreeMap because we need sorting for serialization.
+pub type Kerning = BTreeMap<String, BTreeMap<String, f32>>;
```
cmyr, line 9:
```suggestion
/// A map of kerning pairs.
///
/// This is represented as a map of first half of a kerning pair (glyph name or group name)
/// to the second half of a pair (glyph name or group name), which maps to the kerning value
/// (high-level view: (first, second) => value).
///
/// We use a `BTreeMap` because we need sorting for serialization.
pub type Kerning = BTreeMap<String, BTreeMap<String, f32>>;
```

src/kerning.rs
```diff
@@ -0,0 +1,90 @@
...
+
+/// KerningSerializer is a crutch to serialize kerning values as integers if they are
+/// integers rather than floats. This spares us having to use a wrapper type like
+/// IntegerOrFloat for kerning values.
```
cmyr, line 13:
```suggestion
/// A helper for serializing kerning values.
///
/// KerningSerializer is a crutch to serialize kerning values as integers if they are
/// integers rather than floats. This spares us having to use a wrapper type like
/// `IntegerOrFloat` for kerning values.
```

cmyr (approved):
Couldn't stop my self from offering a few doc tweaks, this looks good though!

PR #141, opened 2021-06-28
Fixup load api

### LINKED ISSUE #138: Font::load <-> Font::load_ufo?
What's the difference?

cmyr:
This is mainly intended to clear up confusion around how to use
load_ufo; the answer is "don't".

I believe that the previous API of splitting up the creation of the font
from the loading was unnecessary. I also suspect that having the Font
object hold on to the DataRequest is unnecessary, although there are
possible arguments in favor of this approach.

In any case, the main confusion with the previous API is that loading
with a data request took `&self` by reference but returned a new font
object, which is confusing. I would expect it either to take a
DataRequest as an argument to an associated function, or I would expect
it to take `&mut self` and load the contents of the file into the
current font?

@ctrlcctrlv Does this change work for you? Do you ever do anything where you create a `Font` and then do some other stuff before calling `load_ufo`? My medium-term vision would be to maybe remove the `data_request` field on `Font`; alternatively we could keep it, but only exposed immutably (via a getter?) my thinking is that mutating the request after the font is loaded doesn't do anything, but could confuse the user?

closes #138

### DIFF
--- a/src/font.rs
+++ b/src/font.rs
@@ -163,35 +163,47 @@ impl Font {
     }
 
     /// Create a new `Font` only with certain fields.
+    #[doc(hidden)]
+    #[deprecated(
+        since = "0.4.1",
+        note = "To load only specific fields, use Font::load_requestd_data"
+    )]
     pub fn with_fields(data_request: DataRequest) -> Self {
         let mut ufo = Self::new();
         ufo.data_request = data_request;
         ufo
     }
 
-    /// Attempt to load a font object from a file. `path` must point to
-    /// a directory with the structure described in [v3 of the Unified Font Object][v3]
-    /// spec.
+    /// Attempt to load a font object from a file.
     ///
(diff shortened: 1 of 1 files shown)


madig:
I don't think we need to keep request around, except when we eventually (?) implement lazy loading?

madig:
@cmyr How bout making this 

```rust
pub fn load(path: &dyn AsRef<Path>) -> Result<Font, Error> {
    Self::load_requested_data(path, DataRequest::default())
}

pub fn load_requested_data(
    path: &dyn AsRef<Path>,
    request: DataRequest,
) -> Result<Font, Error> {
    // ...
}
```

Or something so we don't have to have an inner impl function to aoid monomorphization?

cmyr:
Request was added in #53 because it was needed by @ctrlcctrlv, so I don't anticipate removing it.

I'd be happy to look at various patches to remove that inner function and see how they look?

cmyr:
I'm going to merge this as-is, and if fred has any comments we can always revise it later.

PR #143, opened 2021-07-20
Add UFO dir and mandatory .plist file read path existence checks and invalid path diagnostics

chrissimpkins:
Current formatted error:

```
UFO load failed with error: PlistError(Error { inner: ErrorImpl { kind: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" }), file_position: None } })
```

New formatted error with changes in this PR:

```
UFO load failed with error: invalid path: "bogus/path"
```

Tested in this executable with argument `bogus/path`:

```rust
fn main() {
    for arg in std::env::args().skip(1) {
        let mut ufo = match norad::Font::load(&arg) {
            Ok(u) => u,
            Err(e) => {
                eprintln!("UFO load failed with error: {}", e);
                std::process::exit(1);
            }
        };
    }
}
```

### DIFF
--- a/src/font.rs
+++ b/src/font.rs
@@ -119,7 +119,7 @@ impl Font {
 
     fn load_impl(path: &Path, request: DataRequest) -> Result<Font, Error> {
         if !path.exists() {
-            let msg = format!("invalid path: {:?}", &path);
+            let msg = format!("invalid path: {:?}", path);
             return Err(std::io::Error::new(std::io::ErrorKind::NotFound, msg).into());
         }
 
--- a/src/font.rs
+++ b/src/font.rs
@@ -118,6 +118,11 @@ impl Font {
     }
 
     fn load_impl(path: &Path, request: DataRequest) -> Result<Font, Error> {
+        if !path.exists() {
+            let msg = format!("invalid path: {:?}", &path);
+            return Err(std::io::Error::new(std::io::ErrorKind::NotFound, msg).into());
+        }
+
         let meta_path = path.join(METAINFO_FILE);
         let mut meta: MetaInfo = plist::from_file(meta_path)?;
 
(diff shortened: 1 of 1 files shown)


madig:
Ah, yes, I'm interested in better error messages. I think this can be done more elegantly maybe, though. One could match on `plist::from__file` and if it returns an `Io` error, return that one instead of the generic `PlistError`. Not sure if one can implement a more user friendly user-facing error string on it though 🤔

src/font.rs
```diff
@@ -118,6 +118,11 @@ impl Font {
     }
 
     fn load_impl(path: &Path, request: DataRequest) -> Result<Font, Error> {
+        if !path.exists() {
+            let msg = format!("invalid path: {:?}", path);
+            return Err(std::io::Error::new(std::io::ErrorKind::NotFound, msg).into());
+        }
+
         let meta_path = path.join(METAINFO_FILE);
         let mut meta: MetaInfo = plist::from_file(meta_path)?;
```
chrissimpkins, line 127: You are suggesting to use the first mandatory plist file read at this level and propagate the plist lib read error, possibly modified with path data?

madig: D'oh, wait, I think I misread the function. Doing a path check here is fine, but I think it should be done for all file reads. Mh.

chrissimpkins: It looks like there is auto-UFO version up-conversion?  If so, should we add path checks for the cross-version mandatory files:

- metainfo.plist
- glyphs/contents.plist

and not check layercontents.plist in case this is a UFO < v3 source read?  Or check the previously read metainfo data to determine the UFO version and validate mandatory files at this stage?  This PR was initially intended to address typos in source dir paths, but I'd be happy to add some read side spec validations if this seems appropriate here and isn't supported elsewhere in the library.

madig: Yes we shoot for the stars because no one has time for UFO < v3. The idea is you read files as required by the version in question and then upgrade at the end. What specifically would you do?

chrissimpkins: Maybe something along these lines?

d9219beeac3eb9de25a63da8963886aa3e6cd1cd

cmyr (approved):
Thanks chris, improving diagnostics is always helpful!

cmyr:
@madig do you have further thoughts on this? I'm happy to merge as-is, I think there are other possible approaches but this one feels reasonable. 🤷

cmyr:
Thinking about this a bit more, what I would love is a new `MissingFile` error variant that specifically handles the case of a missing mandatory file; it can have a field where you pass a string representing the file in question ("metainfo.plist", "background/contents.plist"). So maybe a PR that adds that variant along with the check for `metainfo.plist`, using that variant, would be a good start?

chrissimpkins:
Wrap the UFO dir check into that error handling approach and close this PR?

chrissimpkins:
> Thinking about this a bit more, what I would love is a new `MissingFile` error variant that specifically handles the case of a missing mandatory file; it can have a field where you pass a string representing the file in question ("metainfo.plist", "background/contents.plist"). So maybe a PR that adds that variant along with the check for `metainfo.plist`, using that variant, would be a good start?

How about this approach instead? https://github.com/linebender/norad/pull/146

src/font.rs
chrissimpkins: https://github.com/linebender/norad/pull/146

PR #144, opened 2021-07-21
Add new Font struct public API methods for lib keys, group names, kerning names, layer names

chrissimpkins:
This PR includes the following new `Font` public API methods for a semi-consistent "top level" approach to access lib keys, group names, kerning names, and layer names as Vector collections and iterators over String'ish types:

- `Font::lib_keys()`
- `Font::iter_lib_keys()`
- `Font::group_names()`
- `Font::iter_group_names()`
- `Font::kerning_names()`
- `Font::iter_kerning_names()`
- `Font::layer_names()`
- `Font::iter_layer_names()`

### DIFF
--- a/src/font.rs
+++ b/src/font.rs
@@ -353,6 +353,70 @@ impl Font {
             .guidelines
             .get_or_insert_with(Default::default)
     }
+
+    /// Returns a vector of immutable lib key string references
+    pub fn lib_keys(&self) -> Vec<&String> {
+        self.lib.keys().collect::<Vec<&String>>()
+    }
+
+    /// Returns an iterator over immutable lib key string references
+    pub fn iter_lib_keys(&self) -> impl Iterator<Item = &String> {
+        self.lib.keys()
+    }
+
+    /// Returns a vector of immutable group name string references
+    pub fn group_names(&self) -> Vec<&String> {
+        match &self.groups {
+            Some(g) => g.keys().collect::<Vec<&String>>(),
+            None => vec![],
+        }
+    }
+
+    /// Returns an iterator over immutable group name string references
+    pub fn iter_group_names(&self) -> impl Iterator<Item = &String> {
+        match &self.groups {
+            Some(g) => {
+                let mut groupvec = vec![];
+                for groupname in g.keys() {
+                    groupvec.push(groupname)
+                }
+                groupvec.into_iter()
+            }
+            None => vec![].into_iter(),
+        }
+    }
+
+    /// Returns a vector of immutable kerning glyph name string references
+    pub fn kerning_names(&self) -> Vec<&String> {
+        match &self.kerning {
+            Some(k) => k.keys().collect::<Vec<&String>>(),
+            None => vec![],
+        }
+    }
+
+    /// Returns an iterator over immutable kerning glyph name string references
+    pub fn iter_kerning_names(&self) -> impl Iterator<Item = &String> {
+        match &self.kerning {
+            Some(k) => {
+                let mut kerningvec = vec![];
+                for kerningname in k.keys() {
+                    kerningvec.push(kerningname)
+                }
+                kerningvec.into_iter()
+            }
+            None => vec![].into_iter(),
+        }
+    }
+
+    /// Returns a vector of layer name String types
+    pub fn layer_names(&self) -> Vec<String> {
+        self.layers.names().map(|s| s.to_string()).collect::<Vec<String>>()
+    }
+
+    /// Returns an iterator over layer name String types
+    pub fn iter_layer_names(&self) -> impl Iterator<Item = String> + '_ {
+        self.layers.names().map(|s| s.to_string())
+    }
 }
 
 fn load_lib(lib_path: &Path) -> Result<plist::Dictionary, Error> {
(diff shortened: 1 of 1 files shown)


madig:
I have the feeling that this can be done without adding top-level functions for things 🤔 Do you have user code for this somewhere?

src/font.rs
```diff
@@ -353,6 +353,70 @@ impl Font {
             .guidelines
             .get_or_insert_with(Default::default)
     }
+
+    /// Returns a vector of immutable lib key string references
+    pub fn lib_keys(&self) -> Vec<&String> {
```
cmyr, line 358: I don't think there's any real value in returning a `Vec` here. I would have a single function that returns an iterator, and then the user is free to make a vec if they want.

src/font.rs
```diff
@@ -353,6 +353,70 @@ impl Font {
...
+        self.lib.keys().collect::<Vec<&String>>()
+    }
+
+    /// Returns an iterator over immutable lib key string references
+    pub fn iter_lib_keys(&self) -> impl Iterator<Item = &String> {
+        self.lib.keys()
```
cmyr, line 364: Instead of `&String`, I would prefer this return `&str`. This is not that important in practice, but is more idiomatic, and may let the caller avoid needing to explicitly convert to a `&str` on the calling side.
```suggestion
    pub fn iter_lib_keys(&self) -> impl Iterator<Item = &str> {
        self.lib.keys().map(String::as_ref)
```

I'd also just call this `lib_keys`.

src/font.rs
```diff
@@ -353,6 +353,70 @@ impl Font {
...
+    /// Returns a vector of immutable group name string references
+    pub fn group_names(&self) -> Vec<&String> {
+        match &self.groups {
+            Some(g) => g.keys().collect::<Vec<&String>>(),
+            None => vec![],
+        }
+    }
+
+    /// Returns an iterator over immutable group name string references
+    pub fn iter_group_names(&self) -> impl Iterator<Item = &String> {
```
cmyr, line 376: ditto above about only having the iterating version of this function, and returning `&str`.

src/font.rs
```diff
@@ -353,6 +353,70 @@ impl Font {
...
+        match &self.groups {
+            Some(g) => {
+                let mut groupvec = vec![];
+                for groupname in g.keys() {
+                    groupvec.push(groupname)
+                }
+                groupvec.into_iter()
+            }
+            None => vec![].into_iter(),
+        }
```
cmyr, line 386: It isn't very obvious, but there's a reasonable one-liner here
```suggestion
        self.groups.as_ref().into_iter().flat_map(|g| g.keys().map(String::as_ref))
```

src/font.rs
```diff
@@ -353,6 +353,70 @@ impl Font {
...
+            Some(k) => {
+                let mut kerningvec = vec![];
+                for kerningname in k.keys() {
+                    kerningvec.push(kerningname)
+                }
+                kerningvec.into_iter()
+            }
+            None => vec![].into_iter(),
+        }
+    }
```
cmyr, line 409: all the same comments as above.

src/font.rs
```diff
@@ -353,6 +353,70 @@ impl Font {
...
+
+    /// Returns a vector of layer name String types
+    pub fn layer_names(&self) -> Vec<String> {
+        self.layers.names().map(|s| s.to_string()).collect::<Vec<String>>()
+    }
+
+    /// Returns an iterator over layer name String types
+    pub fn iter_layer_names(&self) -> impl Iterator<Item = String> + '_ {
+        self.layers.names().map(|s| s.to_string())
+    }
```
cmyr, line 419: here we have options, but if we wanted symmetry with the other API additions my suggestion would be to return `&str`, and do `self.layers.names().map(|s| s.as_ref())`.

cmyr (commented):
Thanks!

I've offered some suggestions inline for changes I would make here, if we decide this (or some subset of this) is a valuable API addition.

The argument against this API is that it is largely implementing functionality that is already available, and the advantages aren't totally clear. Is `font.iter_lib_keys()` better than calling `font.lib.keys()` directly?

I think there's a better case for `group_names` and `kerning_names`, since the caller would have to dance around the `Option` in that case, and the proposed API is simpler.

I think the argument for `lib_keys` and `layer_names` is weaker, though, and my inclination would be to not add new API for these unless there are additional motivations that I'm overlooking. Does my reasoning here make sense?

chrissimpkins:
> Does my reasoning here make sense?

Yes, absolutely!  Thanks for the review here Colin.  I will run through your comments this week and will be back in touch.

chrissimpkins:
> I think there's a better case for group_names and kerning_names, since the caller would have to dance around the Option in that case, and the proposed API is simpler.
>
>I think the argument for lib_keys and layer_names is weaker, though, and my inclination would be to not add new API for these unless there are additional motivations that I'm overlooking. Does my reasoning here make sense?

Any consensus around whether the group_names and kerning_names API is worth preserving?  I can clean this up and remove the other methods if so. I'm ok closing this if you feel that all of this is of limited benefit.

cmyr:
I would like for these sorts of additions to be guided by use. Have we encountered a situation where we need the glyph or kerning names, and getting them has been annoying?

PR #146, opened 2021-07-27
Add custom missing mandatory UFO path error types & missing mandatory UFO path checks

chrissimpkins:
Supersedes #143 

New approach based on suggestion in https://github.com/linebender/norad/pull/143#issuecomment-887715136

### TODO:

- [x] add metainfo.plist check (d9d26c454f77939e59cb9c40a29a4f01028f439c)
- [x] add glyphs/contents.plist check (a418ef2d9f33a1f509f522322abb26a55649a722)
- [x] add tests for all new missing mandatory path checks


src/error.rs
```diff
@@ -44,6 +43,8 @@ pub enum Error {
     ExpectedPositiveValue,
     #[cfg(feature = "kurbo")]
     ConvertContour(ErrorKind),
+    MissingFile(String),
+    MissingUFODir(String),
```
cmyr, line 47: Rust naming convention is to only capitalize the first letter of acronyms and initialisms 🤷 
```suggestion
    MissingUfoDir(String),
```

cmyr (approved):
Looks good, thanks!

src/error.rs
chrissimpkins: Ty!

7b6cb1e672f5ec81ef155044f3d5f50080d6dec0

chrissimpkins:
I'll wrap up the items in the TODO list and then open this up for a full review.  If this approach seems reasonable, this should get us to read validations /error handling on all required UFO paths

chrissimpkins:
This PR now includes:

- UFO directory validation check
- metainfo.plist @ root validation check
- contents.plist in glyphs and background glyphs layer dirs check

+ new test sources & tests for each UFO spec violation check.  Ready for review.

src/font.rs
```diff
@@ -438,6 +445,69 @@ mod tests {
         assert_eq!(font_obj.features.unwrap(), "# this is the feature from lightWide\n");
     }
 
+    #[test]
+    fn loading_invalid_ufo_dir_path() {
+        let path = "totally/bogus/filepath/font.ufo";
+        let font_load_res = Font::load(path);
+        match font_load_res {
+            Ok(_) => panic!("unxpected Ok result"),
+            Err(Error::MissingUfoDir(_)) => (), // expected value
```
madig, line 454: `assert_eq!(font_load_res, Err(Error::MissingUfoDir(...)))` or something maybe?

madig:
Maybe consider removing inessential files from the test UFOs, like lib.plist.

src/font.rs
chrissimpkins: Tried.  It won't let me perform equality comparisons on the Error type

chrissimpkins:
> Maybe consider removing inessential files from the test UFOs, like lib.plist.

a25963ba662854460daf3e1c81103547e8f32c93

src/font.rs
cmyr: Two things:

As written, I would do this as,

```rust
assert!(matches!(font_load_res, Err(Error::MissingUfoDir(_))));
```

but there's probably a better option: explicitly testing for a panic with the [`should_panic`](https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-for-panics-with-should_panic) attribute.

chrissimpkins: > but there's probably a better option: explicitly testing for a panic with the should_panic attribute.

Ah nice, I didn't realize that you can add an expected string with `should_panic`.  I tried this approach but I can't elicit a panic from the new custom Error.  The attribute doesn't seem to catch errors like it does panics. I see this in the docs that you linked:

> You can’t use the #[should_panic] annotation on tests that use Result<T, E>. Instead, you should return an Err value directly when the test should fail.

Let me tinker with the test Result approach a bit.  If I can't find a better solution, I'll transition things over the matches! macro approach above.

cmyr: Maybe it would work if you just `unwrap` the `Result`?

cmyr: (but `matches!` is totally reasonable too)

chrissimpkins: It looks like the use of Result in unit tests is good if you expect an Ok() but there isn't a way to expect an Err.  We could return a test error when the Result is Ok 💥 but that seems worse than the matches pattern matching approach :) 

I went with the matches! macro. That works well!

b1c17aba624a771de855f48164edb20defec7959

cmyr (approved):
This looks good to me, feel free to merge at your leisure @chrissimpkins.

chrissimpkins:
> This looks good to me, feel free to merge at your leisure @chrissimpkins.

Thank you both for all of the feedback here!

Issue #145, opened 2021-07-26
Update `rust-plist`, use new features


cmyr:
There's a new 1.3.1 release. This update lets us make a few desired enhancements:

- [x] we can sort plist keys on write, which is very helpful for reducing unnecessary diff noise
- [x] we can customize whitespace behaviour. I forget what the exact motivation was for this, but I remember we wanted it.
- [x] we can get more information about the position of parse errors https://github.com/ebarnard/rust-plist/pull/64
- [x] we can review the code to do with plist serialization and use Value directly instead of workarounds (e.g. `layerinfo_from_file`?)

chrissimpkins:
These would be great additions Colin.

> we can customize whitespace behaviour. I forget what the exact motivation was for this, but I remember we wanted it.

This limits the diff too.  https://github.com/chrissimpkins/ufo-format-diff/pulls has examples of diffs across commonly used tooling + Noto sources.  Spacing represents the most lines of diff across the source files.

tabs/space formatting in norad discussed in https://github.com/linebender/norad/issues/101

If customization is now possible, it might be useful to allow a library user to choose the type of write spacing that they want (one tab, two tabs, two spaces, four spaces...)?  There seems to be no formatting consensus coming out of discussions on the UFO spec repo and I am guessing that there is even less consensus across the community.  While it is likely more work, it would make this lib more flexible to use *with* other UFO tools out there if this level of customization is possible.

cmyr:
Sounds good, I think should be simple enough to add an alternative "save" function that accepts some additional arguments for things like customizing whitespace.

cmyr:
Okay, here's an annoyance:

sorting dictionary keys requires mutable access to the dictionaries, which means mutable access to the whole font. The `save` function currently takes `&self`, which feels semantically correct; we don't really expect mutation on save.

Some options:

- bite the bullet and just make `save` take `&mut self`.
- Have a separate `&mut self` method, `normalize`, that does the sorting.
- use interior mutability to wrap plists, and sort during `save` without making it `&mut self`
- clone these collections and sort them before saving. (I really wish I had added an `is_sorted` method to `plist::Dictionary` 😞)


One complication with the `normalize` approach involves object libs. Unlike defcon, we store object libs in the actual objects, and then only reinsert them in the parent objects during serialization. We will need to sort after this step anyway, even if we've already called `normalize`.

[...]

PR #147, opened 2021-07-28
Update plist to 1.2, sort keys on save

cmyr:
I think I've added the sorting code in all the correct places?

- the font's lib
- the layer's lib
- the glyph's lib

Is there anything I'm missing? This is only needed in places where we're
dealing with an arbitrary dictionary, in other places serde will
serialize the keys in the correct order.

progress on #145


chrissimpkins:
It looks fontinfo.plist was already alphabetized, similar to the approach in ufonormalizer

chrissimpkins:
These sources have norad v0.4 run over them if it is helpful:

https://github.com/chrissimpkins/ufo-format-diff/pull/1/files

cmyr:
> It looks like the following are already in alphabetic order:
> 
> * fontinfo.plist
> * glyphs*/contents.plist appears to already be alphabetized with upper case sorted before lower case
> * groups.plist
> * kerning.plist
> * metainfo.plist

This sounds right, all of those files are deserialized into types that are sorted, and that sorting is maintained when we serialize again. The only thing this was not true for is things that norad represents as dictionaries, and these are only things like the various `lib` objects that are freeform.

chrissimpkins:
*Adding this here for documentation of how the new lib key ordering will take place*

[The plist sort_keys approach](https://docs.rs/plist/1.2.0/plist/dictionary/struct.Dictionary.html#method.sort_keys) supports lexicographic ordering by Unicode code point value as defined in the [`str` type docs](https://doc.rust-lang.org/nightly/std/primitive.str.html#impl-Ord):

> Strings are ordered lexicographically by their byte values. This orders Unicode code points based on their positions in the code charts. This is not necessarily the same as “alphabetical” order, which varies by language and locale. Sorting strings according to culturally-accepted standards requires locale-specific data that is outside the scope of the str type.

src/glyph/serialize.rs
```diff
@@ -79,13 +79,15 @@ impl Glyph {
...
         // cloning and write out the original.
+        let mut lib = self.lib.clone();
         let object_libs = self.dump_object_libs();
         if !object_libs.is_empty() {
-            let mut new_lib = self.lib.clone();
-            new_lib.insert(PUBLIC_OBJECT_LIBS_KEY.into(), plist::Value::Dictionary(object_libs));
-            write_lib_section(&new_lib, &mut writer)?;
-        } else if !self.lib.is_empty() {
-            write_lib_section(&self.lib, &mut writer)?;
+            lib.insert(PUBLIC_OBJECT_LIBS_KEY.into(), object_libs.into());
```
chrissimpkins, line 85: The changes as of https://github.com/linebender/norad/pull/147/commits/1ee9066f31ff21a5e44d48b903ad092dbbdb41b8 seem to be writing empty `public.objectLibs` dictionary fields in the lib.plist files for some reason

https://github.com/chrissimpkins/ufo-format-diff/commit/6633c6bc2a9f8a0f8527c4383510068d3ed6a10c

src/font.rs
```diff
@@ -251,17 +251,16 @@ impl Font {
...
-            let mut new_lib = self.lib.clone();
-            new_lib.insert(PUBLIC_OBJECT_LIBS_KEY.into(), plist::Value::Dictionary(object_libs));
-            plist::Value::Dictionary(new_lib).to_file_xml(path.join(LIB_FILE))?;
-        } else if !self.lib.is_empty() {
-            plist::Value::Dictionary(self.lib.clone()).to_file_xml(path.join(LIB_FILE))?;
+        // font.lib in-memory. Instead we clone the lib, add the object libs, and
+        // write out that.
+
+        let mut lib = self.lib.clone();
+        if let Some(object_libs) = self.font_info.as_ref().map(FontInfo::dump_object_libs) {
```
chrissimpkins, line 258: It looks like `FontInfo::dump_object_libs` may be able to return an empty plist Dictionary type and the `if !object_libs.is_empty()` gate was removed?

cmyr: thanks!

PR #148, opened 2021-07-29
Allow specifying whitespace on save

cmyr:
This lets the user optionally customize the whitespace used during
serialization.

The implementation is slightly annoying. the `quick_xml` and `rust-plist`
crates specify whitespace in different ways; `quick_xml` takes a (u8,
usize) pair (the byte and its count) and `rust-plist` takes a
`Cow<'static str>`. Additionally, in our own glyph serialization code we
need to write whitespace directly when appending the 'lib' section.

Anyway: you *create* a `WriteOptions` object by passing it a string, and
then we figure out what the appropriate (u8, usize) pair is. The reason
we want to think about this is because eventually writing out a UFO
should be parallelized, and that could mean cloning `WriteOptions` once
for each glyph; we would like to make sure this isn't allocating.

So this seems to work, even if some of the internals are a bit gross.

---

this is based off of #147, which should go in first.

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -1,0 +1,106 @@
+//! Customize serialization behaviour
+
+use std::{borrow::Cow, fs::File, io::BufWriter, path::Path};
+
+use plist::XmlWriteOptions;
+
+use crate::Error;
+
+/// Options that can be set when writing the UFO to disk.
+///
+/// You construct `WriteOptions` using builder semantics:
+///
+/// ```
+/// # use norad::WriteOptions;
+/// let single_tab = WriteOptions::default();
+///
+/// let two_tabs = WriteOptions::default()
+///     .whitespace("\t\t");
+///
+/// let spaces = WriteOptions::default()
+///     .whitespace("  ");
+/// ```
+#[derive(Debug, Clone)]
+pub struct WriteOptions {
+    // for annoying reasons we store three different representations.
+    pub(crate) indent_str: Cow<'static, str>,
+    xml_opts: XmlWriteOptions,
+    pub(crate) whitespace_char: u8,
+    pub(crate) whitespace_count: usize,
+}
+
+impl Default for WriteOptions {
+    fn default() -> Self {
+        WriteOptions {
+            indent_str: "\t".into(),
+            xml_opts: Default::default(),
+            whitespace_char: b'\t',
+            whitespace_count: 1,
+        }
+    }
+}
+
+impl WriteOptions {
+    /// Builder-style method to customize the whitespace.
+    ///
+    /// By default, we intent with a single tab ("\t").
+    ///
+    /// The argument, may be either a `'static str` or a `String`. You should
+    /// prefer to use a `'static str` where possible.
+    ///
+    /// The string can contain any number of *a single ASCII character*, but must
+    /// not contain multiple different characters. As an example, "\t\t" is
+    /// fine, but "\t  \t" is not, because it contains both tabs and spaces.
+    ///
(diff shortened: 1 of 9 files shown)


src/write.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+impl WriteOptions {
+    /// Create a new set of options.
+    pub fn new() -> WriteOptions {
+        WriteOptions::default()
+    }
+
+    /// Builder-style method to customize the whitespace.
```
chrissimpkins, line 36: The WriteOptions are instantiated like this?:

```rust
// single tab
let wo = WriteOptions::new().whitespace("\t");
```

```rust
// two spaces
let wo = WriteOptions::new().whitespace("  ");
```

cmyr: Yea, I don't love this API, but the idea is when there are more options (like the ' vs " char for the declaration) then those are also set here, and you can call any subset of the builder methods.

chrissimpkins: Sgtm.  Ty!

src/write.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub(crate) whitespace_char: u8,
+    pub(crate) whitespace_count: usize,
+}
+
+impl Default for WriteOptions {
+    fn default() -> Self {
+        WriteOptions {
+            indent_str: "\t".into(),
+            xml_opts: Default::default(),
+            whitespace_char: b'\t',
```
chrissimpkins, line 24: Single tab spacing is the ufonormalizer default? Optimal default for interoperability with other tools?

chrissimpkins:
@cmyr are you actively working on XML declaration single quote support? If not, I will have a shot at it based on the changes in this PR

cmyr:
@chrissimpkins I took a peek haven't implemented anything, but let's merge this first, so we can use `WriteOptions`.

I also think that it's worth being slightly fancier and doing the replacement before the initial write, it just feels more principled, and will be more portable (the write-then-rewrite approach may be slower on older storage media, or on other platforms.)

cmyr:
@chrissimpkins to elaborate on what I would propose:

- add a `quote_char` method to `WriteOptions`
- When serializing the `.glif`s, (which we serialize manually) write the header to the stream ourselves, using the specified character
- When serializing plists, we don't control writing the header, so we'll have to rewrite. There are options here. All `plist` writing is now handled through a method in `write.rs`. 
  - One option is to write to disk, and then use the unix/windows specific methods to go and overwrite the header in place, before closing the file handle: see [`std::os::unix::fs::FileExt`](http://doc.rust-lang.org/1.53.0/std/os/unix/fs/trait.FileExt.html) and [`std::os::windows::fs::FileExt`](http://doc.rust-lang.org/1.53.0/std/os/windows/fs/trait.FileExt.html); you'll need slightly different impls on either platform.
  - the other option would be to have a custom `Write` type, that would wrap or replace the `BufWriter`; this would intercept the header and replace it, if needed, otherwise forwarding the bytes to the inner writer.

Thinking about this, I think the first option is totally reasonable. The second option is probably marginally more 'efficient', but I don't think it matters; the main thing we want to avoid is closing and reopening all the file handles.

src/write.rs
xorgy: The other side of this of course is that in many cases, the plists and the other XML should have slightly different indent rules. The standard plist formatting skips the first indent level.

cmyr: Agree, the opening `<plist>` tag should not be indented. I believe we are currently doing this correctly.

PR #149, opened 2021-07-29
Remove examples/open_ufo, improve examples/load_save

cmyr:
I'll be honest, I didn't see that load_save existed and so I updated
open_ufo to include saving, and now we have two identical examples. The
new code is modestly better, so let's keep that and delete the extremely
redundant open_ufo example?

I want something that times file saving so that I can evaluate the utility of parallelizing glyph writing. 🤷


cmyr:
It turns out we use this in CI. We can use the other one. who knew? 🤦

PR #150, opened 2021-07-29
Use rayon when serializing glyphs

cmyr:
This looks like a modest (~30%?) win in serialization speed.


madig:
We didn't do that before? D'oh! Maybe this helps with write speed on Windows because file handles are getting closed in multiple threads.

cmyr:
@madig yea, I guess it was an oversight? I also assumed we were doing it, but we weren't. In any case, curious to see your results on windows.

PR #153, opened 2021-07-30
Add test helper to dump glif as String

cmyr:
This is... a very small convenience, but it makes writing tests slightly
less boilerplate-y?

Honestly I could go either way on this, happy for opinions


madig:
Hm... I'd need to see a few more functions using this.

cmyr:
agree we really don't need this right now.

PR #157, opened 2021-08-06
Add XML declaration custom single/double quote style support in glif and *.plist files

### LINKED ISSUE #156: Allow customizing quote char in xml header
- [x] add a `quote_char` method to `WriteOptions`
- [x] When serializing the `.glif`s, (which we serialize manually) write the header to the stream ourselves, using the specified character
- [x] When serializing plists, we don't control writing the header, so we'll have to rewrite. There are options here. All `plist` writing is now handled through a method in `write.rs`. 
  - One option is to write to disk, and then use the unix/windows specific methods to go and overwrite the header in place, before closing the file handle: see [`std::os::unix::fs::FileExt`](http://doc.rust-lang.org/1.53.0/std/os/unix/fs/trait.FileExt.html) and [`std::os::windows::fs::FileExt`](http://doc.rust-lang.org/1.53.0/std/os/windows/fs/trait.FileExt.html); you'll need slightly different impls on either platform.
  - the other option would be to have a custom `Write` type, that would wrap or replace the `BufWriter`; this would intercept the header and replace it, if needed, otherwise forwarding the bytes to the inner writer.

Thinking about this, I think the first option is totally reasonable. The second option is probably marginally more 'efficient', but I don't think it matters; the main thing we want to avoid is closing and reopening all the file handles.

_Originally posted by @cmyr in https://github.com/linebender/norad/issues/148#issuecomment-893514934_

chrissimpkins:
Closes #156

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -324,11 +324,7 @@ impl Layer {
 
         util::recursive_sort_plist_keys(&mut dict);
 
-        crate::write::write_plist_value_to_file(
-            &path.join(LAYER_INFO_FILE),
-            &dict.into(),
-            options.xml_options(),
-        )
+        crate::write::write_plist_value_to_file(&path.join(LAYER_INFO_FILE), &dict.into(), options)
     }
 
     /// Attempt to write this layer to the given path.
@@ -341,11 +337,7 @@ impl Layer {
 
     pub fn save_with_options(&self, path: &Path, opts: &WriteOptions) -> Result<(), Error> {
         fs::create_dir(&path)?;
-        crate::write::write_xml_to_file(
-            &path.join(CONTENTS_FILE),
-            &self.contents,
-            opts.xml_options(),
-        )?;
+        crate::write::write_xml_to_file(&path.join(CONTENTS_FILE), &self.contents, opts)?;
 
         self.layerinfo_to_file_if_needed(path, opts)?;
 
--- a/src/write.rs
+++ b/src/write.rs
@@ -174,7 +174,7 @@ mod tests {
         let filepath = tmp.path().join("lib.plist");
         write_plist_value_to_file(&filepath, &plist_read, &opt)?;
         let plist_write = fs::read_to_string(filepath)?;
-        let str_list = plist_write.split("\n").collect::<Vec<&str>>();
+        let str_list = plist_write.split('\n').collect::<Vec<&str>>();
         assert_eq!(str_list[0], "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"); // default uses double quotes
         assert_eq!(str_list[3], "<dict>"); // no space char at first dict tag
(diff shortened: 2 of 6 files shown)


src/layer.rs
```diff
@@ -349,19 +349,14 @@ impl Layer {
 
         self.layerinfo_to_file_if_needed(path, opts)?;
 
-        for (name, glyph_path) in self.contents.iter() {
```
chrissimpkins, line 352: It looks like this is duplicate code for the glif file writes and should be removed elsewhere if this PR is not merged

src/write.rs
```diff
@@ -36,14 +41,15 @@ impl Default for WriteOptions {
             xml_opts: Default::default(),
             whitespace_char: b'\t',
             whitespace_count: 1,
+            quote_style: QuoteStyle::Double,
```
chrissimpkins, line 44: Keep double quotes as default approach

chrissimpkins:
Ready for review.  I'll begin working on tests.

src/write.rs
```diff
@@ -69,21 +81,39 @@ impl WriteOptions {
...
     pub fn xml_options(&self) -> &XmlWriteOptions {
         &self.xml_opts
     }
 }
 
+#[derive(Debug, Clone)]
+pub enum QuoteStyle {
+    Single,
+    Double,
+}
```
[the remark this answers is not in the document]
chrissimpkins: Thanks Colin!

I accepted this commit and then refactored all uses in https://github.com/linebender/norad/pull/157/commits/8b59d3fde05d68adf7d8c8754cdc24f1656c4128

chrissimpkins:
Do we want to add Win CI testing since there are platform-specific approaches added here?

src/layer.rs
chrissimpkins: > I thiiiink what I'd like to do is revert that previous commit, and then I'll redo it and redo my measurements. I'll do that (the reverting) and then you'll need to patch up this PR.

Np, will do

madig:
I think so!

chrissimpkins:
> I think so!

https://github.com/linebender/norad/pull/161

chrissimpkins:
Tests added

src/write.rs
```diff
@@ -69,21 +81,44 @@ impl WriteOptions {
         self
     }
 
+    /// Builder-style method to customize the XML declaration attribute definition quote
+    /// char.
+    ///
+    /// By default, we indent with double quotes.
+    ///
+    /// The quote style is defined with a [QuoteChar] enum argument.
```
cmyr, line 89:
```suggestion
    /// The quote style is defined with a [`QuoteChar`] enum argument.
```

cmyr (approved):
Looks good, one tiny fixup, I'll apply that and merge. :)

PR #160, opened 2021-08-12
Fix bug where glyphs were saved twice

cmyr:
As pointed out in https://github.com/linebender/norad/pull/157.

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -349,11 +349,6 @@ impl Layer {
 
         self.layerinfo_to_file_if_needed(path, opts)?;
 
-        for (name, glyph_path) in self.contents.iter() {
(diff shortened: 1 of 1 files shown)


chrissimpkins:
ping @madig in case you were profiling I/O with the main branch

madig:
Thanks, but I haven't gotten around to profiling at all :(

madig:
So, I did a very quick replacement in my test program of `norad = { git = "https://github.com/linebender/norad.git", rev = "bcb1ddf425aa8e6d9da3a945ac2e7e10e6717e47", features = ["rayon"] }` and the runtime is unchanged. I don't know if I'm overlooking something.

cmyr:
yea, I profiled this and it looks like rayon is a small win, which I interpret as meaning that the bottleneck is actually writing to disk, which makes sense.

PR #165, opened 2021-08-19
Make all FontInfo structs accessible

madig:
So I can access them from iondrive.

Also, validate FontInfo before writing it out.


cmyr (approved):
looks good!

PR #166, opened 2021-08-19
Remove submodules

cmyr:
~This includes the commits from #161 because I didn't want to need to rebase that, and I want to test this all rolled up; if this works then I'll drop those commits and merge, and we can merge #161 separately.~

Okay, this removes submodules and adds a `.gitattributes` file so that git doesn't touch line endings on `.fea`, `.plist`, and `.glif` files in this repository.


madig:
You can also leave the files as text but enforce LF line endings. Not sure what is better.

cmyr:
I chose this approach because it would also let us have test data that uses CRLF.

PR #169, opened 2021-08-20
Cleanup docs and exports

cmyr:
This makes a few modules private that didn't need to be public, and
cleans up or adds docs to a few missing items.

@madig I don't think this should cause any problems, but feel free to take a look?


madig:
Looks good to me. I think I just need the top level exports of `Kerning` and `Groups`.

PR #170, opened 2021-08-20
Add optional Win platform line ending normalization support

### LINKED ISSUE #162: Cross-platform line ending serialization approach
Thoughts on how to approach cross-platform line endings?  Should norad write consistent line endings on all platforms (limits diff when norad serialization is used x-platform)?  Keep line ending format that it reads in on subsequent serialization (maybe limits diff when norad serialization is used with other serializers)?  Allow custom line ending support (flexible and maybe limits diffs across serializers and platforms)?  

@belluzj mentioned today that he likes the idea of a runebender config panel with Unix vs. Win line ending style options on UFO serialization.  I think that we can expand the approaches used for spacing and XML declaration quote customization to address it if this is desirable.  All of these serialization format options might be useful in such a config panel.

_Originally posted by @chrissimpkins in https://github.com/linebender/norad/issues/161#issuecomment-899638531_

chrissimpkins:
Closes #162

---

### TODO

- decide approach to feature files (do we want to update line endings there too?)
- implement fea file line ending edits


src/glyph/serialize.rs
```diff
@@ -111,9 +111,25 @@ impl Glyph {
         }
 
         writer.write_event(Event::End(BytesEnd::borrowed(b"glyph")))?;
-        writer.inner().write_all("\n".as_bytes())?;
-        writer.inner().flush()?;
 
+        // Write `\r\n` final line ending on Win by default
+        // Write `\n` final line ending on all other platforms by default
+        if cfg!(windows) && !options.line_ending_normalization {
```
chrissimpkins, line 117: This is a change in the current default approach to the final line ending in glif files on all platforms.  This will now write a final `\r\n` on Win and `\n` on non-Win.  This is consistent with the decision about how to approach defaults and customization support.

src/write.rs
```diff
@@ -48,6 +53,7 @@ impl Default for WriteOptions {
             whitespace_char: b'\t',
             whitespace_count: 1,
             quote_style: QuoteChar::Double,
+            line_ending_normalization: false,
```
chrissimpkins, line 56: normalization is off by default

src/write.rs
```diff
@@ -92,12 +98,70 @@ impl WriteOptions {
...
+}
+
+/// Iterator implementation for the CarriageReturnRemover struct
+///
+/// Iterates through u8 and removes all `\r` to normalize line endings
+/// as `\n` only irrespective of the platform.  Returns u8 values.
+///
+/// This implementation is a u8 iterable derivative of
+/// https://github.com/derekdreery/normalize-line-endings/blob/master/src/lib.rs
+/// (Apache License v2)
```
chrissimpkins, line 130: adds derivative source under Apache License 2

chrissimpkins:
Hmm... I thought that glif were being serialized with `\r\n` endings on Win?  Seems not to be the case in current tests.

chrissimpkins:
> Hmm... I thought that glif were being serialized with `\r\n` endings on Win?  Seems not now for some reason...

Checked the quick-xml source and it looks like they write `\n` everywhere.  Ugh.  Looks like this approach will need to be revised.

chrissimpkins:
Revised this PR.  As of https://github.com/linebender/norad/pull/170/commits/c42f4806bd130c56ecb89711793565bf1c174d53 we now have (1) platform-specific glif and plist file line endings (Win = `\r\n`, non-Win = `\n`); (2) optional Win line ending normalization to `\n` via the WriteOptions config.  In place of a home baked approach, I opted for a stripped down (all Unicode features removed) regex crate approach to do the replacements with the `bytes` module.  It's highly unlikely that I will write something faster than the implementation there. :) 

Still requires tests/expected file updates and a decision on how to handle fea files, but I'll remove the draft status and open this up for review.

chrissimpkins:
Benched the plist line ending byte replacement iterator approach in this branch vs. a regex::bytes::Regex.replace_all` approach and the regex implementation is about 2-fold faster with and without `\r\n` matches in the file:

```
// in a fontinfo.plist file with \r\n line endings
test write::tests::bench_iterator ... bench:       4,209 ns/iter (+/- 281)
test write::tests::bench_regex    ... bench:       1,942 ns/iter (+/- 165)

// in a fontinfo.plist file with \n line endings (i.e. no matches)
test write::tests::bench_iterator_no_match ... bench:       4,413 ns/iter (+/- 344)
test write::tests::bench_regex_no_match    ... bench:       1,956 ns/iter (+/- 163)
```

Pushed the tests and test functions to https://github.com/chrissimpkins/norad/tree/bench-normalize-line-endings

chrissimpkins:
Any idea what might be going on in the lib.plist normalized line ending test on Win?  For some reason the attempt to make an iterator over the File bytes with `file.bytes()` raises an error on a byte in the test file with the cryptic error message:

```
thread 'write::tests::write_lib_plist_with_normalized_line_endings' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 5, kind: PermissionDenied, message: "Access is denied." }', src\write.rs:236:64
```

https://github.com/linebender/norad/pull/170/checks?check_run_id=3412798623

chrissimpkins:
Pausing work on this until the [additional feedback from Khaled](https://github.com/linebender/norad/issues/162#issuecomment-905049436) is discussed and a decision is made on the approach.

PR #171, opened 2021-08-26
Remove Mutator Sans git submodule configuration

chrissimpkins:
The test data submodule dir was transitioned away from a git submodule.  This PR removes the git submodule config.


cmyr (approved):
Thanks!

PR #172, opened 2021-08-26
Serialize consistent cross-platform line endings across plist, glif, and feature files

chrissimpkins:
Supersedes #170

Based on discussion in #162, we will write line feeds in all UFO files on all platforms by default.

This branch is based on #171  (edit: #171 now merged into master branch)

---

### TODO

Confirm that line feeds are serialized by default in:

- [x] plist files (tests added in https://github.com/chrissimpkins/norad/commit/87879bf59470a019e61e4b34e7cfde57a2a75a03)
- [x] glif files (test added in https://github.com/chrissimpkins/norad/commit/05517ba027de099e597c3546cef16fc0ef67fd37)
- [x] fea files (x-platform normalization added in https://github.com/chrissimpkins/norad/commit/ec898ebff8199a51dfd3139dc7fdb4dfe45baa86, modified in https://github.com/linebender/norad/commit/6548bb3f4f53d6bd58fc0b7084c162472135704f)


chrissimpkins:
Thoughts about this approach @cmyr @madig?

cmyr:
Sorry to let this sit, my attention has been elsewhere. I'm off the computer shortly for some irl appointments but I'll try to take a look tonight or tomorrow!

src/font.rs
```diff
@@ -264,7 +264,13 @@ impl Font {
         }
 
         if let Some(features) = self.features.as_ref() {
-            fs::write(path.join(FEATURES_FILE), features)?;
+            // Normalize feature files with line feed line endings
+            // This is consistent with the line endings serialized in glif and plist files
+            if features.contains('\r') {
```
cmyr, line 269: just a thought, and maybe the compiler is smart enough to optimize this, but I would prefer to scan for the `\r` *byte* over the `\r` *char*; scanning chars means decoding utf-8, whereas scanning bytes doesn't.

cmyr: My only other thought (and this is totally unnecessary) is that if we really wanted to be perf nerds, the better approach would probably be to do this replacement when reading, instead of when writing.

Basically:

- instead of something like `fs::read_to_string`, manually open the fea as a `File`
- allocate two buffers
- read chunks from the file into the first buffer. on each read, scan the chunk for `\r`
- if `\r` is encountered, then manually recopy the lines into the second buffer, skipping an `\r`
- if `\r` is never encountered, just return the first buffer untouched as the result, else return the second buffer.


this is dumb and definitely not worth it. 🤷

cmyr (approved):
This works for me!

src/font.rs
chrissimpkins: > scanning chars means decoding utf-8

I was wondering about the read approach when I worked on this.  It looks like the feature read is to String so the UTF-8 check happens read side.  


```rust
fn load_features(features_path: &Path) -> Result<String, Error> {
    let features = fs::read_to_string(features_path)?;
    Ok(features)
}
```

Does the feature file need to be a UTF-8 vetted String?  On the one hand, I suppose that it provides a built-in linting check at read time.  On the other, it may not be necessary unless there are String'y things that will need to be done with it in the library.  Right now it appears that feature data are exposed as is to users in a String.

chrissimpkins: > I would prefer to scan for the \r byte over the \r char

Would this be something along the lines of searching for `b"\r"` in `features.as_bytes()`?

cmyr: Exactly.

cmyr: > > scanning chars means decoding utf-8

what I mean is simply that the chars() iter has to determine character boundaries in the underlying bytes, which has some overhead. This is on top of the overhead of validating the utf-8 when it's first read. 
> 
> I was wondering about the read approach when I worked on this.  It looks like the feature read is to String so the UTF-8 check happens read side.

Yes, I believe we read directly to string currently, which handles validation. We could alternatively read to bytes and then validate ourselves, to equal effect. 
> 
> 
> 
> 
> 
> ```rust
> 
> fn load_features(features_path: &Path) -> Result<String, Error> {
> 
>     let features = fs::read_to_string(features_path)?;
> 
>     Ok(features)
> 
> }
> 
> ```
> 
> 
> 
> Does the feature file data need to be a UTF-8 vetted String?  On the one hand, I suppose that it provides a built-in encoding linting check at read time.  On the other, it may not be necessary unless there are String'y things that will need to be done with it in the library.  Right now it appears that feature data are exposed as is to users in a String without other feature support.

This is a good question, which is not answered by the spec. The spec does mention that strings are in utf-8, and I would prefer to assume that the whole file is, as well. It might be worth opening an issue in the Adobe repo...

chrissimpkins: converted to a byte check in 2e84bce7ccd126a9fc8b9a3f5338e067eed25139

chrissimpkins:
New benchmarks with conversion to a byte check from char check (https://github.com/linebender/norad/pull/172#discussion_r701905359):

```
test tests::benchmark_bytes_contains_with_cr           ... bench:       2,119 ns/iter (+/- 374)
test tests::benchmark_bytes_contains_with_cr_last_line ... bench:       1,381 ns/iter (+/- 78)
test tests::benchmark_bytes_contains_without_cr        ... bench:         257 ns/iter (+/- 26)

test tests::benchmark_contains_with_cr                 ... bench:       2,201 ns/iter (+/- 200)
test tests::benchmark_contains_with_cr_last_line       ... bench:       1,392 ns/iter (+/- 107)
test tests::benchmark_contains_without_cr              ... bench:         256 ns/iter (+/- 46)

test tests::benchmark_no_contains_with_cr              ... bench:       2,130 ns/iter (+/- 123)
test tests::benchmark_no_contains_with_cr_last_line    ... bench:       1,200 ns/iter (+/- 73)
test tests::benchmark_no_contains_without_cr           ... bench:       1,029 ns/iter (+/- 145)
```

Appears to have the same performance within the margin of error.

src/font.rs
```diff
@@ -264,7 +264,13 @@ impl Font {
...
+                fs::write(path.join(FEATURES_FILE), features.replace("\r", ""))?;
```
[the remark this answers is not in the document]
madig: I think VTT table dumps (in fontTools' TTX format) contain `\r`, but the two characters and not `\u000D`. I think the data folder shouldn't be touched anyway.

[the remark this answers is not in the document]
madig: Oh wait. It actually contains literal `\u000D`, see https://github.com/daltonmaag/ubuntu/blob/master/source/Ubuntu-B.ufo/data/com.github.fonttools.ttx/T_S_I__1.ttx. Maybe because VTT started life on very old Macs that still used `\r`?

I suppose mixing LF and CRLF in any part of the UFO (glyph lib or elsewhere) is a bad idea so I guess this PR is fine, though.

madig:
@chrissimpkins can you please clarify how I need to read your benchmark results?

chrissimpkins:
> @chrissimpkins can you please clarify how I need to read your benchmark results?

https://github.com/linebender/norad/pull/172#issuecomment-906413066 is the original run with a description.  The latest tests added a third test group with the new approach to check for the carriage return byte rather than char.

I ran tests against test strings that included:

- `with_cr` = carriage returns and line feeds on every line
- `with_cr_last_line` = carriage return and line feed on the last line only (to test a full doc traversal before the CR is identified)
- `without_cr` = no carriage returns, only uses line feeds

The test is whether contains is useful in any of the above circumstances and these tests were grouped as follows:

- `bytes_contains` = `input.as_bytes().contains(&b'\r')` check
- `contains` = `input.contains('\r')` check
- `no_contains` = no check for the carriage return, we just run the replacement across the entire string

Full repro source in the attached archive

[benchy.zip](https://github.com/linebender/norad/files/7136459/benchy.zip)

src/font.rs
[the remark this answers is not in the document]
chrissimpkins: The data will be read in as a byte array?

chrissimpkins:
IMO the tl;dr takeaway is that, in the vastly most common case of fea files with line feeds only, there is an improvement in execution time if you *do* gate the string replacement on a check (~250ns/iter vs. ~1000ns/iter).  It appears that char and byte checks run in ~ the same time.

All tested on MBP 2017 macOS 10.15.7 with current rustc nightly

chrissimpkins:
We are dealing with very short times and max one string per ufo master so this is not going to break anything.  Happy to remove the check if anyone feels strongly about more concise/clean source.  It was not intuitive to me that this would be the finding.

cmyr (approved):
Looks good, thanks for measuring!

chrissimpkins:
Came across this in [the Adobe fea spec documentation](https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#2b-white-space):

> You can have multiple line endings, spaces, and tabs between tokens. Macintosh, UNIX and PC line endings are all supported.

Recording it here for posterity

src/font.rs
chrissimpkins: > This is a good question, which is not answered by the spec. The spec does mention that strings are in utf-8, and I would prefer to assume that the whole file is, as well. It might be worth opening an issue in the Adobe repo...

Response from Josh Hadley (Adobe):

> Yes, FEA files should indeed use UTF-8 encoding throughout. This has been an unwritten (well…partially written) assumption for some time now but as of the latest AFDKO (3.7.1), that assumption is enforced by the new Antlr4-based parser in makeotfexe. Older versions of makeotfexe might have been lax on that point, but that will no longer be the case moving forward. Also worth noting that makeotf (the Python interface to makeotfexe) has effectively enforced this for a while already.
>
> But as you note, the OpenType Feature File Specification does not actually explicitly state that the whole file should be UTF-8. We will clarify that in an update soon.

PR #174, opened 2021-09-03
Add glif contour point PointType enum Display trait

chrissimpkins:
I used norad for outline incompatibility diagnostics across multiple UFO masters this week and it was helpful to have the glif contour point type attribute string format in the tests.   This PR includes a Display trait implementation, a refactor of the FromStr trait implementation to the module that contains the PointType enum, and  Display/FromStr/PartialEq tests.


cmyr (approved):
Looks good!

PR #176, opened 2021-09-06
Handle data and images dirs

madig:
This PR adds two new public-facing methods `Font::data_paths()` and `Font::images_paths()`, which return the data and images paths found at UFO loading time. The client can then join the UFO root path and the paths to get the full path.

The primary use-case here is to let iondrive load data and image files to pass over to ufoLib2. Eventually the API here should probably be replaced with full handling of data and images (both of which are basically HashMaps of paths to bytes).

Some notes:

1. This only stores the paths and gives out immutable references. Manipulation is currently not supported.
2. Data and images are not written back out currently.
3. I don't know how to handle symlinks in either directory, see https://github.com/unified-font-object/ufo-spec/issues/188. I decided to error out.
4. The reference fontTools.ufoLib implementation seems to only error out when data and images are files instead of directories, otherwise it seems to largely just pretend the directories are empty. I decided to be more unforgiving. ufoLib will also check that images start with the PNG header and will silently skip them otherwise. I decided to leave that to clients for now.
5. I can't make a symlink test for Windows because you need permissions to create them.
6. Both methods return paths prefixed with `images/` and `data/`, so the client can just join the UFO root and the path. I'm not sure this is what we want, longer term.

I'm not entirely sure about this. Something tells me I should get full handling into norad instead.


madig:
I can successfully load data now in https://github.com/simoncozens/iondrive/pull/6/commits/e1268b5aab76f567e57bf52194cd1ffc216ec135.

madig:
I'm not entirely happy with this... So, here's a bunch of notes:

1. The easiest and most straightforward way of handling data and images would be to have `HashMap<PathBuf, Vec<u8>>` for data and images and load (and validate) everything at load time. Access is easy, writing things out (in-place, elsewhere) is easy.
2. This can lead to waste of time and RAM loading things that are never needed for e.g. font compilation. I haven't measured what impact this would have.
3. ufoLib(2) solves this by loading both lazily, just like glif files. You read things as you access them. This requires keeping a "reader" object around, with, inside, a path to go from.
4. norad has a DataRequest API where a compiler could simply disable e.g. images loading, but the data folder may contain a lot of big files of which only a couple are relevant for compilation.
5. Adding a lazy loader to norad is possible but I don't like the idea of keeping a path around and you might need a `&mut` reference to `Font` to load something (because it has to be written into the struct under the hood). One could potentially use interior mutability but then you just shift the borrow checker to runtime. Is it worth the complexity?
6. How to manage adding and removing data and images? Without lazy loading, you can easily have a Data and Images struct with a couple of HashMap's methods. I'd need to sketch this out.

benkiel:
Fwiw, open readers in Windows is hell. We have been dealing with this since adding ufoz to the spec

madig:
@benkiel interesting! Can you expand a bit on that? I feel like UFOZ support added a whole lot of complexity to ufoLib with very limited benefit. Does anyone even use it?

benkiel:
@madig Windows does not like having references to open files hanging about, so just be careful there. That's what we ran into with ufoLib/defcon. I think it was always an issue, but became highlighted with UFOZ much more.

UFOZ was something that was asked for for years. It solves some things (emailing UFOs, Dropbox choking on collections of UFOs, having one file, not a directory for Windows). Yes, it came at the cost of a lot of complexity in ufoLib, and no, I don't have a good sense of how much it is used. But, it was a clear ask from many for it.

cmyr (commented):
I don't think this is a necessarily a bad approach.

My main question is about what happens on save. Do we currently preserve the data/image directories? We certainly should be doing this, if we're not.

This also would make it easy to update this to allow *removing* data and images; basically we would only copy over the data/images that are in our vecs, and so deleting data/images is as simple as removing an item from the appropriate vec.

I do agree that this doesn't feel *perfect*. I think a better long-term solution would be some sort of `DataSet` type that is used for both images/data, where we will lazily load the data as requested, and allow new items to be added directly to the UFO. It might be worth heading in that direction now, but if this patch gets you working in iondrive then I think that's okay for now,  but.

madig:
No, this PR does not preserve images and data across saves.

I am tempted to go all-in non-lazy DataSet and ImageSet now, to get it right-ish from the get-go... and make adding and removing simple.

chrissimpkins:
> non-lazy 

What is a typical filled data and img directory size in your experience? One or two small files?  Hundreds of large files?

madig:
I have typically seen six-ish files weighting 1 MB or so, per UFO. I have never seen anything use the images folder. @benkiel have you seen higher usage?

benkiel:
I've had 30mb+ directories for color fonts using image SVGs. I think usually it's going to be sketch layers, not a ton, but it can be a whole huge thing with lots of files for color fonts.

madig:
Superseded by #180.

PR #180, opened 2021-09-15
Handle UFO data and images v2

madig:
New PR:

Implement data and images stores.

- Load data lazily.
- Provide basic HashMap-like interface in `Font.data` and `Font.images` for managing data.
- `RefCell` setup to allow `get` to only take `&self`.

---

Original PR:

Implement data and images stores.

- Load data and images at UFO load, write them back out at save time.
- No lazy loading, as it [complicates the logic](https://github.com/linebender/norad/pull/176#issuecomment-916446455). Maybe that's a good enough approach until we start doing Big Data in UFOs.
- Provide basic HashMap-like interface in `Font.data` and `Font.images` for managing data.
- Ensure sanity at insertion time so we can just write things out at save time.


madig:
Note to self, maybe turn off mut methods if they can circumvent validity checks

src/datastore.rs
```diff
@@ -0,0 +1,373 @@
...
+
+    /// An iterator visiting all path-content pairs in arbitrary order, with mutable
+    /// references to the content.
+    pub fn iter_mut(&mut self) -> impl Iterator<Item = (&PathBuf, &mut Vec<u8>)> {
+        self.store.iter_mut()
+    }
+}
+
+/// There are no methods for getting mutable references to images to prevent circumvention of the PNG validity check.
+impl ImageStore {
```
cmyr, line 209: I think we should be able to combine these into a single type.

From what I can tell, there are two places where these two types differ: in how they do the initial data loading, and in how they handle insertion.

So what we can imagine is some shared `DataStore` type that allows customization of this behaviour. One option would be to use a trait, like:

```rust
trait DataType {
    type Error;
    fn try_load(&self, path: &Path) -> Result<HashMap<PathBuf, Vec<u8>>>;
    fn try_insert(
        &self,
        collection: &mut HashMap<PathBuf, Vec<u8>>,
        path: PathBuf,
        data: Vec<u8>,
    )  -> Result<Option<Vec<u8>>, ImageStoreError>;
}


struct Data;
impl DataType for Data { 
    // impl here
}

struct Image;

impl DataType for Image {
    // image specific code here
}

struct Store<T> {
    store: HashMap<PathBuf, Vec<u8>>,
    kind: T,
}

type DataStore = Store<Data>;
type ImageStore = Store<Image>;
```

but this might be overkill? Another simpler option would be to have separate `try_new_images` and `try_new_data` methods, and then just store a `bool` indicating whether we've loaded images or data, and switch on this bool in `try_insert` in order to do validation correctly?

In any case these are just a couple possible options, I'm sure you can think of others. :)

cmyr (commented):
Cool, I do think this is an improvement.

My main note is that I think we should try and unify these two storage types. This will be easier to maintain in the long run, and will also make it much easier to go and move to lazy loading, because we will only have to implement that in one place.

On that note: I think that longer-term we will want to go ahead and add lazy loading, it really shouldn't be too much more difficult and it should have real performance implications in some situations.


My only note there would be that we should try and be aware of how the lazy loading version might require a different API (does it? in what ways?) and see if we can adopt that API now, so that making it lazy in the future isn't a breaking change. :)

madig:
I think for lazy loading, all you need is to store the path to the UFO root plus have a canary value for unloaded content (`enum Entry { Unloaded, Loaded(Vec<u8>) }` I suppose). Then you can on-demand concatenate the UFO root path + data or images + the key path.

There's additional logic needed for saving:
- Are you doing `fs::remove_dirs_all`? Then you need to first load everything.
- Are you updating the UFO in place? Then you can just write out what you have in memory but may need to remove fiels that you no longer track in the store.

cmyr:
we currently do `remove_dir_all.` For saving then we could either load all the images before saving, or we could try and do something fancier, like copy them to a temporary directory and then copy them back when we're done? 🤔

madig:
I think getting rid of remove_dir_all is desirable to speed up writing small changes, but needs a more wholistic approach to saving selectively. You'd need to build a list of "diffs" you'd write out, which includes a list of files to delete. I think that's a topic for its own issue.

I'm currently sketching out a LazyDataStore.

src/datastore.rs
```diff
@@ -0,0 +1,609 @@
...
+///
+/// * `PathBuf::from("image1.png")` → `b"<content>".to_vec()`
+/// * `PathBuf::from("image2.png")` → `b"<content>".to_vec()`
+/// * `PathBuf::from("image3.png")` → `b"<content>".to_vec()`
+///
+/// [spec_data]: https://unifiedfontobject.org/versions/ufo3/data/
+/// [spec_images]: https://unifiedfontobject.org/versions/ufo3/images/
+#[derive(Debug, Clone)]
+pub struct Store<T, E> {
+    items: HashMap<PathBuf, Arc<RwLock<Item<E>>>>,
```
cmyr, line 67: We definitely don't want a lock here. If we want interior mutability, we could consider a `RefCell`.

I get wanting to keep `get()` and `iter()` non `&mut self`.

Maybe the easiest way to do this is to split up `Item`, like:

```rust
struct ItemInner {
    Loaded(Arc<[u8]>),
    NotLoaded,
    Error(Arc<Error>),
}

Item {
    inner: RefCell<ItemInner>,
}

impl DataStore {
    fn get(&self, key: &Path) -> Result<Option<Arc<[u8]>>, Error> {
        let item = match self.items.get(key) {
            Some(item) => item,
            None => return Ok(None),
        };

        if let Item::NotLoaded = item {
            match self.impl_type.try_load_item(path) {
                Ok(data) => *item = Item::Loaded(data),
                Err(e) => *item = Item::Error(e),
            }
        }

        match item {
            Item::Loaded(data) => Ok(Some(data.clone())),
            Item::Error(err) => Err(err.clone()),
            Item::NotLoaded => unreachable!(),
        }
    }
```

I think that would work? And be a bit simpler maybe?

cmyr (commented):
Okay A few possible suggestions inline, but I think overall we're getting close!

src/datastore.rs
```diff
@@ -0,0 +1,609 @@
...
+            return false;
+        }
+
+        self.iter().all(|(key, value)| {
+            other.get(key).map_or(false, |value_other| match (value, value_other) {
+                (Ok(s), Ok(o)) => *s == *o,
+                _ => false,
+            })
+        })
+    }
```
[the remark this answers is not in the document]
madig: So no comparison of actual data?

src/data_request.rs
```diff
@@ -63,6 +77,32 @@ impl DataRequest {
...
+
+    /// Request that returned UFO data include data.
+    pub fn data(&mut self, b: bool) -> &mut Self {
+        self.data = b;
+        self
+    }
+
+    /// Request that returned UFO data include data (loaded into memory).
+    /// Otherwise, they are loaded on first access.
+    pub fn data_eager(&mut self, b: bool) -> &mut Self {
```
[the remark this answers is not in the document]
madig: I imagine being able to fully cut the ties to the file system? I mean, one can also just iter through all data and images I suppose?

src/datastore.rs
madig: Will using a RefCell impede multithreading when e.g. multiple threads access the data store while some item us `NotLoaded` still?

src/datastore.rs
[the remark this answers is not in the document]
cmyr: I don't think that's necessary.

src/data_request.rs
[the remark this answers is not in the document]
cmyr: In general I prefer not to write code unless there is a clearly demonstrated need.

src/datastore.rs
cmyr: `Font` is not `Sync`, and cannot be used concurrently from multiple threads.

belluzj: Hello! I was looking at this with Nikolaus the other day and now I'm curious too to know more about RefCell vs RwLock. The parallel scenarios we had in mind are:

* open a UFO with norad in the main thread, then launch a thread pool to build individual glyphs in parallel
* open a designspace with several UFOs attached, and launch thread to interpolate instances from that.

All the threads would share the UFO and also this store, and potentially all would try to read from the store. Does that make sense?

belluzj: (I just saw your reply about the Font not being Sync)

cmyr: so the images/data is definitely something that would complicate that. Everything else in the font object doesn't require any mutation, and so if you used some kind of scoped thread pool you should be able to do things like get glyphs from multiple threads, although I haven't played around with this.

If you need images/data to build glyphs, then we could look at the `RwLock` design, but it isn't anything we've needed anywhere else in the library so far.

In practice, it might be useful to know how the image/data resources are commonly used? are items there commonly shared between many glyphs?

belluzj: To be honest in the workflows I know of, it's not the case at all, usually in the data folder we put XML dumps of fontTools data to be merged into the final TTF, or background images that are useful while designing but not while compiling, so it would probably not be an issue.

belluzj: Other foundries might be doing different things though, not sure about that

src/data_request.rs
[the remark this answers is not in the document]
madig: Ok, so all in lazy-loading?

src/datastore.rs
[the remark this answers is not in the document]
madig: Hm, doesn't `keys` return keys in arbitrary order so we get mismatches?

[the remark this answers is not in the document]
madig: I'm gonna do `... && self.items.keys().all(|key| other.items.contains_key(key))`

src/datastore.rs
madig: Using just a RefCell<Item> now. I indeed also had to remove the test where we launched 100 threads and read stuff from an outer Font because RefCell could not be passed over :)

belluzj: Yes sorry about saying we should add that test, I didn't know concurrent access was not in the scope

src/datastore.rs
```diff
@@ -0,0 +1,561 @@
...
+        items: &HashMap<PathBuf, RefCell<Item>>,
+        data: &[u8],
+    ) -> Result<(), StoreError>;
+}
+
+#[derive(Debug, Clone)]
+#[doc(hidden)]
+pub enum Item {
+    NotLoaded,
+    Loaded(Arc<Vec<u8>>),
```
cmyr, line 104: minor thing but I would store `Arc<[u8]>` (a reference counted slice) instead of `Arc<Vec<u8>>`. This is minorly more efficient (less pointer indirection, although this is unlikely to matter in practice) but also more clearly expresses that the data cannot be mutated.

cmyr (approved):
great! sorry about the runaround, but I think this is pretty much ready to go. One last little nit, otherwise I'm happy to get this merged. Thanks!

src/datastore.rs
madig: Ah, ok, so we fully expect people to remove the data they want to mod and then insert it again?

PR #181, opened 2021-09-17
Add Glyph components public methods

chrissimpkins:
Adds a few QoL methods around Glyph components.

### DIFF
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -122,8 +122,8 @@ impl Glyph {
     }
 
     /// Return a Vec of immutable [`Component`] references filtered by base glyph name.
-    pub fn get_components_with_base(&self, basename: &str) -> Vec<&Component> {
-        self.components.iter().filter(|x| *x.base == *basename).collect::<Vec<&Component>>()
+    pub fn get_components_with_base<'b, 'a: 'b>(&'a self, basename: &'b str) -> impl Iterator<Item=&'a Component> + 'b {
+        self.components.iter().filter(move |x| *x.base == *basename)
     }
 
     /// Move libs from the lib's `public.objectLibs` into the actual objects.
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -104,6 +104,28 @@ impl Glyph {
         }
     }
 
+    /// Returns boolean value indicating whether [`Glyph`] is defined with one or
+    /// more [`Component`].
+    pub fn has_component(&self) -> bool {
+        !self.components.is_empty()
+    }
+
+    /// Returns a usize that represents the number of [`Component`] defined on the Glyph.
+    pub fn component_number(&self) -> usize {
+        self.components.len()
+    }
+
+    /// Returns boolean indicating the presence of one or more [`Component`] with base
+    /// glyph name `basename`.
+    pub fn has_component_with_base(&self, basename: &str) -> bool {
+        self.components.iter().any(|x| *x.base == *basename)
+    }
+
+    /// Return a Vec of immutable [`Component`] references filtered by base glyph name.
+    pub fn get_components_with_base(&self, basename: &str) -> Vec<&Component> {
+        self.components.iter().filter(|x| *x.base == *basename).collect::<Vec<&Component>>()
+    }
+
     /// Move libs from the lib's `public.objectLibs` into the actual objects.
     /// The key will be removed from the glyph lib.
     fn load_object_libs(&mut self) -> Result<(), ErrorKind> {
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -111,7 +111,7 @@ impl Glyph {
     }
 
     /// Returns a usize that represents the number of [`Component`] defined on the Glyph.
-    pub fn component_number(&self) -> usize {
+    pub fn component_count(&self) -> usize {
         self.components.len()
     }
 
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -121,8 +121,11 @@ impl Glyph {
         self.components.iter().any(|x| *x.base == *basename)
     }
 
-    /// Return a Vec of immutable [`Component`] references filtered by base glyph name.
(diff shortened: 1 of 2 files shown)


src/glyph/mod.rs
```diff
@@ -104,6 +104,28 @@ impl Glyph {
...
     }
 
+    /// Returns boolean value indicating whether [`Glyph`] is defined with one or
+    /// more [`Component`].
+    pub fn has_component(&self) -> bool {
+        !self.components.is_empty()
+    }
+
+    /// Returns a usize that represents the number of [`Component`] defined on the Glyph.
+    pub fn component_number(&self) -> usize {
```
cmyr, line 114: I have a minor preference for:

```suggestion
    pub fn component_count(&self) -> usize {
```

src/glyph/mod.rs
```diff
@@ -104,6 +104,28 @@ impl Glyph {
...
+    /// Returns boolean indicating the presence of one or more [`Component`] with base
+    /// glyph name `basename`.
+    pub fn has_component_with_base(&self, basename: &str) -> bool {
+        self.components.iter().any(|x| *x.base == *basename)
+    }
+
+    /// Return a Vec of immutable [`Component`] references filtered by base glyph name.
+    pub fn get_components_with_base(&self, basename: &str) -> Vec<&Component> {
+        self.components.iter().filter(|x| *x.base == *basename).collect::<Vec<&Component>>()
+    }
```
cmyr, line 127: Instead of returning a `Vec` (and allocating) I would prefer to return an `Iterator` and then give the caller the choice of allocating if necessary.

This ends up being slightly complicated for subtle lifetime reasons: we're returning an iterator that itself returns references to items owned by self, but the iterator *itself* also needs to outlive `basename`, since we need to compare with `basename` each time we `filter`.

The final complication is that we need to express that the two lifetimes are not independent: items borrowed from `self` must be valid for at least as long as the reference to `basename`. This means we end up with something like the following (it's possible there is a more concise way of expressing this, but I couldn't think of one):

```suggestion
    pub fn get_components_with_base<'b, 'a: 'b>(&'a self, basename: &'b str) -> impl Iterator<Item=&'a Component> + 'b {
        self.components.iter().filter(move |x| *x.base == *basename)
    }
```

cmyr (approved):
Sorry for letting this sit, a few little suggestions but these seem like very reasonable additions :)

src/glyph/mod.rs
chrissimpkins: Ty!

src/glyph/mod.rs
chrissimpkins: I started on this path but chickened out when I needed to add the `move` in order to make it work.  Does that move ownership out of `self.components` and can `self.components` be reused for access to the components data after  `get_components_with_base` execution?

cmyr: I suspect someone else could explain this better, but basically the move only implicates the items referenced *inside* the closure that are defined outside the closure's scope, so in this case the `basename` argument. It's definitely confusing in this case, but I would take it to mean that the whole `&str` is moved into the closure, instead of being referenced (as an `&&str`?)

src/glyph/tests.rs
```diff
@@ -608,3 +608,98 @@ fn pointtype_from_str_trait() {
...
+    let glyph = parse_glyph(bytes).expect("initial load failed");
+    assert!(glyph.has_component());
+
+    let bytes = include_bytes!("../../testdata/MutatorSansLightWide.ufo/glyphs/A_.glif");
+    let glyph = parse_glyph(bytes).expect("initial load failed");
+    assert!(!glyph.has_component());
+}
+
+#[test]
+fn component_number() {
```
cmyr, line 648:
```suggestion
fn component_count() {
```

src/glyph/tests.rs
```diff
@@ -608,3 +608,98 @@ fn pointtype_from_str_trait() {
...
+    let bytes = include_bytes!("../../testdata/MutatorSansLightWide.ufo/glyphs/A_dieresis.glif");
+    let glyph = parse_glyph(bytes).expect("initial load failed");
+    assert_eq!(glyph.component_number(), 2);
```
cmyr, line 651:
```suggestion
    assert_eq!(glyph.component_count(), 2);
```

src/glyph/tests.rs
```diff
@@ -608,3 +608,98 @@ fn pointtype_from_str_trait() {
...
+
+    let bytes = include_bytes!("../../testdata/MutatorSansLightWide.ufo/glyphs/A_.glif");
+    let glyph = parse_glyph(bytes).expect("initial load failed");
+    assert_eq!(glyph.component_number(), 0);
```
cmyr, line 655:
```suggestion
    assert_eq!(glyph.component_count(), 0);
```

cmyr (approved):
need to update the tests to reflect the name changes, I can do that from here I hope

chrissimpkins:
> need to update the tests to reflect the name changes, I can do that from here I hope

Thanks for this review Colin. No worries at all.  I'll update the tests to address the new method name and will update the `get_component_with_base` to return an iterator rather than a vector.  That will need test updates too.  I'll address it this week.

cmyr:
I updated the test names but didn't merge the iterator change. I'm going to do a release shortly, would you like this included?

chrissimpkins:
> I updated the test names but didn't merge the iterator change. I'm going to do a release shortly, would you like this included?

Feel free to release without these changes. I don't want to hold anything up and likely won't be able to address the tests until later this week.  I don't think that anything here is high priority.  Thanks Colin!

chrissimpkins:
All of Colin's changes were accepted, rebased on the recent changes in main (required f64 transition here), and updated tests. The CI should pass as of https://github.com/linebender/norad/pull/181/commits/d4493f7798181bbee5f4e4bc05dc30cbde2793c0.  If so, GTG from my standpoint.

cmyr (approved):
Thanks!

chrissimpkins:
Thanks for all of your help on this Colin!  Much appreciated.

Issue #164, opened 2021-08-19
Use f64 for floating point numbers


madig:
Python apparently uses f64 internally for floats but norad uses f32. This leads to vanilla Python code and norad to read the odd floating point coordinate differently (the f32 is upconverted to f64 in PyO3 bindings, changing the precision and making equality testing difficult). Maybe norad should use f64 where the spec talks about floats?

cmyr:
fine by me!

madig:
A designer notified me that this is especially annoying with e.g. guideline angles because they often happen to have a lot of decimal places. When multiple tools read and write those and keep higher precision, norad will always "shorten" them.

PR #184, opened 2021-09-27
Use f64 instead of f32 everywhere

madig:
Python floats as used by ufoLib are implicitly f64.

Closes https://github.com/linebender/norad/issues/164.

Maybe we need to change i32 to i64 as well at some point.


cmyr (approved):
looks good, thanks!

PR #185, opened 2021-09-28
Implement MetaInfo to spec

madig:
`formatVersionMinor` is new.

See https://unifiedfontobject.org/versions/ufo3/metainfo.plist/.

### DIFF
--- a/src/font.rs
+++ b/src/font.rs
@@ -66,6 +66,12 @@ pub enum FormatVersion {
 pub struct MetaInfo {
     pub creator: String,
     pub format_version: FormatVersion,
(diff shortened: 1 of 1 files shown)


cmyr:
Where does it say that creator is optional?

madig:
It didn't, but ufoLib does it that way. I opened https://github.com/unified-font-object/ufo-spec/pull/192 to clarify the spec.

madig:
Amended the PR to contain just the implementation of `formatVersionMinor` until the spec is clarified about the optionlaity of `creator`.

PR #186, opened 2021-09-29
Font::load: First check data request, then path

### LINKED ISSUE #182: Font::load test data request first, then for existence on disk
Instead of `let groups = if groups_path.exists() && request.groups` do `let groups = if request.groups && groups_path.exists()`. Not performance sensitive, but the right thing to do :)

madig:
Also fix a copy-paste error: check request.images for images.

Closes #182


cmyr (approved):
This seems reasonable :)

PR #190, opened 2021-09-30
MetaInfo: creator is optional now

madig:
See https://github.com/unified-font-object/ufo-spec/pull/192.

### DIFF
--- a/src/font.rs
+++ b/src/font.rs
@@ -64,7 +64,7 @@ pub enum FormatVersion {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "camelCase")]
 pub struct MetaInfo {
-    pub creator: String,
+    pub creator: Option<String>,
     pub format_version: FormatVersion,
     #[serde(default, skip_serializing_if = "is_zero")]
     pub format_version_minor: u32,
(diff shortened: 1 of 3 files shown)


src/font.rs
```diff
@@ -77,7 +77,7 @@ fn is_zero(v: &u32) -> bool {
 impl Default for MetaInfo {
     fn default() -> Self {
         MetaInfo {
-            creator: DEFAULT_METAINFO_CREATOR.to_string(),
+            creator: Some(DEFAULT_METAINFO_CREATOR.to_string()),
```
cmyr, line 80: Maybe just default to `None`?

src/font.rs
```diff
@@ -64,7 +64,7 @@ pub enum FormatVersion {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "camelCase")]
 pub struct MetaInfo {
-    pub creator: String,
+    pub creator: Option<String>,
```
cmyr, line 67: I think this is right, but in line with the other changes you've been making we could also just treat the empty string as none. (I prefer this way, though, it's more explicit and more "correct".

cmyr (commented):
A few thoughts; I don't particularly love this spec change and I don't really like changing a spec without some kind of spec version bump but those things don't matter, but if this is the consensus then let's certainly make the change :)

src/font.rs
madig: Mh, an empty string would still land in the serialized stream unless I stop serde; and the spec doesn't say that the field must be non-empty :thinking:

cmyr (approved):
looks good!

src/font.rs
madig: Hm, and set the creator on saving? I'd prefer to keep the explicit norad reference, since norad is the only user of it currently.

Issue #177, opened 2021-09-09
Have a bunch of benchmarks


madig:
I every now and then do profiling of UFO loading (and writing). I think it would be nice to have a bunch of benchmarks (https://doc.rust-lang.org/cargo/commands/cargo-bench.html) ready to be run. They could also serve as entry points for profilers. 

I'd say we could import a copy of Noto Sans and maybe even a custom to-UFO translation of Noto Sans CJK to have norad work up a sweat. Need to figure out the CJK part still, until the sources are opened.

Scenarios: 

- Serial loading of UFOs, no parallel loading of glifs (without `rayon` feature)
- Serial loading of UFOs, parallel loading of glifs (with `rayon` feature)
- Parallel loading of UFOs, no parallel loading of glifs
- Parallel loading of UFOs, parallel loading of glifs
- Loading a single UFO without any font or glyph libs
- Loading a single UFO where every glyph has a glyph lib and the font lib is big

chrissimpkins:
I really like this idea.  Are the GHA runners reliable enough envs to run these benchmarks?  If not, how would you standardize the execution/reporting?

madig:
I wasn't thinking about GHA actually, I'm not sure how good CI infrastructure is for relibale benchmarking. This is more aimed towards running benches on various machines easily to compare e.g. platform diffs. Benchmarking on commits does sound enticing, though...

chrissimpkins:
https://fast.vlang.io/ appears to use a free instance on AWS, possibly related to #175 too

cmyr:
I wouldn't benchmark on CI infrastructure, and generally wouldn't want to benchmark on a virtual machine. I do think benchmarks are important, although I would prefer [criterion](https://docs.rs/criterion/0.3.5/criterion/) to the built in cargo bench.

madig:
Another idea: look for quadratic runtime by having a massive CJK UFO and loading incrementally more of it and seeing if the timings form a line or upward slope.

chrissimpkins:
> having a massive CJK UFO

I looked into Noto CJK sources. They are not available and won't be in the near term.

chrissimpkins:
I wandered across this project from the Criterion developer that claims to be a way to support benchmark tests on CI infrastructures:

https://github.com/bheisler/iai

> Precision: High-precision measurements allow you to reliably detect very small optimizations to your code
Consistency: Iai can take accurate measurements even in virtualized CI environments
[...]

PR #192, opened 2021-10-10
WIP: add benchmarking

madig:
I'm playing a bit with criterion, as per https://github.com/linebender/norad/issues/177.

Needs https://github.com/madig/noto-amalgamated and the glyphs2ufo output for NotoSans-MM.glyphs.

### DIFF
--- a/src/data_request.rs
+++ b/src/data_request.rs
@@ -13,6 +13,7 @@
 pub struct DataRequest {
     pub layers: bool,
     pub lib: bool,
+    pub glyph_lib: bool,
     pub groups: bool,
     pub kerning: bool,
     pub features: bool,
@@ -22,7 +23,16 @@ pub struct DataRequest {
 
 impl DataRequest {
     fn from_bool(b: bool) -> Self {
-        DataRequest { layers: b, lib: b, groups: b, kerning: b, features: b, data: b, images: b }
+        DataRequest {
+            layers: b,
+            lib: b,
+            glyph_lib: b,
+            groups: b,
+            kerning: b,
+            features: b,
+            data: b,
+            images: b,
+        }
     }
 
     /// Returns a `DataRequest` requesting all UFO data.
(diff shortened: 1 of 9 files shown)


src/layer.rs
```diff
@@ -41,7 +41,11 @@ impl LayerSet {
     ///
     /// The `glyph_names` argument allows norad to reuse glyph name strings,
     /// reducing memory use.
-    pub fn load(base_dir: &Path, glyph_names: &NameList) -> Result<LayerSet, Error> {
+    pub fn load(
+        base_dir: &Path,
+        glyph_names: &NameList,
+        request: &crate::DataRequest,
```
cmyr, line 47: I would not include `DataRequest` as part of the base API here, especially if it is only being used to thread through whether or not we load the glyph libs?

src/data_request.rs
```diff
@@ -13,6 +13,7 @@
 pub struct DataRequest {
     pub layers: bool,
     pub lib: bool,
+    pub glyph_lib: bool,
```
cmyr, line 16: Is this a useful addition, or is it only useful in the context of benchmarking?

cmyr (commented):
benchmarks that involve IO are going to be especially difficult to interpret, since so much depends on hardware, read congestion, caching, etc etc. My inclination would be to just use the `load_save` example which does time those operations, which is useful enough for getting a loose sense of the impact of various changes?

If you want to do actual benchmarks, I think it is best to keep IO out of the benchmarks: load some files into memory first, and then benchmark the parsing.

I think it would also be useful to try and keep these benchmarks narrow in scope, e.g. benchmark reading just some `.glif` files, and benchmark just parsing `font_info.plist`, etc?

src/data_request.rs
madig: This was primarily added to test the impact of glyphlib loading on loading times. Result: it roughly doubles parsing time on my Linux and Windows machines. Ouch.

rsheeter:
This is very old, do you plan to finish it or could it be closed?

Issue #195, opened 2021-10-22
norad v1


chrissimpkins:
[List is being developed, this is not complete]

Collection of issues to reach v1 release status

- [x] Verify we have all validators from ufoLib on board. Component circular reference checking (from #31)
- [x] Enforce use of name Strings without control characters (from #31, tracker https://github.com/linebender/norad/issues/191)
- [x] More complete user name to filename conversion (from #31, tracker #85)
- [x] Better error reporting 
    - [x] https://github.com/linebender/norad/issues/137
- [ ] review / stabilize the serialization format
- [ ] https://github.com/linebender/norad/issues/194
- [x] Review/expand/clarify source documentation and add the [docs.rs doc link](https://docs.rs/norad/0.6.0/norad/) to repo README (https://github.com/linebender/norad/pull/197)
- [x] Update code to take advantage of newer plist crate (https://github.com/linebender/norad/issues/145)
- [x] Check we prevent downgrades (https://github.com/linebender/norad/issues/25)

Feel free to edit the list in this post.

madig:
@cmyr you wrote on Zulip:

> So the main question is, "what decisions are we unsure of?". One I can think of is the big FontInfo struct, which we might prefer to have be a dict (for instance)

what other things can you think of? My other idea was custom plist ser/de code to drop the slow xml-rs dependency, or maybe we want this upstream?

chrissimpkins:
I started source documentation updates in https://github.com/linebender/norad/pull/197

madig:
Opinions on #202 and #203?

madig:
I want to review our error architecture. Articles I need to ruminate about:

- https://www.unwoundstack.com/blog/rust-error-handling.html
- https://www.lpalmieri.com/posts/error-handling-rust/
- https://kazlauskas.me/entries/errors

madig:
Closing this after closing #31. There's grand stuff left to do, but no manpower to do it right now. Let's use https://github.com/linebender/norad/milestone/2 to track progress to a 1.0.

PR #197, opened 2021-11-08
Update public API docs

chrissimpkins:
Adds the `#![warn(missing_docs)]` attribute and full public API documentation

Related https://github.com/linebender/norad/issues/195

### DIFF
--- a/src/data_request.rs
+++ b/src/data_request.rs
@@ -5,18 +5,52 @@
 /// By default, we load all components of the UFO file; however if you only
 /// need some subset of these, you can pass this struct to [`Ufo::with_fields`]
 /// in order to only load the fields specified in this object. This can help a
-/// lot with performance with large UFO files if you don't need the glyph data.
+/// lot with performance in large projects when you don't need data intensive
+/// areas of the UFO font.
+///
+/// # Examples
+///
+/// A [DataRequest] that excludes all glyph layer, point, and kerning data:
+///
+/// ```
+/// use norad::DataRequest;
+///
+/// let datareq = DataRequest::default().layers(false).kerning(false);
+/// ```
+///
+/// A [DataRequest] that excludes all UFO data and images:
+///
+/// ```
+/// use norad::DataRequest;
+///
+/// let datareq = DataRequest::default().data(false).images(false);
+/// ```
+///
+/// A [DataRequest] that only includes parsed lib.plist data:
+///
+/// ```
+/// use norad::DataRequest;
+///
+/// let datareq = DataRequest::none().lib(true);
+/// ```
 ///
 /// [`Ufo::with_fields`]: struct.Ufo.html#method.with_fields
 #[derive(Clone, Copy, Debug, PartialEq)]
 #[non_exhaustive]
 pub struct DataRequest {
+    /// Load parsed glyph layers and points data
     pub layers: bool,
+    /// Load parsed lib.plist data
     pub lib: bool,
+    /// Load parsed groups.plist data
     pub groups: bool,
+    /// Load parsed kerning.plist data
     pub kerning: bool,
+    /// Load Adobe .fea format feature file data
     pub features: bool,
+    /// Load data
     pub data: bool,
+    /// Load images
     pub images: bool,
 }
 
@@ -25,12 +59,12 @@ impl DataRequest {
         DataRequest { layers: b, lib: b, groups: b, kerning: b, features: b, data: b, images: b }
     }
 
-    /// Returns a `DataRequest` requesting all UFO data.
+    /// Returns a [DataRequest] requesting all UFO data.
     pub fn all() -> Self {
         DataRequest::from_bool(true)
     }
 
-    /// Returns a `DataRequest` requesting no UFO data.
+    /// Returns a [DataRequest] requesting no UFO data.
     pub fn none() -> Self {
         DataRequest::from_bool(false)
     }
--- a/src/data_request.rs
+++ b/src/data_request.rs
@@ -59,7 +59,7 @@ impl DataRequest {
         DataRequest { layers: b, lib: b, groups: b, kerning: b, features: b, data: b, images: b }
     }
 
-    /// Returns a [DataRequest] requesting all UFO data.
+    /// Returns a [`DataRequest`] requesting all UFO data.
     pub fn all() -> Self {
         DataRequest::from_bool(true)
     }
--- a/src/data_request.rs
+++ b/src/data_request.rs
@@ -38,7 +38,7 @@
 #[derive(Clone, Copy, Debug, PartialEq)]
 #[non_exhaustive]
 pub struct DataRequest {
-    /// Load parsed glyph layers and points data
+    /// Load and parse all layers and glyphs.
     pub layers: bool,
     /// Load parsed lib.plist data
     pub lib: bool,
--- a/src/data_request.rs
+++ b/src/data_request.rs
@@ -64,7 +64,7 @@ impl DataRequest {
         DataRequest::from_bool(true)
     }
 
-    /// Returns a [DataRequest] requesting no UFO data.
+    /// Returns a [`DataRequest`] requesting no UFO data.
     pub fn none() -> Self {
         DataRequest::from_bool(false)
     }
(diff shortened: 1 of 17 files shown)


src/data_request.rs
```diff
@@ -5,18 +5,52 @@
...
+/// use norad::DataRequest;
+///
+/// let datareq = DataRequest::none().lib(true);
+/// ```
 ///
 /// [`Ufo::with_fields`]: struct.Ufo.html#method.with_fields
 #[derive(Clone, Copy, Debug, PartialEq)]
 #[non_exhaustive]
 pub struct DataRequest {
+    /// Load parsed glyph layers and points data
```
cmyr, line 41:
```suggestion
    /// Load and parse all layers and glyphs.
```

src/data_request.rs
```diff
@@ -25,54 +59,67 @@ impl DataRequest {
         DataRequest { layers: b, lib: b, groups: b, kerning: b, features: b, data: b, images: b }
     }
 
-    /// Returns a `DataRequest` requesting all UFO data.
+    /// Returns a [DataRequest] requesting all UFO data.
```
cmyr, line 62:
```suggestion
    /// Returns a [`DataRequest`] requesting all UFO data.
```

src/data_request.rs
```diff
@@ -25,54 +59,67 @@ impl DataRequest {
...
     pub fn all() -> Self {
         DataRequest::from_bool(true)
     }
 
-    /// Returns a `DataRequest` requesting no UFO data.
+    /// Returns a [DataRequest] requesting no UFO data.
```
cmyr, line 67:
```suggestion
    /// Returns a [`DataRequest`] requesting no UFO data.
```

src/data_request.rs
```diff
@@ -25,54 +59,67 @@ impl DataRequest {
...
     pub fn none() -> Self {
         DataRequest::from_bool(false)
     }
 
-    /// Request that returned UFO data include the glyph layers and points.
+    /// Returns a mutable reference to a [DataRequest] that includes
+    /// glyph layers and points if `b` is `true` or excludes these
+    /// data if `b` is `false`.
```
cmyr, line 74: Rust doc comments work a lot like git commits: the first line is used as a short summary (and is shown beside an item in the module level docs), and then anything after the first newline is "additional detail" and is only shown on the docs page for that specific item.

So basically: one short line that is a summary, and then anything else below the fold.

In this case though, I'm not sure that the new docs are an improvement on the previous ones? In particular I don't think it is necessary to mention that this returns `&mut Self`. This pattern should be familiar to the caller from `std` (see [`std::process::Command`](http://doc.rust-lang.org/1.56.0/std/process/struct.Command.html#method.stdin)). If the previous docstring was too vague, I might instead say,

```
/// Specify whether layers (and glyphs) should be loaded.
```

src/error.rs
```diff
@@ -17,35 +17,63 @@ pub enum Error {
...
     InvalidFontInfo,
+    /// An error returned when there is a problem during fontinfo.plist version up-conversion.
     FontInfoUpconversion,
+    /// An error returned when there is invalid groups.plist data.
     InvalidGroups(GroupsValidationError),
+    /// An error returned when there is a problem during groups.plist version up-conversion.
     GroupsUpconversionFailure(GroupsValidationError),
-    // the string is the key
+    /// An error returned when there is a problem parsing plist data into
+    /// [plist::Dictionary] types.  The string is the dictionary key.
```
cmyr, line 61:
```suggestion
    /// An error returned when there is a problem parsing plist data into
    /// [`plist::Dictionary`] types. 
    ///
    /// The string is the dictionary key.
```

src/error.rs
```diff
@@ -17,35 +17,63 @@ pub enum Error {
...
+    /// An error returned when there is a problem with kurbo contour conversion.
     #[cfg(feature = "kurbo")]
+    /// An error returned when there is a problem with contour conversion.
     ConvertContour(ErrorKind),
+    /// An error returned when there is a missing mandatory file.
     MissingFile(String),
+    /// An error returned when the requested UFO directory path is not present.
     MissingUfoDir(String),
+    /// An error returned when there is an invalid [crate::datastore::Store<T>] entry.
+    /// This error wraps a [StoreError] type and provides additional path data.
```
cmyr, line 76:
```suggestion
    /// An error returned when there is an invalid entry in an image or data store.
    ///
    /// This error wraps a [`StoreError`] type and provides additional path data.
```

cmyr (commented):
Some comments inline that you are free to ignore :)

chrissimpkins:
> Some comments inline that you are free to ignore :)

Ty!!

src/data_request.rs
chrissimpkins: Noted. I will have a look at this one tonight and respond.

chrissimpkins:
font module doc updates in https://github.com/linebender/norad/pull/197/commits/1462bca05c2f3715c13857d127ccdafde45bc009

chrissimpkins:
fontinfo docs https://github.com/linebender/norad/pull/197/commits/e038f7a9668e6bf7fc14b5c78342cb8882ac9acc

src/data_request.rs
chrissimpkins: Thanks for this information Colin.  I agree with your comments here and revised the initial draft in https://github.com/linebender/norad/pull/197/commits/fb78ef7c09a22e7fb941899726499a7b4f3cbe08

Please let me know if you feel that this improves the documentation in the section.

chrissimpkins: Related:

I've tended towards adding a `# Notes` header in documentation sections that have longer and/or multi-issue text content below the summary line.  IMO this sets it off nicely in the HTML site, but this is very subjective.  If you feel that this leads to unnecessary header clutter in the docs please let me know.  I should have a full first draft available by the end of the day so that we can have a look.

cmyr: I think that `std` is a great guideline, and we should try to match the conventions used there. They don't seem to use `# Note` as a header very often, but they do often have lines that begin with 'Note:', for providing information that is not important in the general case, but is important to a deeper understanding of the API: see for instance the docs for [`String::drain`](http://doc.rust-lang.org/1.56.0/std/string/struct.String.html#method.drain).

I do *think* I've seen `# Note:` before, but generally for a longform exposition of some idea or concept, but I can't find an example. The best I can come up with is the [A ghastly note 👻👻👻](http://doc.rust-lang.org/1.56.0/std/marker/struct.PhantomData.html#a-ghastly-note-) section of `PhantomData`.

chrissimpkins:
As of https://github.com/linebender/norad/pull/197/commits/30a30a88e7a49d2aca4b85bd75df0ab422ba2778, the entire public API is documented and clippy *should* no longer raise errors.  Converted out of Draft status.  This is ready for review.  I continue to work on the revisions noted in https://github.com/linebender/norad/pull/197#issuecomment-965391991

src/data_request.rs
chrissimpkins: Tyvm!  I'll do another pass through all of the new docs and remove unnecessary Note headers that do not follow this guideline.

chrissimpkins: This is how it appears in the HTML

chrissimpkins: vs. this after https://github.com/linebender/norad/pull/197/commits/314e2168eb3b9ab6eb846248cc990ef70038821a

chrissimpkins:
Rebased on #201

madig:
Question about the breadth of the docs: In e.g. a `Glyph`, the name should not contain control characters and in the list of Unicode codepoints, the first one is the primary codepoint for the glyph (see https://unifiedfontobject.org/versions/ufo3/glyphs/glif/). Should this information be included in the norad docs?

madig (approved):

chrissimpkins:
Should we merge this before any significant source changes land? I can keep an eye out and rebase / edit in this branch if you'd prefer to keep this open for now.

madig:
I'd like it merged so we have less work with subsequent PRs :) @cmyr ?

cmyr:
Sure let's get this in and we can always touch it up later :)

chrissimpkins:
Thank you both very much for the reviews!

PR #201, opened 2021-11-14
Refactor read and write configurations

### LINKED ISSUE #200: Read / write configuration builder consistency
The [`DataRequest` read configuration builder method approach](https://github.com/linebender/norad/blob/5a069856cf5dacf76b20a883e061d36ccc3822f7/src/data_request.rs#L38-L80) returns mutable struct references and works like so:

```rust
let datareq = DataRequest::default().layers(false).kerning(false).to_owned();

let ufo = Font::load_requested_data("path/to/font.ufo", datareq).expect("failed to load");
```

whereas the [`WriteOptions` serialization configuration builder method approach](https://github.com/linebender/norad/blob/6c558317ca727426b99a885600ad3249ffd8b6d6/src/write.rs#L74-L93) returns owned structs:

```rust
let spaces = WriteOptions::default().whitespace("  ");

ufo.save_with_options("path/to/out-font3.ufo", &spaces);
```

`load_requested_data` takes ownership and `save_with_options` borrows.

It might be worth keeping the approach consistent across these config struct builders and read/write implementations?

chrissimpkins:
Closes #200 

This PR refactors:

- the read and write configuration builder approaches to always return an owned configuration type across all builder methods
- the load and save methods where they are used to accept a reference parameter (always borrows)

The new approaches look like this:

### Configured Read

```rust
let datareq = DataRequest::default().layers(false).kerning(false);

let ufo = Font::load_requested_data("path/to/font.ufo", &datareq).expect("failed to load");
```

### Configured Write

```rust
let spaces = WriteOptions::default().whitespace("  ");

ufo.save_with_options("path/to/out-font3.ufo", &spaces);
```

### DIFF
--- a/src/font.rs
+++ b/src/font.rs
@@ -103,18 +103,18 @@ impl Font {
     ///
     /// [v3]: http://unifiedfontobject.org/versions/ufo3/
     pub fn load<P: AsRef<Path>>(path: P) -> Result<Font, Error> {
-        Self::load_requested_data(path, DataRequest::default())
+        Self::load_requested_data(path, &DataRequest::default())
     }
 
     /// Attempt to load the requested elements of a font object from a file.
     pub fn load_requested_data(
         path: impl AsRef<Path>,
-        request: DataRequest,
+        request: impl Borrow<DataRequest>,
     ) -> Result<Font, Error> {
-        Self::load_impl(path.as_ref(), request)
+        Self::load_impl(path.as_ref(), request.borrow())
     }
 
-    fn load_impl(path: &Path, request: DataRequest) -> Result<Font, Error> {
+    fn load_impl(path: &Path, request: &DataRequest) -> Result<Font, Error> {
         if !path.exists() {
             return Err(Error::MissingUfoDir(path.display().to_string()));
         }
@@ -211,7 +211,7 @@ impl Font {
             groups: groups.unwrap_or_default(),
             kerning: kerning.unwrap_or_default(),
             features,
-            data_request: request,
+            data_request: *request,
             data,
             images,
         })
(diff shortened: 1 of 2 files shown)


src/font.rs
```diff
@@ -103,18 +103,18 @@ impl Font {
...
     pub fn load<P: AsRef<Path>>(path: P) -> Result<Font, Error> {
-        Self::load_requested_data(path, DataRequest::default())
+        Self::load_requested_data(path, &DataRequest::default())
     }
 
     /// Attempt to load the requested elements of a font object from a file.
     pub fn load_requested_data(
         path: impl AsRef<Path>,
-        request: DataRequest,
+        request: impl Borrow<DataRequest>,
```
cmyr, line 112: I believe it is preferred to use `impl AsRef<T>` here, instead of `impl Borrow<T>`; `AsRef` can be implemented for more types, but `Borrow` only has a single implementor. As an example, if a method accepts `impl AsRef<Path>` it can accept `&str` or `&OsStr` in addition to `&Path`, `&PathBuf`, `PathBuf`, etc.

In this particular case they are functionally equivalent, but I'd still prefer to go with `AsRef`, if only to match the convention.


Another general question to ask about these things is whether or not there is a clear benefit. There are two potential downsides: the first is that this makes for slightly more awkward type signatures, and the second is that this could increase code size, if the caller ends up passing multiple concrete types, e.g. if they happen to use both `&DataRequest` and `DataRequest` in the same code base.

My hunch is that the second case is unlikely; I would expect the caller to use one or the other, and likely to only use this API in a single place anyway, so I think that this is fine. My own inclination would be to probably just accept `&DataRequest` here, but I think `AsRef<DataRequest>` is fine. In general, though, I would save the `AsRef` signature for places where it is likely to provide a clear benefit; a good example is `AsRef<Path>`, which allows the user to pass in some string type, such as from a command-line argument.

cmyr (approved):
Great, thanks! one note inline. :)

src/font.rs
chrissimpkins: Thanks Colin!

> Another general question to ask about these things is whether or not there is a clear benefit. 

Concerns noted.  I don't feel strongly about this change.  In my mind, it was simply to avoid breakage for anyone who has existing code that is based on an owned type parameter.  

> My own inclination would be to probably just accept &DataRequest here,

Happy to change this

cmyr: > Concerns noted. I don't feel strongly about this change. In my mind, it was simply to avoid breakage for anyone who has existing code that is based on the previous defaults.

I think this is good rationale! I'd be happy with either `&DataRequest` or `AsRef<DataRequest>`.

PR #209, opened 2021-11-24
Error tweaks

cmyr:
With this patch, we now track the path involved anytime we fail to load a `.glif` or a `.plist` file.

This also removes the position tracking for errors that occur when parsing glifs; there isn't a great way to get this information from `quick_xml`, and the information we *did* have was often inaccurate enough that it wasn't even on the correct line, which made it not very helpful for producing diagnostics.

@madig if you're interested in adding path information to other errors, I think you should be able to follow the pattern I've used here.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -170,17 +170,6 @@ impl InvalidColorString {
     }
 }
 
-/// An error representing a failure during .glif file parsing.
-#[derive(Debug)]
-pub struct GlifError {
-    /// The glif file path.
-    pub path: Option<PathBuf>,
-    /// The buffer position.
-    pub position: usize,
-    /// The kind of error.
-    pub kind: ErrorKind,
-}
-
 /// An error when attempting to write a .glif file.
 #[derive(Debug)]
 pub struct GlifWriteError {
--- a/src/error.rs
+++ b/src/error.rs
@@ -100,7 +100,7 @@ pub enum Error {
 #[derive(Debug)]
 pub enum GlifLoadError {
     /// An [`std::io::Error`].
-    IoError(IoError),
+    Io(IoError),
     /// A [`quick_xml::Error`].
     Xml(XmlError),
     /// The .glif file was malformed.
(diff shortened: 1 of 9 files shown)


src/glyph/parse.rs
```diff
@@ -69,26 +64,24 @@ impl<'names> GlifParser<'names> {
                     match tag_name.borrow() {
                         "outline" => {
                             // ufoLib parses `<outline/>` as an empty outline.
-                            self.builder
-                                .outline(Outline::default(), HashSet::new())
-                                .map_err(|e| err!(reader, e))?;
+                            self.builder.outline(Outline::default(), HashSet::new())?;
+                            //.map_err(|e| err!(reader, e))?;
```
madig, line 68: Is this a leftover?

src/error.rs
```diff
@@ -31,9 +31,19 @@ pub enum Error {
     },
     IoError(IoError),
     ParseError(XmlError),
-    Glif(GlifError),
+    Glif {
```
madig, line 34: Nitpick: should this be `GlifLoadError` and below `GlifWriteError`?

madig:
So the general pattern is, provide specific error types for some things and then wrap them with a path in the top-level `Error` type?

madig:
With this branch, chnaging a content.plist file entry to intentionally point nowhere, I now get

> `Glif { path: "C:\\Users\\nikolaus.waxweiler\\AppData\\Local\\Temp\\aaa\\MutatorSansLightWide.ufo\\glyphs\\E__.glif", inner: IoError(Os { code: 2, kind: NotFound, message: "The system cannot find the file specified." }`

That's better. The path implicitly also includes the layer name, though it could be presented a bit nicer maybe 🤔

madig:
Also, this makes the Error::ParseError variant redundant. Maybe I can even get rid of Error::IoError?

cmyr:
> So the general pattern is, provide specific error types for some things and then wrap them with a path in the top-level `Error` type?

This is a function of the code's "separation of concerns": the glyph parsing code does not know the path, it just receives a pointer to some bytes. This means that whatever error it returns cannot include the path, so we have to add that information later.

If the glif parsing code *did* happen to know the path, this would be unnecessary, and I would include the path in the error there. I'm not advocating this, though, since the separation of concerns is good for any number of reasons: as a simple example it means that we can test glyph parsing without needing to worry about having the files written to disk.

cmyr:
okay, fixed up and rebased :)

cmyr (commented):
I can't approve this since I opened it, but I'm happy to merge. :)

madig (approved):

madig:
Approved it for you 😁

PR #211, opened 2021-12-16
Use thiserror

### LINKED ISSUE #204: Errors: use https://lib.rs/crates/thiserror to reduce description duplication?
The doc updates sorta duplicate the meaning of the error message in the docstring _and_ the Display trait. Should we use thiserror in norad?

madig:
An experiment to see how using https://docs.rs/thiserror/latest/thiserror/ would look like.

Closes #204

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -116,14 +116,14 @@ pub fn write_plist_value_to_file(
     options: &WriteOptions,
 ) -> Result<(), Error> {
     let mut file =
-        File::create(path).map_err(|inner| Error::UfoWrite { path: path.into(), inner })?;
+        File::create(path).map_err(|source| Error::UfoWrite { path: path.into(), source })?;
     let writer = BufWriter::new(&mut file);
     value
         .to_writer_xml_with_options(writer, options.xml_options())
-        .map_err(|error| Error::PlistWrite { path: path.to_owned(), error })?;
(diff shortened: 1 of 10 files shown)


madig:
I suppose there are a few more errors in other modules that could be thiserror'ified if this is the approach we'd like.

src/write.rs
```diff
@@ -166,16 +173,16 @@ mod tests {
         let opt = WriteOptions::default();
         let plist_read = Value::from_file("testdata/MutatorSansLightWide.ufo/lib.plist")
             .expect("failed to read plist");
-        let tmp = TempDir::new("test")?;
+        let tmp = TempDir::new("test").unwrap();
```
cmyr, line 176: what's the motivation for these unwraps? If we're going to move to unwrapping then the test doesn't need to return a `Result`, but it seems easier to leave it as is?

cmyr (commented):
I'm happy to see this merged, I took a few quick measurements and compile time does not seem to be significantly impacted.

src/write.rs
madig: I didn't want to add a new `From` converter thing, but haven't looked at it too closely :grimacing:

cmyr: Sure, that's reasonable. Another option might be to make the return type `Box<dyn std::error::Error>`, but whatever is easiest. :)

cmyr (approved):
thanks!

PR #214, opened 2021-12-17
Use LayerInfo struct to deserialize layerinfo.plist

madig:
The new plist crate can hande nested `Value`s when deserializing, so we can now cut some code and an error.

Not sure yet what to do about serialization, because we have to do some post-processing.

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -2,7 +2,6 @@ use std::borrow::Borrow;
 use std::collections::BTreeMap;
 use std::fs;
 use std::path::{Path, PathBuf};
-use std::str::FromStr;
 use std::sync::Arc;
 
 #[cfg(feature = "rayon")]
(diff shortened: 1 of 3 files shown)


src/layer.rs
```diff
@@ -480,6 +454,20 @@ impl Default for Layer {
...
 }
 
+/// [UFO layer info], corresponding to a 'layerinfo.plist' file in a layer directory.
+///
+/// This will be merged into a [`Layer`].
+///
+/// [UFO layer info]: https://unifiedfontobject.org/versions/ufo3/glyphs/layerinfo.plist/
+#[derive(Debug, Clone, Deserialize)]
+struct LayerInfo {
+    /// Color field.
```
cmyr, line 464: there is no obligation to document a private type, and while I think it is useful to do this much of the time in this case it does not feel helpful.

src/layer.rs
```diff
@@ -480,6 +454,20 @@ impl Default for Layer {
     }
 }
 
+/// [UFO layer info], corresponding to a 'layerinfo.plist' file in a layer directory.
+///
+/// This will be merged into a [`Layer`].
+///
+/// [UFO layer info]: https://unifiedfontobject.org/versions/ufo3/glyphs/layerinfo.plist/
+#[derive(Debug, Clone, Deserialize)]
```
cmyr, line 462: Just a thought: given that this is just a helper type used during deserialization, I would probably put this all in a little helper fn like:


```rust
fn parse_layer_info(path: &Path) -> Result<(Option<Color>, Plist), SomeError> {
    #[derive(Deserialize)]
    struct LayerInfoHelper {
        color: Option<Color>,
        #[serde(default)]
        lib: Plist,
    }
    // do your logic here, and return the error or the contents
}
```

The idea is to make it clear that the type is a helper by keeping it out of the main scope, and having all of the code relevant for deserializing this type be defined in the same place; you can just look at this function and know exactly what's happening, without needing to jump around.

cmyr (approved):
cool, happy to delete some code! Some notes inline but nothing blocking, feel free to merge when you're happy. :)

src/layer.rs
madig: Good thinking. Maybe one day we don't have to do gymnastics to support different UFO normalization methods, then we can break it out again.

PR #220, opened 2021-12-23
Tweak Font object docs

cmyr:
This also makes data_request private, which I think is okay? It can be
provided when the font is loaded, but I don't know if it makes much
sense to be mutating it afterwards.

@ctrlcctrlv does this break anything for you?

### DIFF
--- a/src/font.rs
+++ b/src/font.rs
@@ -71,7 +71,7 @@ pub struct Font {
     pub groups: Groups,
     /// Horizontal kerning pairs, corresponding to the [`kerning.plist`][k] file.
     ///
-    /// This file is otional, and will not be serialized if no pairs are specified.
+    /// This file is optional, and will not be serialized if no pairs are specified.
     ///
     /// [k]: https://unifiedfontobject.org/versions/ufo3/kerning.plist/
     pub kerning: Kerning,
--- a/src/font.rs
+++ b/src/font.rs
@@ -31,46 +31,63 @@ static DEFAULT_METAINFO_CREATOR: &str = "org.linebender.norad";
 pub(crate) static DATA_DIR: &str = "data";
 pub(crate) static IMAGES_DIR: &str = "images";
 
+/// A font object, corresponding to a [UFO directory].
 /// A Unified Font Object.
 ///
 /// See the [UFO specification] for a description of the underlying data.
 ///
 /// [UFO specification]: https://unifiedfontobject.org/versions/ufo3/
+/// [UFO directory]: https://unifiedfontobject.org/versions/ufo3/index.html#directory-structure
 #[derive(Clone, Debug, Default, PartialEq)]
 #[non_exhaustive]
 pub struct Font {
-    /// [metainfo.plist][mi] parsed data field
+    /// The font's metainfo, corresponding to the [`metainfo.plist`][mi] file.
     ///
     /// [mi]: https://unifiedfontobject.org/versions/ufo3/metainfo.plist/
     pub meta: MetaInfo,
-    /// [fontinfo.plist][fi] parsed data field
+    /// The font info, corresponding to the [`fontinfo.plist`][fi] file.
     ///
     /// [fi]: https://unifiedfontobject.org/versions/ufo3/fontinfo.plist/
     pub font_info: FontInfo,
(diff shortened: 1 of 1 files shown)


src/font.rs
```diff
@@ -31,46 +31,63 @@ static DEFAULT_METAINFO_CREATOR: &str = "org.linebender.norad";
...
+    /// Glyph groups, corresponding to the [`groups.plist`][g] file.
+    ///
+    /// This file is optional; if no groups are specified it will not be serialized.
     ///
     /// [g]: https://unifiedfontobject.org/versions/ufo3/groups.plist/
     pub groups: Groups,
-    /// [kerning.plist][k] parsed data field
+    /// Horizontal kerning pairs, corresponding to the [`kerning.plist`][k] file.
+    ///
+    /// This file is otional, and will not be serialized if no pairs are specified.
```
madig, line 74:
```suggestion
    /// This file is optional, and will not be serialized if no pairs are specified.
```

madig:
I also don't think it is used for anything. I deleted the field and ran cargo test and it all worked 😬

cmyr:
It is very possible that we don't have a test that checks whether it works, but that's an oversight. See #53 for the original PR.

madig (approved):

madig:
Hm. Let's wait for @ctrlcctrlv to weight in before we remove it. Or we remove it and see if he screams 😁

ctrlcctrlv:
This doesn't break MFEKmetadata, no. I just tried to build with your master instead of with 0.6, and the only breakage is that `Font::load_requested_data` now takes `DataRequest` by reference instead of taking ownership of it, but that's probably a good change.

ctrlcctrlv:
If you're going to push this as a release to crates.io though you have to bump the version to 0.7 due to the API change.

ctrlcctrlv:
Although, the API change may not make perfect sense. It's called DataRequest, which makes me think that its ownership _should_ be lost by me if I pass it into `Font::load_requested_data`, since I'd assume I need to make a new request each time, and it also gets put in the struct, except now inaccessible. So you might consider not changing the API, then 0.6.1 would be fine…

madig:
Do you see any reason we should store the data request in Font? Is it used for anything outside loading a UFO?

ctrlcctrlv:
For me it seems much more logical to have the data request in the Font as `pub` because one could imagine a function that might receive a `Font` and expect it to have glyphs and then get confused when there aren't any. Having the DataRequest would allow it to know that the font may very well have glyphs but they were not loaded.

cmyr:
@ctrlcctrlv for me the bigger question is mutation. Is there a reasonable case where the user might mutate the `data_request` after creating the font? I could maybe come up with some contrived examples, but generally it feels unlikely. In particular, the case where you create a font with a `DataRequest` that (for instance) loads layers, what does it mean if the user manually change this to `false`, after creation? What are the user's expectations?


The way I see it: `DataRequest` is an accommodation  for certain advanced use-cases where the user wants to avoid the overhead of loading certain components of a font. I think it is reasonable to say that in these advanced cases, it is the responsibility of the user to know what they're doing. It seems like the alternative is a situation where any API that uses norad has to include a bunch of code-paths that check the state of the `data_request` before doing anything?

One possible compromise would be to include a `data_request()` method, that returns `&DataRequest`, and so doesn't support mutation, but I wouldn't want to add this API unless somebody will actually use it. Remember, any API included in 1.0 cannot be removed without releasing 2.0, but API can always be *added*.

ctrlcctrlv:
You are right, there is no good reason for mutation.

You also have to think a lot more about API coherence than I do, I regularly drop API's or change them in ways which would probably annoy third parties, but because my libraries aren't meant to necessarily be third party consumable revision to revision I don't worry too much about this.

madig:
Hm, then I'd say let's drop the field completely and just take a reference in the load function?

cmyr:
@ctrlcctrlv I tend to take your approach initially, but I'm currently hoping to make a stable release and so I want to address as many of those little quirks as I can before then. :)

cmyr:
@madig that... might work? In any case as long as it's private we can remove it later if it's easy.

PR #229, opened 2021-12-29
Take some load/save methods private

### LINKED ISSUE #225: Expose only Font and Glyph saving and loading methods
I noticed in https://github.com/linebender/norad/pull/224 that we have various public loading and saving methods that maybe don't have an audience. Exposing just the ones for Font and Glyph should be enough, and helps keeping some errors private.

madig:
They should probably not be used in normal operation.

Closes #225

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -40,7 +40,7 @@ impl LayerSet {
     ///
     /// The `glyph_names` argument allows norad to reuse glyph name strings,
     /// reducing memory use.
-    pub fn load(base_dir: &Path, glyph_names: &NameList) -> Result<LayerSet, Error> {
+    pub(crate) fn load(base_dir: &Path, glyph_names: &NameList) -> Result<LayerSet, Error> {
         let layer_contents_path = base_dir.join(LAYER_CONTENTS_FILE);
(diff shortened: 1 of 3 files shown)


cmyr (approved):
Thanks!

src/layer.rs
```diff
@@ -330,7 +332,7 @@ impl Layer {
     /// [`WriteOptions`] serialization format configuration.
     ///
     /// The path should not exist.
-    pub fn save_with_options(&self, path: &Path, opts: &WriteOptions) -> Result<(), Error> {
+    pub(crate) fn save_with_options(&self, path: &Path, opts: &WriteOptions) -> Result<(), Error> {
```
chrissimpkins, line 335: We are using save_with_options in ufofmt.  I think you do want this pub so users can modify their write formats?

cmyr: are you using it on `Layer`, explicitly? My thinking was that it should be public at the `Font` level, but not for individual layers.

chrissimpkins: Apologies.  I shouldn't review on the mobile app... :)

Only using with Font. Thanks!

PR #231, opened 2021-12-30
Simplify Glyph.load_object_libs with a macro

madig:
Saves some screen real estate.

### DIFF
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -148,45 +148,37 @@ impl Glyph {
     /// Move libs from the lib's `public.objectLibs` into the actual objects.
     /// The key will be removed from the glyph lib.
     fn load_object_libs(&mut self) -> Result<(), ErrorKind> {
+        // Use a macro to reduce boilerplate, to avoid having to mess with the typing system.
+        macro_rules! transfer_lib {
+            ($object:expr, $object_libs:expr) => {
+                if let Some(id) = $object.identifier().map(|v| v.as_str()) {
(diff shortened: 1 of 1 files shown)


src/glyph/mod.rs
```diff
@@ -148,45 +148,37 @@ impl Glyph {
     /// Move libs from the lib's `public.objectLibs` into the actual objects.
     /// The key will be removed from the glyph lib.
     fn load_object_libs(&mut self) -> Result<(), ErrorKind> {
+        // Use a macro to reduce boilerplate, to avoid having to mess with the typing system.
+        macro_rules! transfer_lib {
+            ($object:expr, $object_libs:expr) => {
+                if let Some(id) = $object.identifier().map(|v| v.as_str()) {
+                    if let Some(lib) = $object_libs.remove(id) {
```
cmyr, line 155: silly enough thing but I'd write this as,
```suggestion
                if let Some(lib) = $object.identifier().and_then(|v| $object_libs.remove(id.as_str())) {
```

this suggestion won't work if applied directly because the parens aren't balanced, and it really doesn't matter anyway, but thought I'd mention it.

cmyr (approved):
Looks good, always happy for some cleanup :)

src/glyph/mod.rs
madig: Ah, I need `id` a second time in the rejig-error PR to store it in the error. I'll leave this like it is for now and maybe we find a shorter way to write it in the rejigging PR.

Issue #95, opened 2021-03-30
Revamp GlyphBuilder


madig:
It is now easier to outfit a Glyph without all the `Option`s, so maybe it's time to revisit GlyphBuilder. It currently checks that:

1. ... all identifiers we use are unique
2. ... some methods are not called for format 1 glyphs (e.g. adding guidelines)
3. ... some methods are not called more than once

The last point is due to glif file parsing, some elements are supposed to occur at most once. I think I should move that logic back into the glif parser and turn `GlyphBuilder` into a simpler general builder that just ensures all identifiers are unique and only format-appropriate methods are called :thinking:  If there's a better way to do that, the builder can maybe go away completely. As long as there is no other use-case for the builder, the once-checking may as well stay, though.

Semi-related: The split GlyphBuilder <-> OutlineBuilder is somewhat problematic in that there needs to be a handover point for components, contours and identifiers (the latter need to be merged over so later builder commands can check against outline identifiers). If the builders are to be exposed to the outside, it would currently also require to expose the `Outline` intermediate struct for handover (making `OutlineBuilder` return components, contours and identifiers separately triggers clippy). This is... suboptimal. Maybe OutlineBuilder should be merged back into Glyphbuilder?

cmyr:
I think this bit of code is now pretty much all yours, so I'm happy to defer to your judgement. If there's anything in particular you want to talk through I'm happy to do that? I agree that a single `GlyphBuilder` type might make the most sense.

madig:
Running into this again while trying to revamp the error hierarchy in #224. There is a bit of a concern sharing going on between `(Glyph|Outline)Builder` and the `GlifParser`, e.g. the aforementioned v1 checking, that makes clear error separation difficult.

madig:
Also, `GlyphBuilder.finish()` needs no Result return type.

madig:
Thinking about it a bit more, maybe the not-called-more-than-once logic should also be handled while parsing, because it is part of the UFO XML specification. I'm not sure if there's a benefit to limiting the programmatic builder interface... which leaves  identifier checking, which is relevant for parsing _and_ programmatic glyph building. Identifiers are barely used as of now, but could benefit from a builder checking things behind the scenes. Otherwise, the GlyphBuilder API provides no discernible advantage over stuffing objects into a Glyph directly.

[...]

PR #235, opened 2022-01-05
Cut down builder

madig:
Reduces OutlineBuilder to the minimum and moves the responsibilities of GlpyhBuilder back into the parser.

Closes https://github.com/linebender/norad/issues/95.

### DIFF
--- a/src/glyph/parse.rs
+++ b/src/glyph/parse.rs
@@ -84,7 +84,7 @@ impl<'names> GlifParser<'names> {
                         seen_note = true;
                         self.parse_note(reader, buf)?
                     }
-                    _other => return Err(ErrorKind::UnexpectedTag.into()),
+                    _other => return Err(ErrorKind::UnexpectedElement.into()),
                 },
                 // The rest are expected to be empty element tags (exception: outline) with attributes.
                 Event::Empty(start) => match start.name() {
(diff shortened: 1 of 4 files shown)


src/glyph/parse.rs
```diff
@@ -46,43 +50,91 @@ impl<'names> GlifParser<'names> {
         raw_xml: &[u8],
         buf: &mut Vec<u8>,
     ) -> Result<Glyph, GlifLoadError> {
+        let mut seen_advance = false;
+        let mut seen_image = false;
+        let mut seen_lib = false;
+        let mut seen_note = false;
+        let mut seen_outline = false;
```
cmyr, line 57: can't we just check the various parts of the glyph object to see if they've been set?

madig: Well, outline, lib and advance.(height|width) aren't using Option, so we can't tell apart not-yet-encountered and default-value-of-zero-or-empty... but yes for image and note.

src/glyph/parse.rs
```diff
@@ -46,43 +50,87 @@ impl<'names> GlifParser<'names> {
...
-                        "lib" => self.parse_lib(reader, raw_xml, buf)?, // do this at some point?
-                        "note" => self.parse_note(reader, buf)?,
-                        _other => return Err(ErrorKind::UnexpectedTag.into()),
+                Event::Start(start) => match start.name() {
+                    b"outline" => {
+                        if seen_outline {
+                            return Err(ErrorKind::DuplicateElement("outline").into());
+                        }
+                        seen_outline = true;
+                        self.parse_outline(reader, buf)?
```
cmyr, line 66: is this clearer as,

```suggestion
                    b"outline" if seen_outline => return Err(ErrorKind::DuplicateElement("outline").into()),
                    b"outline" => {
                        seen_outline = true;
                        self.parse_outline(reader, buf)?
                    }
                        
```

cmyr (approved):
okay one more little note that you can take or leave, it applies to all these match statements; I find that sometimes using two matches with the conditional inline is clearer than having the conditional inside the match block?

src/glyph/parse.rs
madig: Let's try.

Issue #191, opened 2021-10-04
Stricter names for things


madig:
The spec contains wording like 

> The name of the glyph. This must be at least one character long. Names must not contain control characters. 

in various places, specifically for glyph, layer and group names (which therefore also extends to the contents of kerning.plist).

Maybe this can be enforce with a newtype `struct Name(String)` that implements a constructor that ensures the string has > 0 characters and contains no control characters.

This type would supersede `GlyphName` and `LayerName` or at least the `str` inside them.

cmyr:
Okay I like this, since it provides a way to abstract out the concrete type of a glyph name, I'll implement this now.

cmyr:
Okay, so i've dug into this a bit and there are a some decisions to make.

Because names now need to be validated, we need to decide what to do with API that take names, like `Glyph::new_named` or `Layer::get`. Currently these just take `&str`. I can see at least three possible options:

- make these methods fallible, and return an `Error` if the name is malformed.
- make these methods take a `Name`, and require the caller to construct a `Name` struct before calling
- allow these methods to take a `&str`, with the invariant that it is a valid identifer; panic if it is not.

There are maybe others, like taking `&str` and replacing invalid contents, but I think that is worse than any of these three options.

In practice, all of these feel bad in different ways. The first two feel bad because they make the API harder to use for a very unusual case. The characters that are disallowed in glyph names are all characters that you cannot type with a keyboard; constructing an invalid name would require specific effort on the part of the caller, so in practice I would expect the user to just `unwrap` everwhere. An API where you always `unwrap` is not helpful. The only case that could arise naturally would be when the user passes the empty string.

It's also possible that we could mix and match these approaches somewhat. Methods that already return an error could return an error in the bad-name case without it being any additional headache for the user, for instance.

In other cases, my preference is probably for number 3, where we panic explicitly. I think that since the user is going to unwrap in the case of the other designs (and so panic anyway) having us panic in a way that is clearly documented lets us keep the API useable.

I'll think about this a bit more, but wanted to write down my thoughts.

madig:
Option 3 sounds good. Not ideal, but eh. Another option would be to not deal with this at a type level and instead check it at load/save time. I would prefer a type-level solution though...

PR #240, opened 2022-01-07
Name type and related API changes

cmyr:
This adds a `Name` type, shared between glyphs and layers. This enforces the constraints on names documented in [the spec](https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#attributes) and also hides the concrete type we use to represent names, leaving us free to change that in future.

This also attempts to standardize the various places where names are used in the API. With this patch, most API that takes a name (for retrieving a glyph or a layer, for instance) now just takes a `&str`. In places where we are creating a glyph, this is converted into a `Name`, and panics if the name is malformed. This is unlikely for the reasons outlined at https://github.com/linebender/norad/issues/191#issuecomment-1006047293. Methods that may panic in this way are all documented.

There are a few exceptions:
- some methods that already return a `NamingError` will return `NamingError::Invalid` if the name is invalid, since it's easy enough and doesn't increase the user burden.
- `Component::new` takes a `Name` directly, because we expect you to take the name from an existing glyph.

There are a few other little tweaks included in this patch, which were related to these changes and would have been annoying to extract.

### DIFF
--- a/src/font.rs
+++ b/src/font.rs
@@ -2,18 +2,18 @@
 
 #![deny(rustdoc::broken_intra_doc_links)]
 
-use std::borrow::Borrow;
 use std::fs;
 use std::path::{Path, PathBuf};
 
 use crate::datastore::{DataStore, ImageStore};
 use crate::error::{FontLoadError, FontWriteError};
 use crate::fontinfo::FontInfo;
-use crate::glyph::{Glyph, GlyphName};
+use crate::glyph::Glyph;
 use crate::groups::{validate_groups, Groups};
 use crate::guideline::Guideline;
 use crate::kerning::Kerning;
 use crate::layer::{Layer, LayerSet, LAYER_CONTENTS_FILE};
+use crate::name::Name;
 use crate::names::NameList;
 use crate::shared_types::{Plist, PUBLIC_OBJECT_LIBS_KEY};
 use crate::upconversion;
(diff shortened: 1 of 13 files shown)


src/layer.rs
```diff
@@ -129,18 +131,18 @@ impl LayerSet {
...
-    pub fn names(&self) -> impl Iterator<Item = &LayerName> {
+    pub fn names(&self) -> impl Iterator<Item = &Name> {
         self.layers.iter().map(|l| &l.name)
     }
 
     /// Returns a new layer with the given name.
-    pub fn new_layer(&mut self, name: &str) -> Result<(), NamingError> {
-        if self.layers.iter().any(|l| &*l.name == name) {
-            Err(NamingError::Duplicate(name.into()))
+    pub fn new_layer(&mut self, name: &str) -> Result<&mut Layer, NamingError> {
```
madig, line 139: Oops! Did we forget returning something previously?

src/layer.rs
```diff
@@ -240,9 +252,10 @@ impl Layer {
     /// You generally shouldn't need this; instead prefer to load all layers
     /// with [`LayerSet::load`] and then get the layer you need from there.
     #[cfg(test)]
-    pub(crate) fn load(path: impl AsRef<Path>, name: LayerName) -> Result<Layer, LayerLoadError> {
+    pub(crate) fn load(path: impl AsRef<Path>, name: &str) -> Result<Layer, LayerLoadError> {
```
madig, line 255: Maybe needs a panic notice? In case we make it public again.

src/glyph/tests.rs
```diff
@@ -639,6 +639,22 @@ fn unexpected_line_after_offcurve2() {
     let _ = parse_glyph(data.as_bytes()).unwrap();
 }
 
+#[test]
+#[should_panic(expected = "InvalidName")]
+fn invalid_name() {
+    let data = "
```
madig, line 645: You could use `r#"..."#` so you don't have to escape stuff inside :)

src/font.rs
```diff
@@ -580,27 +580,20 @@ impl Font {
     }
 
     /// Returns an iterator over all the glyph names _in the default layer_.
-    pub fn iter_names(&self) -> impl Iterator<Item = GlyphName> + '_ {
+    pub fn iter_names(&self) -> impl Iterator<Item = Name> + '_ {
+        //FIXME: why not &Name here?
```
madig, line 584: What do you mean?

src/layer.rs
cmyr: good catch! yes, noticed while updating the signatures that this didn't return anything but probably should.

src/glyph/tests.rs
cmyr: Unfortunately this test case requires including a control character in the name, which has to be escaped, and so we can't use a raw string. 😢

src/font.rs
cmyr: I mean why does the iterator return `Name` instead of `&Name`? I vaguely recall there was a reason at some point but it doesn't make sense to me now.

madig: I also don't. 

```text
error[E0271]: type mismatch resolving `<std::collections::btree_map::Keys<'_, name::Name, std::sync::Arc<glyph::Glyph>> as std::iter::Iterator>::Item == &&name::Name`      
   --> src\font.rs:583:33
    |
583 |     pub fn iter_names(&self) -> impl Iterator<Item = &Name> + '_ {
    |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&name::Name`, found struct `name::Name`
    |
    = note: expected reference `&&name::Name`
               found reference `&name::Name`
    = note: required because of the requirements on the impl of `std::iter::Iterator` 
for `std::iter::Cloned<std::collections::btree_map::Keys<'_, name::Name, std::sync::Arc<glyph::Glyph>>>`

For more information about this error, try `rustc --explain E0271`.
```

madig (approved):

src/font.rs
cmyr: that's just because there's a call to `clone()` on the iterator that clones the `Names`.

PR #237, opened 2022-01-05
Support Glif formatMinor attribute

madig:
Closes #232

Issue #230, opened 2021-12-30
Remove `Font.meta.format_version(_minor)` and `Glyph.format` fields?


madig:
The current decided-on philosophy is that we support only the latest format versions and try to upgrade older ones. This makes making these versions accessible and changeable pointless and we currently error out on downgrades anyway.

Should they instead be read for loading and upgrading purposes and then discarded? Saving a UFO could take an additional parameter to specify the target version if we find a use-case for that.

This would require keeping up with minor UFO format versions though, otherwise we'd be reading 3.1 UFOs and write out 3.0 ones. But I suppose keeping up would be necessary anyway.

cmyr:
I think we can keep these, since it might be useful to know what the version was when it was loaded? and then if we write it out, we set it ourselves? 🤷

PR #241, opened 2022-01-07
Add minor version support to GlifVersion

cmyr:
If we find a version other than 1.0 or 2.0 we will error, but we will at
least avoid silently ignoring an unknown version.

This will clash with #240, so I'll fix it up once that is decided on.

### DIFF
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -22,14 +22,14 @@ use crate::{Color, Guideline, Identifier, Line, Plist, WriteOptions};
 
 /// A glyph, loaded from a [`.glif` file][glif].
 ///
+/// Norad can load glif version 1.0 and 2.0, and can save 2.0 only.
+///
 /// [glif]: http://unifiedfontobject.org/versions/ufo3/glyphs/glif/
 #[derive(Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "druid", derive(Lens))]
 pub struct Glyph {
     /// The name of the glyph.
     pub name: Name,
-    /// Glif file format version.
-    pub format: GlifVersion,
     /// Glyph height.
     pub height: f64,
     /// Glyph width.
@@ -86,9 +86,6 @@ impl Glyph {
         path: &Path,
         opts: &WriteOptions,
     ) -> Result<(), GlifWriteError> {
-        if self.format != GlifVersion::V2 {
-            return Err(GlifWriteError::Downgrade);
-        }
         if self.lib.contains_key(PUBLIC_OBJECT_LIBS_KEY) {
             return Err(GlifWriteError::PreexistingPublicObjectLibsKey);
         }
@@ -106,14 +103,15 @@ impl Glyph {
     /// panics if `name` is empty or if it contains any [control characters].
     ///
     /// [control characters]: https://unifiedfontobject.org/versions/ufo3/conventions/#controls
-    pub fn new_named(name: &str) -> Self {
-        Glyph::new(Name::new_raw(name), GlifVersion::V2)
+    pub fn new(name: &str) -> Self {
+        Glyph::new_impl(Name::new_raw(name))
     }
 
-    pub(crate) fn new(name: Name, format: GlifVersion) -> Self {
+    // this impl lets the crate pass an explicit `Name`, which is reused between
+    // multiple layers and components
+    pub(crate) fn new_impl(name: Name) -> Self {
         Glyph {
             name,
-            format,
             height: 0.0,
             width: 0.0,
             codepoints: Vec::new(),
@@ -239,7 +237,6 @@ impl Glyph {
 impl Data for Glyph {
     fn same(&self, other: &Glyph) -> bool {
         self.name.same(&other.name)
-            && self.format.same(&other.format)
             && self.height.same(&other.height)
             && self.width.same(&other.width)
             && self.codepoints == other.codepoints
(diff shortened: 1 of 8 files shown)


cmyr:
Maybe a less complicated approach would be to just treat these both as u32/u16 and stop trying to be fancy. 🤷

madig:
You mean making them a u32 (u16+u16) and using bit twiddling for getting and setting versions?

cmyr:
no I just mean having `version: u16` and `versionMinor: u16`, or maybe using `u32` for both.

The bigger question is: what should we be doing with these? The situation is a bit funny since we always upconvert. Thinking about this, it feels like the main question is what we should do with a glif with an unknown version. If it's an unknown major version we should just panic because we don't know anything, and if it's an unknown minor version we could either just parse what we know about *or* panic. In practice if there is a spec revision we will need to update things anyway, so my inclination is just to panic?

We also don't respect the version when serializing, so maybe we can actually keep this private? idk.

cmyr:
In any case I think this current patch is too fancy.

cmyr:
Okay I've rewritten this to get rid of `GlifVersion` completely; we now only track the version during parsing, in order to correctly validate the input, and then we discard it. When we save to disk, we were already only saving version 2.0 glyphs.

If there is a revision to the spec that introduces a new version, we can handle that then.

madig:
The sense or no sense is what I thought about in https://github.com/linebender/norad/issues/230 :) glad you see it too, now :D

src/glyph/parse.rs
```diff
@@ -119,7 +122,8 @@ impl<'names> GlifParser<'names> {
         }
 
         self.glyph.load_object_libs()?;
-        self.glyph.format = GlifVersion::V2;
+        // we upconvert 1->2, so set that here
+        //self.glyph.format.major = 2;
```
madig, line 126: Leftovers?

src/glyph/serialize.rs
```diff
@@ -52,7 +52,9 @@ impl Glyph {
         }
         let mut start = BytesStart::borrowed_name(b"glyph");
         start.push_attribute(("name", &*self.name));
-        start.push_attribute(("format", self.format.as_str()));
+        // we always serialize 2.0
+        //TODO: write out formatMinor if we start to support glif 2.1?
```
madig, line 56: Yes.

madig:
I kind of find it hard to judge if we should hard-code specific versions and hard-error on any minor versions we don't know until a new minor version appears and we know what it entails. But being strict and unyielding may be just fine.

src/glyph/parse.rs
```diff
@@ -560,19 +561,22 @@ impl<'names> GlifParser<'names> {
...
             Event::Decl(_decl) => (),
             Event::Start(ref start) if start.name() == b"glyph" => {
                 let mut name: Option<Name> = None;
-                let mut format: Option<GlifVersion> = None;
-                //let mut pos
+                let mut format_major = 0;
+                let mut format_minor = 0;
                 for attr in start.attributes() {
                     let attr = attr?;
                     let value = attr.unescaped_value()?;
```
madig, line 582: We can now remove the XXX comment below :)

src/glyph/parse.rs
```diff
@@ -18,8 +17,15 @@ pub(crate) fn parse_glyph(xml: &[u8]) -> Result<Glyph, GlifLoadError> {
     GlifParser::from_xml(xml, None)
 }
 
+// major, minor
+type Version = (u32, u32);
+
+const VERSION_1: (u32, u32) = (1, 0);
```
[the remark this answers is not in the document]
cmyr: oh yes definitely :)

cmyr:
Okay, updated this. I'm not sure what I think about `format_version` in metainfo. We could keep it there but just add docs saying that files are upconverted when loaded? Or we could make `font.meta` private, and have a separate method for getting the creator...

madig (approved):

src/glyph/serialize.rs
madig: I think this commend is outdated? We should indeed write out anything other than zero.

src/glyph/tests.rs
```diff
@@ -27,9 +27,33 @@ fn serialize_empty_glyph() {
     );
 }
 
+#[test]
```
madig, line 30: You may want to include the test cases I wrote in #237.

src/glyph/mod.rs
```diff
@@ -86,9 +86,6 @@ impl Glyph {
         path: &Path,
         opts: &WriteOptions,
     ) -> Result<(), GlifWriteError> {
-        if self.format != GlifVersion::V2 {
-            return Err(GlifWriteError::Downgrade);
```
madig, line 90: Actually, why remove the check here? I'd leave it in and change the Glyph element serializer to write out any minor version other than zero. Otherwise we may silently do something the user did not see coming.

madig: The `Downgrade` error could also be renamed to `UnsupportedGlifVersion` because a user could also request 2.123 or something.

src/glyph/serialize.rs
cmyr: We discard the version after loading (and error if it's something funny) so there's not much we can do here right now.

src/glyph/mod.rs
cmyr: We remove the check because we no longer store a version (we're always 2.0)

src/glyph/tests.rs
cmyr: those test cases don't really translate to this world where we don't hold on to the version after parsing.

PR #245, opened 2022-01-11
Use our errors in custom deserialize functions

### LINKED ISSUE #212: Use our errors in serde Results?
In custom `(de)serialize` functions, we currently use custom error strings a lot, e.g. in https://github.com/linebender/norad/blob/ddaaaf6dc5d50229d201af02bb475f7387df96ab/src/shared_types.rs#L70. I'd rather we use norad error enum variants instead. Need to investigate how.

madig:
Closes #212 

Results in errors like:

```
Error: failed to load file

Caused by:
    0: failed to load font info data
    1: failed to parse fontinfo.plist file
    2: Serde("invalid color string 'asbc'")
```

Not good, not bad...


cmyr (approved):
thanks!

PR #247, opened 2022-01-11
Harden Color

madig:


src/shared_types.rs
```diff
@@ -7,43 +7,70 @@ use serde::{Deserialize, Serialize};
...
     /// Blue channel value. Must be in the range 0 to 1, inclusive.
-    pub blue: f64,
+    blue: f64,
     /// Alpha (transparency) channel value. Must be in the range 0 to 1, inclusive.
-    pub alpha: f64,
+    alpha: f64,
+}
+
+impl Color {
+    /// Create a new color.
```
cmyr, line 32:
```suggestion
    /// Create a color with RGBA values in the range `0..=1.0`.
    ///
    /// Returns an error if any of the provided values are not in the allowed range.
```

cmyr (approved):
Thanks!

madig:
At first I thought about just checking this at write time but then I'd introduce one more point of failure and I'm trying to reduce those :)

PR #251, opened 2022-01-19
Implement full user name to file name conversion

### LINKED ISSUE #85: Extend crate::glyph::default_file_name_for_glyph_name to reference algorithm
Needs to be used for both glyph and layer filenames.

https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/ufoLib/filenames.py

### LINKED ISSUE #252: remove_glyph_raw missing self.content removal?
https://github.com/linebender/norad/blob/2789dbeedc64a84f41e87e1214c83e930685b5d2/src/layer.rs#L473-L491

Should remove_glyph_raw also contain a `self.contents.remove(name);`?

madig:
Closes #85
Closes #252

Alright, this took me a week to implement.

1. `Name::is_valid` must check `chars()` instead of `bytes()`, because the spec talks about "Any character with a Unicode value in the ranges [...]". E.g. the name "hello💝" encoded as bytes happens to include a banned control character but is a valid name.
2. `user_name_to_file_name` now...
    - takes a `&Name`, which guarantees some properties relevant for file name generation already
    - ensures a file name does not end with a period or space, as that's disallowed on Windows
    - ensures a file name is not using a Windows-reserved stem like "con", but unlike the official spec example does not replace more than necessary ("con.txt" must be turned into "_con.txt" but "hello.con.txt" is fine)
    - ensures a file name is unique among a case-insensitive set of existing file names. Or it panics after 99 attempts.
    - tries to not do something stupid if some form of `usize` over- or underflow occurs, e.g. by one of name, prefix or suffix being pathologically long. I'm not too confident about it though... I should probably fuzz it?
    - returns a PathBuf, because that's what's needed down the code chain. This allows me to simplify `Layer::new` and streamline its callers.
3. LayerSet and Layer are now responsible for providing `user_name_to_file_name` with a set of lowercased strings of existing file names for clash checking.
4. Layer::new now expects callers to do the Name instantiation work up front.

I think clashes are very unlikely, so this does feel a bit excessive... but I also can't rule them out completely. At least, I only try one strategy to resolve clashes instead of the two in the UFO spec example.

There is a discussion at https://github.com/unified-font-object/ufo-spec/issues/164 about a different and better file naming scheme, but so far without a definitive conclusion.

I need to look more closely at `Font::save()` to see how to implement erroring out when a file should be unique but the filesystem thinks differently. Should `Glyph::save_with_options()` error out if the destination exists already?

~~Also, random observation: do we need to change or regenerate file names when removing/renaming glyphs?~~ Doesn't seem like it.

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashSet};
 use std::fs;
 use std::path::{Path, PathBuf};
 
@@ -29,6 +29,10 @@ pub(crate) static DEFAULT_GLYPHS_DIRNAME: &str = "glyphs";
 pub struct LayerSet {
     /// A collection of [`Layer`]s.  The first [`Layer`] is the default.
     layers: Vec<Layer>,
+    /// A set of lowercased layer paths (excluding the default layer, as it is
+    /// always unique) for clash detection. This relies on Layer.path being
+    /// immutable.
+    path_set: HashSet<String>,
 }
 
 #[allow(clippy::len_without_is_empty)] // never empty
@@ -66,7 +70,7 @@ impl LayerSet {
             .ok_or(FontLoadError::MissingDefaultLayer)?;
         layers.rotate_left(default_idx);
 
-        Ok(LayerSet { layers })
+        Ok(LayerSet { layers, path_set: HashSet::new() })
     }
 
     /// Returns a new [`LayerSet`] from a `layers` collection.
@@ -75,7 +79,7 @@ impl LayerSet {
     pub fn new(mut layers: Vec<Layer>) -> Self {
         assert!(!layers.is_empty());
         layers.first_mut().unwrap().path = DEFAULT_GLYPHS_DIRNAME.into();
-        LayerSet { layers }
+        LayerSet { layers, path_set: HashSet::new() }
     }
 
     /// Returns the number of layers in the set.
@@ -125,7 +129,10 @@ impl LayerSet {
         } else if self.layers.iter().any(|l| l.name == name) {
             Err(NamingError::Duplicate(name.to_string()))
         } else {
-            let layer = Layer::new(name, None)?;
+            let name = Name::new(name).map_err(|_| NamingError::Invalid(name.into()))?;
+            let path = crate::util::default_file_name_for_layer_name(&name, &self.path_set);
+            let layer = Layer::new(name, path);
+            self.path_set.insert(layer.path.to_string_lossy().to_lowercase());
             self.layers.push(layer);
             Ok(self.layers.last_mut().unwrap())
         }
@@ -135,11 +142,18 @@ impl LayerSet {
     ///
     /// The default layer cannot be removed.
     pub fn remove(&mut self, name: &str) -> Option<Layer> {
-        self.layers
+        let removed_layer = self
+            .layers
             .iter()
             .skip(1)
             .position(|l| l.name.as_ref() == name)
-            .map(|idx| self.layers.remove(idx + 1))
+            .map(|idx| self.layers.remove(idx + 1));
+
+        if let Some(layer) = &removed_layer {
+            self.path_set.remove(&layer.path.to_string_lossy().to_lowercase());
+        }
+
+        removed_layer
     }
 
     /// Rename a layer.
@@ -148,8 +162,9 @@ impl LayerSet {
     /// be replaced.
     ///
     /// Returns an error if `overwrite` is false but a layer with the new
-    /// name exists, if no layer with the old name exists, or if the new name
-    /// is not a valid [`Name`].
+    /// name exists, if no layer with the old name exists, if the new name
+    /// is not a valid [`Name`] or when anything but the default layer should
+    /// be renamed to "public.default".
     pub fn rename_layer(
         &mut self,
         old: &str,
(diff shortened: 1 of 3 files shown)


src/layer.rs
```diff
@@ -458,8 +460,10 @@ impl Layer {
     ) {
         let glyph = glyph.into();
         if !self.contents.contains_key(&glyph.name) {
-            let path = crate::util::default_file_name_for_glyph_name(&glyph.name);
-            self.contents.insert(glyph.name.clone(), path.into());
+            let existing_paths: HashSet<String> =
+                self.contents.values().map(|p| p.to_string_lossy().to_lowercase()).collect();
+            let path = crate::util::default_file_name_for_glyph_name(&glyph.name, &existing_paths);
+            self.contents.insert(glyph.name.clone(), path);
```
cmyr, line 466: it feels really bad to be creating this collection anew every time a glyph is inserted; think about the situation where you're making a layer and adding a bunch of glyphs, you'll be rebuilding this set each time?

I would prefer to just maintain a set of existing paths or something?

src/name.rs
```diff
@@ -45,9 +45,13 @@ impl Name {
 
 fn is_valid(name: &str) -> bool {
     !(name.is_empty()
-        || name
-            .bytes()
-            .any(|b| (0x0..=0x1f).contains(&b) || (0x80..=0x9f).contains(&b) || b == 0x7f))
+        // Important: check the chars(), not the bytes(), as UTF-8 encoding
```
cmyr, line 48: yep this is a good catch!

src/util.rs
```diff
@@ -12,87 +16,289 @@ pub(crate) fn recursive_sort_plist_keys(plist: &mut plist::Dictionary) {
...
+                boundary -= 1;
+            }
+            result.truncate(boundary);
+        } else {
+            // Cutting off the suffix should land on a `char` boundary.
+            result.truncate(result.len().saturating_sub(suffix.len()));
+        }
+
+        let mut found_unique = false;
+        for counter in 1..1_000_000_000_000_000u64 {
```
cmyr, line 148: I am flabbergasted that this is the recommended strategy for determining a name. 

using some napkin math: if we assume the average length of an existing name is 8 bytes (and we need 15 bytes to store at least the numbers, for the larger numbers) then you would need something like 7000 terabytes of ram in order to just keep the *existing* names in memory.

similarly: if each check takes a microsecond (and I assume it's longer) checking all 10^15 possible names would take ~27 hours.

cmyr (commented):
this is certainly something, huh.


some thoughts inline: 10^15 is definitely way more checks than could possibly make sense, but whatever.

More importantly, I think constructing the set of paths each time is a mistake; we should just maintain a set of normalized paths, similarly to how we maintain a set of names.

src/util.rs
madig: I simply copied the approach from the spec, see https://unifiedfontobject.org/versions/ufo3/conventions/#common-user-name-to-file-name-algorithm. I agree that it's trying a bit hard.

I had the idea of appending a UUIDv4 and calling it a day, that'd also solve the existing paths set problem. But uh.

src/layer.rs
madig: I implemented a HashSet cache for LayerSet and Layer now. Please have a look.

madig:
Needs to be rebased on top of #254 if that goes through :)

src/util.rs
cmyr: yea, I see that this is the spec, and I think it's fine to follow it! I just think it's silly that this is what the spec demands.

madig: I think it's more of a suggestion. The spec itself notes that there is no definitive scheme.

cmyr: if it's a suggestion then let's just stop at 1000 or so, since if this situation is happing it's probably a bug and it's better to crash then to hang.

src/layer.rs
```diff
@@ -195,12 +213,21 @@ impl LayerSet {
...
+            } else {
+                // Non-default layer.
+                let old_path = self.layers[layer_pos].path.to_string_lossy().to_lowercase();
+                self.path_set.remove(&old_path);
+                let new_path = crate::util::default_file_name_for_layer_name(&name, &self.path_set);
+                self.path_set.insert(new_path.to_string_lossy().to_lowercase());
+
+                let layer = &mut self.layers[layer_pos];
+                layer.name = name;
+                layer.path = new_path;
```
cmyr, line 229: ditto, this whole block can just be:

```rust
let layer_pos = self.layers.iter().position(|l| l.name.as_ref() == old).unwrap();

// change path only if not the default layer:
if layer_pos != 0 {
    let old_path = self.layers[layer_pos].path.to_string_lossy().to_lowercase();
    self.path_set.remove(&old_path);
    let new_path = crate::util::default_file_name_for_layer_name(&name, &self.path_set);
    self.path_set.insert(new_path.to_string_lossy().to_lowercase());
    self.layers[layer_pos].path = new_path;
}

self.layers[layer_pos].name = name;
```

do you understand the rationale here? it's about having the only code in the conditional block be the code that is 'special' to the condition, which makes the code easier to read overall; this is a good practice to follow whenever possible.

src/util.rs
cmyr: Thinking about this more: we *do* already have the UUID crate, and I think you were on a good track when you mentioned using that. It should really simplify the code, and hopefully also help binary size?

what I would try here:

- do the 'increment' trick for a small number of integers (9? 99?)
- if that fails, try with a UUID
- if that fails, panic.


If you don't like this then we don't have to do it, but it feels much cleaner to me?

cmyr (commented):
Okay a few more little comments inline but I think this is close. Thanks!

src/layer.rs
madig: Ah duh, I didn't see that. Thanks!

src/util.rs
madig: I'll try a 100 times and then just panic. I think we can still add UUIDv4 if we ever panic in the real world :) Thanks for the suggestion!

madig:
I noticed that inserting something to escape a reserved name could push the string over 255 characters. Maybe I need to do the escaping before.

cmyr (approved):
Looks good, thanks!

madig:
One day I must fuzz this :thinking:

PR #254, opened 2022-01-24
Stricter layer renaming handling

### LINKED ISSUE #253: LayerSet.rename_layer does not set new path
A new path should be generated when renaming layers, just like for glyphs.

madig:
Broken out of #251 to ease reviewing.

1. Layer renaming should assign new paths as well, ...
2. ... except the default layer must always have the default layer path.
3. Protect the default layer name harder, it must only ever be assigned to the default layer.

Closes #253

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -704,10 +704,12 @@ mod tests {
         assert_eq!(layer_set.get("aaa").unwrap().path().as_os_str(), "glyphs.aaa");
 
         layer_set.rename_layer("aaa", "bbb", false).unwrap();
+        assert!(layer_set.get("aaa").is_none());
         assert_eq!(layer_set.get("bbb").unwrap().path().as_os_str(), "glyphs.bbb");
 
         layer_set.rename_layer("bbb", "aaa", false).unwrap();
         assert_eq!(layer_set.get("aaa").unwrap().path().as_os_str(), "glyphs.aaa");
+        assert!(layer_set.get("bbb").is_none());
     }
 
     #[test]
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -100,13 +100,20 @@ impl LayerSet {
     ///
     /// # Panics
     ///
-    /// panics if `name` is empty or if it contains any [control characters].
+    /// panics if
+    /// 1. `name` is empty or if it contains any [control characters].
+    /// 2. `name` is the default layer name but the default layer has a
+    ///    different name. Use [`Self::default_layer`] or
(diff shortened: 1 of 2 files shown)


src/layer.rs
```diff
@@ -100,13 +100,20 @@ impl LayerSet {
     ///
     /// # Panics
     ///
-    /// panics if `name` is empty or if it contains any [control characters].
+    /// panics if
+    /// 1. `name` is empty or if it contains any [control characters].
+    /// 2. `name` is the default layer name but the default layer has a
+    ///    different name. Use [`Self::default_layer`] or
+    ///    [`Self::default_layer_mut`] instead.
```
cmyr, line 107: with this getting so complicated I wonder if we should just remove `get_or_create`?

src/layer.rs
```diff
@@ -175,17 +184,29 @@ impl LayerSet {
...
+
+            if layer_pos == 0 {
+                // Default layer: just change the name.
+                let layer = &mut self.layers[layer_pos];
+                layer.name = name;
+            } else {
+                // Non-default layer: change name and path.
+                let layer = &mut self.layers[layer_pos];
+                layer.path = crate::util::default_file_name_for_layer_name(&name).into();
+                layer.name = name;
```
cmyr, line 206: not a blocker but I find it more readable in general to use conditional blocks to only perform the conditional work, e.g:

```rust
let layer_pos = self.layers.iter().position(|l| l.name.as_ref() == old).unwrap();
if layer_pos != 0 {
    self.layers[layer_pos].path =
        crate::util::default_file_name_for_layer_name(&name).into();
}
self.layers[layer_pos].name = name;
```

src/layer.rs
cmyr: Alternatively we could just return the default layer if the user passes the default layer name, even if the default layer has another name? idk

cmyr (approved):
Looks good, a couple little comments inline. Is `get_or_create` worth keeping? It was intended as a convenience method but now it's just seeming uglier and uglier :/

src/layer.rs
madig: Hm. I made the method for two particular calls in my custom to-UFO converter thing. I suppose I could make it work without the method. Shall I remove it in this PR?

src/layer.rs
madig: (I may have to undo this in the filenames PR but let's see.)

src/layer.rs
madig: I'll do it in a separate PR.

PR #255, opened 2022-01-26
Remove LayerSet::get_or_create

madig:
As noted in https://github.com/linebender/norad/pull/254#discussion_r791810131, I think the method serves no valuable purpose currently but expands the surface for API mistakes. Remove it for now; it can be added back when we find a good use case.


cmyr:
Agreed. I think I added this for use in Runebender, but it would be easier for me to just have a separate helper function to handle this on my side.

cmyr (approved):

PR #257, opened 2022-02-17
Use Name for glyph parts

madig:
Because no control characters. Found this while skimming the spec. Oops :D But also, these parts don't have a minimum length specified, but empty names make no sense. May need to open a discussion on the spec tracker.

Also, how did we not find the "Glyph::new_named" issue in one of the tests?

Edit: opened issue over at https://github.com/unified-font-object/ufo-spec/issues/206

### DIFF
--- a/src/name.rs
+++ b/src/name.rs
@@ -2,7 +2,7 @@
 
 use std::sync::Arc;
 
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::{Deserialize, Deserializer};
 
 use crate::error::NamingError;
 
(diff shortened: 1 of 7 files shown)


src/name.rs
```diff
@@ -92,6 +92,18 @@ impl std::borrow::Borrow<str> for Name {
     }
 }
 
+// NOTE: This custom function ensures that serde serializes to a `Token::Str`
+// rather than `Token::NewType` that appears with a derived `Serialize`.
+impl Serialize for Name {
```
cmyr, line 97: can we use [#[serde(transparent)]](https://serde.rs/container-attrs.html#transparent)?

cmyr (approved):
Looks good, what `new_named` issue do you mean?

madig:
This: https://github.com/linebender/norad/pull/257/files#diff-5eab58a7f7ef848629a8be1084b12464ced845ebd69087b9cbc1a2263465212aL804 -- the `new_named` method does not exist anymore.

PR #260, opened 2022-02-21
Implement FromStr for Name

madig:
Useful when you want to deserialize into Name in clap.

### DIFF
--- a/src/identifier.rs
+++ b/src/identifier.rs
@@ -1,4 +1,5 @@
 use std::hash::Hash;
+use std::str::FromStr;
 use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


src/identifier.rs
```diff
@@ -90,6 +91,14 @@ impl std::borrow::Borrow<str> for Identifier {
     }
 }
 
+impl FromStr for Identifier {
+    type Err = ErrorKind;
```
cmyr, line 95: I don't love using `ErrorKind` as an actual error, since it doesn't implement the `Error` trait.

Can we think of any reason why we would actually need this impl? We can just cut it for now, and keep `Name`...

cmyr (approved):
thanks!

PR #261, opened 2022-03-03
Use Name instead of String in Groups and Kerning

madig:
I think I missed this in https://github.com/linebender/norad/pull/257.


cmyr (approved):
Thanks!

PR #265, opened 2022-03-16
Ignore `path_set` when equality testing LayerSet

madig:
Found this while testing for equality when loading a layer set vs. recreating it manually.

### DIFF
--- a/src/fontinfo.rs
+++ b/src/fontinfo.rs
@@ -659,7 +659,7 @@ impl FontInfo {
                         .unitsPerEm
                         .map(|v| NonNegativeIntegerOrFloat::new(v.abs()).unwrap()),
                     version_major: fontinfo_v2.versionMajor,
-                    version_minor: fontinfo_v2.versionMinor.map(|v| v.abs() as NonNegativeInteger),
+                    version_minor: fontinfo_v2.versionMinor.map(|v| v.unsigned_abs()),
                     x_height: fontinfo_v2.xHeight,
                     year: fontinfo_v2.year,
                     ..FontInfo::default()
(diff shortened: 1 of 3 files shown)


cmyr:
SO 

> Colin says:
> 
> > a custom PartialEq impl that ignores pathset would be reasonable"
> 
> but I don't know whether this will have observable differences in behavior in some edge cases... Need to check the code.

So regardless of current behaviour, this comes down to us needing to decide what 'equality' means, in the context of a layer. If a layer is contains exactly the same glyphs as another layer, but has a different path, is it a different layer? (probably yes?)

In any case: to my eye, `path_set` is an implementation detail, and doesn't seem important to determining equality. All that should matter is equality of name, path, info, and the glyph set? Let me know if you see things differently :)

src/fontinfo.rs
```diff
@@ -1353,16 +1353,16 @@ impl<'de> Deserialize<'de> for Os2Panose {
...
+            family_type: value.family_type.unsigned_abs(),
+            serif_style: value.serif_style.unsigned_abs(),
+            weight: value.weight.unsigned_abs(),
+            proportion: value.proportion.unsigned_abs(),
+            contrast: value.contrast.unsigned_abs(),
+            stroke_variation: value.stroke_variation.unsigned_abs(),
+            arm_style: value.arm_style.unsigned_abs(),
+            letterform: value.letterform.unsigned_abs(),
+            midline: value.midline.unsigned_abs(),
+            x_height: value.x_height.unsigned_abs(),
```
cmyr, line 1365: are these changes related to this patch?

cmyr (approved):
I'm curious about some of the stuff in here which seems maybe stale? but over all the fix makes sense!

madig:
What seems stale to you?

src/fontinfo.rs
cmyr: ahhh okay I get it, I was confused by `NonNegativeInteger`.

As a general thing I like to do clippy fixups as separate patches, since that way if we end up wanting to revert a commit we don't suddenly also get clippy fixes rolled-back, but no worries :)

PR #267, opened 2022-07-13
Take glyph name crate-public

### LINKED ISSUE #266: `Layer::get_glyph_mut` allows user to change glyph name without updating `Layer.glyphs`
I think. `Glyph.name` should probably be taken private, so that you have to use `Layer::rename_glyph`?

madig:
Closes #266

### DIFF
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -29,6 +29,7 @@ use crate::{Color, Guideline, Identifier, Line, Plist, WriteOptions};
 #[cfg_attr(feature = "druid", derive(Lens))]
 pub struct Glyph {
     /// The name of the glyph.
+    #[cfg_attr(feature = "druid", lens(ignore))]
     pub(crate) name: Name,
     /// Glyph height.
     pub height: f64,
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -29,7 +29,7 @@ use crate::{Color, Guideline, Identifier, Line, Plist, WriteOptions};
 #[cfg_attr(feature = "druid", derive(Lens))]
 pub struct Glyph {
     /// The name of the glyph.
-    pub name: Name,
+    pub(crate) name: Name,
     /// Glyph height.
     pub height: f64,
     /// Glyph width.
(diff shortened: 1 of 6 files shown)


src/glyph/mod.rs
```diff
@@ -125,6 +126,11 @@ impl Glyph {
         }
     }
 
+    /// Returns the name of the glyph.
+    pub fn name(&self) -> Name {
+        self.name.clone()
+    }
```
cmyr, line 132: Although this clone is not expensive, I think as a general pattern it's good to leave this decision up to the caller, unless we are returning these as cloned values in other places?

```suggestion
    pub fn name(&self) -> &Name {
        &self.name
    }
```

src/glyph/mod.rs
```diff
@@ -29,7 +29,8 @@ use crate::{Color, Guideline, Identifier, Line, Plist, WriteOptions};
 #[cfg_attr(feature = "druid", derive(Lens))]
 pub struct Glyph {
     /// The name of the glyph.
-    pub name: Name,
+    #[cfg_attr(feature = "druid", lens(ignore))]
```
cmyr, line 32:
```suggestion
```

I don't *think* this is necessary, I think non-pub items are ignored by default? In any case I can fix this up if/when I pick up runebender next.

tests/save.rs
```diff
@@ -62,8 +62,8 @@ fn save_fancy() {
     assert_eq!(pre_layer.iter().count(), post_layer.iter().count());
 
     for glyph in pre_layer.iter() {
-        let other = post_layer.get_glyph(&glyph.name);
-        assert!(other.is_some(), "missing {}", &glyph.name);
+        let other = post_layer.get_glyph(&glyph.name());
+        assert!(other.is_some(), "missing {}", &glyph.name());
```
cmyr, line 66: If we're already returning a reference from the method, we won't need to take a reference here:

```suggestion
        let other = post_layer.get_glyph(glyph.name());
        assert!(other.is_some(), "missing {}", glyph.name());
```

cmyr (approved):
Some notes but this looks good 👍

src/glyph/mod.rs
madig: I get a clippy/test failure unless I put that line in, because the macro seems to want to make a name() function, which already exists.

PR #268, opened 2022-07-18
Deduplicate codepoints on read and write

### LINKED ISSUE #263: GLIF load/save: Filter duplicate codepoints
See https://github.com/unified-font-object/ufo-spec/pull/202.

madig:
Closes #263

### DIFF
--- a/src/glyph/parse.rs
+++ b/src/glyph/parse.rs
@@ -122,6 +122,12 @@ impl<'names> GlifParser<'names> {
         }
 
         self.glyph.load_object_libs()?;
+
+        let mut seen_codepoints = HashSet::new();
+        let deduplicated_codepoints =
+            self.glyph.codepoints.iter().filter(|c| seen_codepoints.insert(**c)).cloned().collect();
+        self.glyph.codepoints = deduplicated_codepoints;
+
         Ok(self.glyph)
     }
 
--- a/src/glyph/parse.rs
+++ b/src/glyph/parse.rs
@@ -123,10 +123,18 @@ impl<'names> GlifParser<'names> {
 
         self.glyph.load_object_libs()?;
 
-        let mut seen_codepoints = HashSet::new();
-        let deduplicated_codepoints =
-            self.glyph.codepoints.iter().filter(|c| seen_codepoints.insert(**c)).cloned().collect();
-        self.glyph.codepoints = deduplicated_codepoints;
+        // Deduplicate code points while preserving order, but only if there is more
+        // than 1 codepoint, to avoid memory allocation.
+        if self.glyph.codepoints.len() > 1 {
+            let mut seen_codepoints = HashSet::new();
+            self.glyph.codepoints = self
+                .glyph
+                .codepoints
+                .iter()
+                .filter(|c| seen_codepoints.insert(**c))
+                .cloned()
+                .collect();
+        }
 
         Ok(self.glyph)
     }
--- a/src/glyph/parse.rs
+++ b/src/glyph/parse.rs
@@ -127,13 +127,7 @@ impl<'names> GlifParser<'names> {
         // than 1 codepoint, to avoid memory allocation.
         if self.glyph.codepoints.len() > 1 {
             let mut seen_codepoints = HashSet::new();
-            self.glyph.codepoints = self
-                .glyph
-                .codepoints
-                .iter()
-                .filter(|c| seen_codepoints.insert(**c))
-                .cloned()
-                .collect();
+            self.glyph.codepoints.retain(|c| seen_codepoints.insert(*c));
         }
 
         Ok(self.glyph)
--- a/src/glyph/parse.rs
+++ b/src/glyph/parse.rs
@@ -123,13 +123,6 @@ impl<'names> GlifParser<'names> {
 
         self.glyph.load_object_libs()?;
 
-        // Deduplicate code points while preserving order, but only if there is more
-        // than 1 codepoint, to avoid memory allocation.
-        if self.glyph.codepoints.len() > 1 {
-            let mut seen_codepoints = HashSet::new();
-            self.glyph.codepoints.retain(|c| seen_codepoints.insert(*c));
-        }
-
         Ok(self.glyph)
     }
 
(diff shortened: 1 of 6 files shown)


src/glyph/parse.rs
```diff
@@ -122,6 +122,12 @@ impl<'names> GlifParser<'names> {
         }
 
         self.glyph.load_object_libs()?;
+
+        let mut seen_codepoints = HashSet::new();
+        let deduplicated_codepoints =
+            self.glyph.codepoints.iter().filter(|c| seen_codepoints.insert(**c)).cloned().collect();
+        self.glyph.codepoints = deduplicated_codepoints;
+
```
cmyr, line 130: In general, how many codepoints do we expect a glyph to have? I would expect very few, so it might be simpler to just write this as:

```suggestion
        self.glyph.codepoints.sort_unstable();
        self.glyph.codepoints.dedup();
```

it's possible that this would even be better in the case where the collection might be larger, but my intuition there is not well formed.

src/glyph/serialize.rs
```diff
@@ -57,8 +60,11 @@ impl Glyph {
         start.push_attribute(("format", "2"));
         writer.write_event(Event::Start(start)).map_err(GlifWriteError::Xml)?;
 
+        let mut seen_codepoints = HashSet::new();
```
cmyr, line 63: maybe we just enforce deduplication whenever the user modifies the codepoints? As in we always keep them sorted, and we don't insert dupes (we can use `slice::binary_search` to find the insertion point, or whether the item already exists)

cmyr (approved):
Oops sorry about this, slipped by me somehow. Might want to rebase on #274 when that gets in, but overall this makes sense. I have some notes on the implementation inline. :)

src/glyph/parse.rs
madig: I expect the list to always be very small.

madig: Oh, but: the Unicode code point list is ordered, and the first codepoint is considered the "primary" one in the Python libraries, so we can't just sort them 🤔

src/glyph/serialize.rs
madig: Hm... see my note above, not sure if it's just easier to do it at read/write time?

madig:
Oops, need to rebase.

src/glyph/parse.rs
cmyr: Okay yea a good reason to prefer this.


The other option I could see is using an [`IndexSet`](https://docs.rs/indexmap/1.9.1/indexmap/set/struct.IndexSet.html) to preserve insertion order but also enforce uniqueness? as use it to always store the codepoints. But maybe not worth taking on a dependency for something so small.


The thing that bothers me (irrationally, it is unlikely to matter in theory) is that with this patch we're now making an additional allocation, per-glyph, on both load and save. It might make sense to at least only do this work if there is > 1 codepoint?

madig: Hm! Yes sounds reasonable. I'll look into that.

cmyr (approved):
lgtm 🚀

madig:
Codepoints are ordered lists in the UFO spec. In defcon and ufoLib2, a glyph has the `unicode` and `unicodes` properties, where the former returns the first item of the latter.

I think the point of a custom data struct here is to avoid memory allocations on reading and writing, and itertools uses a HashSet :)

madig:
I did a list transform before, but @cmyr said he'd prefer a type 😬 I mean, I haven't settled on anything and would gladly take the simplest solution.

cmyr:
I understand that the order of the codepoints is important insofar as the first codepoint is special, but does the order of subsequent codepoints also matter?

madig:
I don't know.

anthrotype:
> does the order of subsequent codepoints also matter?

it doesn't. I probably only does if this order is carried over when roundtripping from xml to xml, e.g. if one wishes to reduce diff noise

by the way, as you all well know, the whole notion of "primary unicode codepoint" is flawed because it doesn't exist in opentype cmap tables, which is a map from codepoints to glyphs and not the other way around, as font editors make it seem.

madig:
We could go full-on set then, and sort on serialization, if we didn't care about git noise with other libraries.

cmyr:
so my original thinking with the `Codepoints` struct is that we would separate out the 'primary' codepoint, like:

```rust
Codepoints {
    primary: char,
   others: Vec<char>,
}
```

but it might be nice if internally we *do* use a single `Vec`, like in this patch, because then we can have `AsRef<[char]>` and get all the iterators etc for free.

In this case, I *do* think that we should preserve the position of the first item in the array, but otherwise ensure that it is always deduplicated. To do this at construction time, we could have something like:

```rust
use std::cmp::Ordering;

struct Codepoints(Vec<char>);

impl Codepoints {
    fn new(mut raw: Vec<char>) -> Self {
        if raw.len() <= 1 {
            return Self(raw);
        }
        let first = *raw.first().unwrap();
        
        // custom sorting that always ranks first item lowest
        raw.sort_unstable_by(|a, b| match (a, b) {
            (a, b) if *a == first && *b == first => Ordering::Equal,
            (a, _) if *a == first => Ordering::Less,
            (_, b) if *b == first => Ordering::Greater,
            (a, b) => a.cmp(b),
        });
        raw.dedup();
        Self(raw)
    }

    fn insert(&mut self, codepoint: char) {
        if !self.0.contains(&codepoint) {
            self.0.push(codepoint);
            // don't include first codepoint in sort
            self.0[1..].sort_unstable();
        }
    }
}
```

And then we would want AsRef and Deref (but not AsMut or DerefMut) impls, and maybe a few more methods? for instance do we want a method for changing the primary codepoint? what other tasks are common?

anthrotype:
> do we want a method for changing the primary codepoint?

No! We should instead deprecate that in existing APIs like defcon or ufoLib2 and nudge users to ever only use the unicodes list, in the plural. There's no use case for a "primary unicode".

madig:
I can get behind using a set (and sorting on serialization) and declaring that people should use `unicodes` in Py libs.

cmyr:
Okay, given all this input I vote for just using `indexset`. This will ensure that we maintain the original order, but also ensure that we do not contain duplicates.

PR #270, opened 2022-09-22
Mark more types as Eq

madig:
Clippy says so.


cmyr:
I've covered this in #274, which will also fix up the broken CI situation 😅

PR #272, opened 2022-11-10
Add a basic designspace parser

rsheeter:
Intended for use by Rust compiler so it can use norad to read designspace+ufo(s) to produce IR.

Serde xml seems slightly limited in ability to do things like flatten pointless container elements. If we want more control we might need to parse w/o using it.

### DIFF
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -79,8 +79,9 @@ pub struct Instances {
 /// [instance]: https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#instance-element
 #[derive(Clone, Debug, Default, PartialEq, Deserialize)]
 pub struct Instance {
-    pub familyname: String,
-    pub stylename: String,
+    // per @anthrotype, contrary to spec, familyname and stylename are optional
+    pub familyname: Option<String>,
+    pub stylename: Option<String>,
     pub name: String,
     pub filename: String,
     pub postscriptfontname: Option<String>,
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -44,7 +44,8 @@ pub struct Axis {
     /// The default value for this axis, in user space coordinates.
     pub default: f32,
     /// Records whether this axis needs to be hidden in interfaces.
-    pub hidden: Option<bool>,
+    #[serde(default)]
+    pub hidden: bool,
     /// The minimum value for a continuous axis, in user space coordinates.
     pub minimum: Option<f32>,
     /// The maximum value for a continuous axis, in user space coordinates.
@@ -106,7 +107,7 @@ pub struct Instances {
 /// [instance]: https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#instance-element
 #[derive(Clone, Debug, Default, PartialEq, Deserialize)]
 pub struct Instance {
-    // per @anthrotype, contrary to spec, familyname and stylename are optional
+    // per @anthrotype, contrary to spec, filename, familyname and stylename are optional
     /// The family name of the instance font. Corresponds with font.info.familyName
     pub familyname: Option<String>,
     /// The style name of the instance font. Corresponds with font.info.styleName
@@ -114,7 +115,7 @@ pub struct Instance {
     /// A unique name that can be used to identify this font if it needs to be referenced elsewhere.
     pub name: String,
     /// A path to the instance file, relative to the root path of this document. The path can be at the same level as the document or lower.
-    pub filename: String,
+    pub filename: Option<String>,
     /// Corresponds with font.info.postscriptFontName
     pub postscriptfontname: Option<String>,
     /// Corresponds with styleMapFamilyName
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -1,0 +1,144 @@
+//! Reading and writing designspace files.
+
+#![deny(rustdoc::broken_intra_doc_links)]
+
+use quick_xml::de::from_reader;
+use std::{fs::File, io::BufReader, path::Path};
+
+use crate::error::DesignSpaceLoadError;
+
+/// A [designspace]].
+///
+/// [designspace]: https://fonttools.readthedocs.io/en/latest/designspaceLib/index.html
+#[derive(Clone, Debug, Default, PartialEq, Deserialize)]
+#[serde(rename = "designspace")]
+pub struct DesignSpaceDocument {
+    pub format: f32,
+    pub axes: Axes,
+    pub sources: Sources,
+    pub instances: Instances,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Deserialize)]
+#[serde(rename = "axes")]
+pub struct Axes {
+    pub axis: Vec<Axis>,
+}
+
+/// A [axis]].
+///
+/// [axis]: https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#axis-element
+#[derive(Clone, Debug, Default, PartialEq, Deserialize)]
+#[serde(rename = "axis")]
+pub struct Axis {
+    pub name: String,
+    pub tag: String,
+    pub default: f32,
+    pub hidden: Option<bool>,
+    pub minimum: Option<f32>,
+    pub maximum: Option<f32>,
+    pub values: Option<Vec<f32>>,
+    pub map: Vec<AxisMapping>,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Deserialize)]
+#[serde(rename = "map")]
+pub struct AxisMapping {
+    pub input: f32,
+    pub output: f32,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Deserialize)]
+#[serde(rename = "sources")]
+pub struct Sources {
+    pub source: Vec<Source>,
+}
+
+/// A [source]].
+///
+/// [source]: https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#id25
+#[derive(Clone, Debug, Default, PartialEq, Deserialize)]
+#[serde(rename = "source")]
+pub struct Source {
+    pub familyname: Option<String>,
+    pub stylename: Option<String>,
+    pub name: String,
+    pub filename: String,
+    pub layer: Option<String>,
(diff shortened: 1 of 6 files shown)


madig:
(Will have a look later. Did you take inspiration from https://github.com/simoncozens/rust-font-tools/blob/main/designspace/src/lib.rs?)

src/error.rs
```diff
@@ -4,13 +4,26 @@ use std::io::Error as IoError;
...
 use plist::Error as PlistError;
-use quick_xml::Error as XmlError;
+use quick_xml::{DeError, Error as XmlError};
 use thiserror::Error;
 
 pub use crate::shared_types::ColorError;
 use crate::write::CustomSerializationError;
 use crate::Name;
 
+/// An error that occurs while attempting to read a UFO package from disk.
```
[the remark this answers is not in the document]
rsheeter: My copy paste prowess knows no bounds. Good catch.

src/designspace.rs
```diff
@@ -0,0 +1,144 @@
...
+use crate::error::DesignSpaceLoadError;
+
+/// A [designspace]].
+///
+/// [designspace]: https://fonttools.readthedocs.io/en/latest/designspaceLib/index.html
+#[derive(Clone, Debug, Default, PartialEq, Deserialize)]
+#[serde(rename = "designspace")]
+pub struct DesignSpaceDocument {
+    pub format: f32,
+    pub axes: Axes,
```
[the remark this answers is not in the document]
rsheeter: I would prefer that but was not sure - admittedly am noob so maybe I just missed it - how to tell serde to do that.

madig:
Ah, you may want to rebase this on #274 because it got a newer quick-xml.

src/designspace.rs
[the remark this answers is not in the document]
rsheeter: Per IM in Oxidize this seems to be a limitation of serde + xml

src/designspace.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub instance: Vec<Instance>,
+}
+
+/// An [instance]].
+///
+/// [instance]: https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#instance-element
+#[derive(Clone, Debug, Default, PartialEq, Deserialize)]
+pub struct Instance {
+    pub familyname: String,
+    pub stylename: String,
```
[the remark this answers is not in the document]
rsheeter: I've fixed here, I'll leave the spec to you :)

src/designspace.rs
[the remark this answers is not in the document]
madig: You could also write a custom deserializer function that does the possible thing and returns the desired thing👀

src/designspace.rs
```diff
@@ -0,0 +1,212 @@
...
+pub struct Instance {
+    // per @anthrotype, contrary to spec, familyname and stylename are optional
+    /// The family name of the instance font. Corresponds with font.info.familyName
+    pub familyname: Option<String>,
+    /// The style name of the instance font. Corresponds with font.info.styleName
+    pub stylename: Option<String>,
+    /// A unique name that can be used to identify this font if it needs to be referenced elsewhere.
+    pub name: String,
+    /// A path to the instance file, relative to the root path of this document. The path can be at the same level as the document or lower.
+    pub filename: String,
```
anthrotype, line 117: this should be optional too, only makes sense when interpolating static UFO instances, for defining VF named instances the output filename is not used and should not be required. Spec needs updating too

src/designspace.rs
```diff
@@ -0,0 +1,212 @@
...
+#[serde(rename = "axis")]
+pub struct Axis {
+    /// Name of the axis that is used in the location elements.
+    pub name: String,
+    /// 4 letters. Some axis tags are registered in the OpenType Specification.
+    pub tag: String,
+    /// The default value for this axis, in user space coordinates.
+    pub default: f32,
+    /// Records whether this axis needs to be hidden in interfaces.
+    pub hidden: Option<bool>,
```
anthrotype, line 47: instead of `Option<bool>` would it make sense to type this just `bool` and add `#[serde(default)]` annotation?

https://serde.rs/attr-default.html

that way, if the `hidden` attribute is omitted, it would default to `false` which I think is what we want

src/designspace.rs
rsheeter: Done, ty

src/designspace.rs
rsheeter: SGTM. Note that the spec seems not to specify what the default is - "hidden: optional, 0 or 1. Records whether this axis needs to be hidden in interfaces." - so it might be good to update that too?

src/designspace.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub instances: Instances,
+}
+
+/// https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#axes-element
+#[derive(Clone, Debug, Default, PartialEq, Deserialize)]
+#[serde(rename = "axes")]
+pub struct Axes {
+    /// One or more axis definitions.
+    pub axis: Vec<Axis>,
+}
```
cmyr, line 32: Yea, I think we want some kind of custom derive here, to just get a `Vec<Axis>`.

src/designspace.rs
```diff
@@ -0,0 +1,213 @@
...
+/// A [axis]].
+///
+/// [axis]: https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#axis-element
+#[derive(Clone, Debug, Default, PartialEq, Deserialize)]
+#[serde(rename = "axis")]
+pub struct Axis {
+    /// Name of the axis that is used in the location elements.
+    pub name: String,
+    /// 4 letters. Some axis tags are registered in the OpenType Specification.
+    pub tag: String,
```
cmyr, line 43: Maybe we should be using `font-types` in norad now?

src/designspace.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub default: f32,
+    /// Records whether this axis needs to be hidden in interfaces.
+    #[serde(default)]
+    pub hidden: bool,
+    /// The minimum value for a continuous axis, in user space coordinates.
+    pub minimum: Option<f32>,
+    /// The maximum value for a continuous axis, in user space coordinates.
+    pub maximum: Option<f32>,
+    /// The possible values for a discrete axis, in user space coordinates.
+    pub values: Option<Vec<f32>>,
```
cmyr, line 54: do we want to distinguish between a `None` and an empty vec?

src/designspace.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub output: f32,
+}
+
+/// https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#sources-element
+#[derive(Clone, Debug, Default, PartialEq, Deserialize)]
+#[serde(rename = "sources")]
+pub struct Sources {
+    /// One or more sources.
+    pub source: Vec<Source>,
+}
```
cmyr, line 75: ditto, would be nice to avoid the wrapper type.

src/designspace.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub xvalue: Option<f32>,
+    /// Separate value for anisotropic interpolations.
+    pub yvalue: Option<f32>,
+}
+
+impl DesignSpaceDocument {
+    /// Load a designspace.
+    pub fn load<P: AsRef<Path>>(path: P) -> Result<DesignSpaceDocument, DesignSpaceLoadError> {
+        let reader = BufReader::new(File::open(path).map_err(DesignSpaceLoadError::Io)?);
+        from_reader(reader).map_err(DesignSpaceLoadError::DeError)
```
cmyr, line 157: (an observation, not a suggestion:) there's a common convention in rust projects to rarely import free functions directly, but to call them via the containing module/crate; this has the advantage of making it clear when a function is external or part of the current module.

cmyr (approved):
Overall looks good, fairly straight-forward. I do think the API would be improved by removing the wrapper types. One possible pattern I see to handle something like this, while still leveraging derive, is to have a private `MyTypeSerdeRepr` type that is designed to work with serde, and then you deserialize that and  then convert it into your actual type? Not sure if this would make sense here but it's one idea. Otherwise there are various other ways to do custom deserialize.

src/designspace.rs
rsheeter: If we do that I'd prefer it be in it's own PR. Note that currently those types don't support serde and I'm not sure they should - unless it can be optional - because we want low dependencies for users of read-fonts. I tried to use something akin to Tag in IR and found it surprisingly unpleasent to get it to interact nicely with both toml and bincode IO because for toml in particular you want to see the string tag. That may simply have been a noob error.

rsheeter: Tracking in https://github.com/linebender/norad/issues/276

src/designspace.rs
rsheeter: Oh that's a nice convention. Updated to use qualified name.

src/designspace.rs
rsheeter: I wasn't sure how but it seems https://stackoverflow.com/questions/58571896/rust-serde-deserialize-xml-directly-to-vect offers a way

src/designspace.rs
rsheeter: Done

src/designspace.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    use super::DesignSpaceDocument;
+
+    fn dim_name_xvalue(name: &str, xvalue: f32) -> Dimension {
+        Dimension { name: name.to_string(), uservalue: None, xvalue: Some(xvalue), yvalue: None }
+    }
+
+    #[test]
+    fn read_single_wght() {
+        let ds = DesignSpaceDocument::load(Path::new("testdata/single_wght.designspace")).unwrap();
```
dfrg, line 248: Just a note that AsRef<Path> is implemented for &str so the Path::new is unnecessary here.

rsheeter: Removed the pointless Path::new, ty.

src/designspace.rs
rsheeter: I wish there was a way to write Option<NonEmptyVec>.

PR #273, opened 2022-11-10
impl Default for Os2WidthClass

RickyDaMa:
Came across needing this for some testing, and it seems that there's a logical default value (Normal)


cmyr (approved):
Makes sense to me!

You could also write this (as of a recent version of Rust) by adding `#[derive(Default)]`, and then marking the appropriate variant with the `#[default]` attribute.

madig:
Let's go with the newest and best. I don't recall caring about the MSRV.

cmyr:
> Tangent: https://matklad.github.io/2022/10/24/actions-permissions.html

cool, didn't know about that!

PR #280, opened 2022-11-24
Finer grained control over layer loading

### LINKED ISSUE #279: load a specific layer instead of all at once?
In ufoLib2 or defcon, layers (and even specific glyphs within those) can be loaded lazily upon request.

I see norad has a DataRequest that allows to parametrize font loading and either request to load all the layers or none of them.  It would be nice to have something in between.

For example, designspace sources have a `layer` attribute to specify which layer in given UFO contains the glyphs to be used (by default if omitted, it's the default layer). I would like to load a UFO with norad and only read a specific layer, be it the default one or one with a given name, instead of having to eagerly load them all and then get the one I need.

right now, in some code of mine, I am doing the job of parsing layercontents.plist myself to get to the glif paths I need, ideally norad would provide an API for this

cmyr:
With this change, the user can now request to load only specific layers when loading a font.

closes #279

### DIFF
--- a/src/data_request.rs
+++ b/src/data_request.rs
@@ -1,6 +1,6 @@
 //! Load only requested font data.
 
-use std::path::Path;
+use std::{path::Path, rc::Rc};
 
 /// A type that describes which components of a UFO should be loaded.
 ///
@@ -56,11 +56,11 @@ pub struct DataRequest {
 }
 
 /// A type describing which layers to load.
-#[derive(Clone, Debug, Default, PartialEq, Eq)]
+#[derive(Clone)]
 pub(crate) struct LayerFilter {
     all: bool,
     load_default: bool,
-    named_layers: Vec<String>,
+    custom: Option<Rc<dyn Fn(&str, &Path) -> bool + 'static>>,
 }
 
 impl LayerFilter {
@@ -71,7 +71,7 @@ impl LayerFilter {
     pub(crate) fn should_load(&self, name: &str, path: &Path) -> bool {
         self.all
             || (self.load_default && path == Path::new("glyphs"))
-            || self.named_layers.iter().any(|req_name| name == req_name)
+            || self.custom.as_ref().map(|f| f(name, path)).unwrap_or(false)
     }
 
     /// `true` if this filter includes the default layer
@@ -104,6 +104,11 @@ impl DataRequest {
     }
 
     /// Request that returned UFO data include layers and their glyph data.
+    ///
+    /// See also the [`filter_layers`] and [`default_layer`] options.
+    ///
+    /// [`filter_layers`]: Self::filter_layers
+    /// [`default_layer`]: Self::default_layer
     pub fn layers(mut self, b: bool) -> Self {
         self.layers.all = b;
         self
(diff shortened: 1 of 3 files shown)


src/data_request.rs
```diff
@@ -53,9 +55,43 @@ pub struct DataRequest {
...
+impl LayerFilter {
+    fn from_bool(b: bool) -> Self {
+        LayerFilter { all: b, ..Default::default() }
+    }
+
+    pub(crate) fn should_load(&self, name: &str, path: &Path) -> bool {
+        dbg!(name, path);
+        self.all
+            || (self.load_default && path == Path::new("glyphs"))
+            || self.named_layers.iter().any(|req_name| name == req_name)
```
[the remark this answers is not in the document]
anthrotype: I believe it was deliberate because

```rust
let v = [String::from("hello"), String::from("world")]; // slice of `String`
assert!(v.contains("hello"));  // would fail because "expected struct `String`, found `str`"
```

The Vec::contains docs mention this too
https://doc.rust-lang.org/std/vec/struct.Vec.html#method.contains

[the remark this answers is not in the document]
anthrotype: there's a open issue on that in https://github.com/rust-lang/rust/issues/42671, pity that Vec::contains doesn't just Do The Right Thing™

src/font.rs
```diff
@@ -200,16 +201,16 @@ impl Font {
     ///
     /// let datareq = DataRequest::none().lib(true);
     ///
-    /// let ufo = Font::load_requested_data("path/to/font.ufo", &datareq).expect("failed to load");
+    /// let ufo = Font::load_requested_data("path/to/font.ufo", datareq).expect("failed to load");
     /// ```
     pub fn load_requested_data(
         path: impl AsRef<Path>,
-        request: &DataRequest,
+        request: DataRequest,
```
anthrotype, line 208: wouldn't this DataRequest -> &DataRequest be a breaking change?

anthrotype: nevermind, norad hasn't reached 1.0 yet so it should be ok

anthrotype (approved):
LGTM!

src/data_request.rs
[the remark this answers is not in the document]
cmyr: yea, I tried this with contains but ran into the issue described. :)

src/font.rs
cmyr: it is a breaking change, yes, and it will be made in the bump from 0.7 to 0.8.

cmyr: Another breaking change is the removal of the public `layers` field in `DataRequest`.

cmyr:
Update: I've redone the API on this to now be closure based, like Iterator::filter, instead of taking specific strings. I think this is strictly more flexible.

madig:
> There is one change worth noting, here: we no longer store DataRequest in the font object, and instead discard it after use. As far as I can tell this doesn't change anything, but if there's something I'm overlooking please let me know!

I suppose one day we might want to implement lazy loading or some such and having the DataRequest around means we know what we haven't loaded yet, but we can wait until we cross that bridge.

ctrlcctrlv:
@cmyr Doesn't break MFEKmetadata. Merged in ff8a29a1d0cc60b1cbd49e52627dd249654ae3f4. At least one of your changes _did_ break it though, and should be marked semver-breaking, removing `Glyph::name`: https://github.com/MFEK/metadata/commit/ff8a29a1d0cc60b1cbd49e52627dd249654ae3f4#diff-a41dce0b085407380a67e5cd12c9dcdca21da77924dc7d524dd7b9650cada3f3. But that was easy enough to fix :-)

cmyr:
> @cmyr Doesn't break MFEKmetadata. Merged in [MFEK/metadata@ff8a29a](https://github.com/MFEK/metadata/commit/ff8a29a1d0cc60b1cbd49e52627dd249654ae3f4). At least one of your changes _did_ break it though, and should be marked semver-breaking, removing `Glyph::name`: [MFEK/metadata@ff8a29a#diff-a41dce0b085407380a67e5cd12c9dcdca21da77924dc7d524dd7b9650cada3f3](https://github.com/MFEK/metadata/commit/ff8a29a1d0cc60b1cbd49e52627dd249654ae3f4#diff-a41dce0b085407380a67e5cd12c9dcdca21da77924dc7d524dd7b9650cada3f3). But that was easy enough to fix :-)

Okay great, will merge this. The next release will be 0.8, and will include these breaking changes (including non-pub Glyph name field)

PR #281, opened 2022-11-26
Add Layer::names method

cmyr:
cc @anthrotype


cmyr:
thinking about this more I think it's so easy to get the names already that I'm inclined to leave this out.

PR #282, opened 2022-11-30
Make name-to-file-name conversion functions public

cmyr:
cc @anthrotype


anthrotype (approved):
LGTM thanks!

anthrotype (changes requested):
what i'd like is to make `user_name_to_file_name` public directly, so we can use with other suffixes than .glif

anthrotype (approved):
LGTM for real now!

PR #283, opened 2022-11-30
Use IndexSet instead of Vec for codepoints

cmyr:
This ensures that codepoints are unique, but also respects the original order in which they appear in the file.

This is an alternative to #268. Although it brings in a dependency, it leads to a much simpler and more efficient implementation overall.

Thoughts @madig ?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ quick-xml = { version = "0.26.0", features = ["serialize"] }
 rayon = { version = "1.3.0", optional = true }
 kurbo = { version = "0.9.0", optional = true }
 thiserror = "1.0"
+indexmap = {version = "1.9.2", features = ["serde"] }
 
 [dependencies.druid]
 default-features = false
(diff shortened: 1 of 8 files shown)


madig:
The readme says:

> Preserves insertion order as long as you don't call .remove().

Hm :) Are we gonna let that stop our OCD?

madig:
Otherwise, I'm ok with pulling it in and exposing it directly in the API. I don't like making delegation functions...

cmyr:
So I'm okay going either way with this. Fwiw it would be easy enough to modify the contents without needing to have the crate in scope, e.g.

```rust
my_glyph.codepoints.clear();
my_glyph.codepoints.extend(['a', 'b', 'c']);
```

but I think maybe not having it be public API *is* a strictly better design, so I'll go ahead and do that.

cmyr:
Okay, I've updated this with a new custom Codepoints type, wrapping an indexset.

cmyr:
@RickyDaMa unfortunately I can't use `impl Iterator` with the `IntoIterator` trait. :(

PR #285, opened 2022-12-08
LayerSet: add retain & remove_empty_layers

RickyDaMa:
This is functionality I have a use for and can't implement anywhere except in the library, given `layers` is a private member of `LayerSet`

I don't like that with `retain` I check every layer to see if its name matches the default, but there's no good way (without allocating) to 'skip one and then retain' that I'm aware of

Simple positive test case added for each method

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -896,4 +896,54 @@ mod tests {
         let bglayer = layerset.get("background").unwrap();
         assert_eq!(bglayer.len(), 1);
     }
+
+    #[test]
+    fn test_retain() {
+        let mut layers = LayerSet {
+            layers: vec![
+                Layer::default(),
+                Layer::new(Name::new("fizz").unwrap(), PathBuf::new()),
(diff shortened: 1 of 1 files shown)


src/layer.rs
```diff
@@ -213,6 +213,25 @@ impl LayerSet {
...
+    pub fn retain<F>(&mut self, mut predicate: F)
+    where
+        F: FnMut(&Layer) -> bool,
+    {
+        self.layers.retain(|layer| layer.name == DEFAULT_LAYER_NAME || predicate(layer))
+    }
+
+    /// Removes any layers that contain no glyphs
+    pub fn remove_empty_layers(&mut self) {
+        self.retain(|l| !l.is_empty());
```
cmyr, line 233: we should probably also guard against the default layer being removed here?

cmyr (commented):
looks reasonable, one note inline. :)

cmyr (approved):
🚀

cmyr:
thanks! if you want a new release, feel free to make a PR bumping the version to 0.8.1.

PR #287, opened 2023-02-23
Release v0.8.1 (bumped dependencies, fed clippy, fixed CI builds)

RickyDaMa:
Making this PR as suggested in #285 so I can have access to the functionality I need :)

Bumped kurbo version and all tests pass, so seems like that upgrade could be made without issue. Held back `quick-xml` as it both causes test failures, and `plist` depends on 0.26 (meaning you'd get duplicate dependencies pulled in, which smells)

Tagging this and releasing it to crates.io would be amazing when you have time ❤️

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -30,6 +30,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: install druid system dependencies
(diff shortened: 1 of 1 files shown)


cmyr:
oops sorry I was just on vacation for a couple weeks, will get this done shortly!

PR #289, opened 2023-03-06
Update custom whitespace API

cmyr:
The old API was designed to work around the fact that the plist and quick-xml crates had different APIs for specifying custom indent behaviour. The plist crate has since been updated to match the API of quick-xml, which means that we can do the same.


cmyr:
@RickyDaMa feel free to give this a review if you like! I'm going to hold the release a day because I think we can get a release of the plist crate done, and then we can update quick-xml.

cmyr:
My desire was to not require an additional namespace or import for these, and so they're provided here as a convenience. In general I think the pattern of defining constants on the type where they're used is at least modestly idiomatic? I could get fancy and define a custom enum or something but I think that's overthinking a fairly niche feature. :)

PR #290, opened 2023-03-06
Update deps, bump version

cmyr:
Just the 'updating deps' bit of #287. I assume from that PR that this breaks CI somehow.


cmyr:
I think the issue is that druid, which is an optional dependency, now requires more system dependencies.

I've updated this to include the new quick-xml, which ended up being a pain. The plist crate has been updated, though, so at least now we're all on 0.27.

dfrg (approved):
Very nice to drop the excess types!

PR #291, opened 2023-03-09
Add `LayerSet::get_or_create_layer`

RickyDaMa:
This one's a pain to do intuitively because of the borrow checker

You want to do something like:

```rust
match ufo.layers.get_mut("foobar") {
    Some(layer) => layer,
    None => ufo.layers.create_layer("foobar")?,
}
```

But the borrow checker won't let you because of multiple mutable borrows (which is arguably incorrect since `None` shouldn't be considered tied to `ufo.layers`, but improving the borrow checker is beyond my skill)

So what I've implemented is accepted by the borrow checker and doesn't iterate through `layers` multiple times. I've pulled out the happy path of `new_layer` to avoid the code duplication, hence the creation of the private function `create_layer`. Bikeshedding the naming is welcome because it's definitely not amazing, but I just wanted to get the PR out there and worry about naming later

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -136,6 +136,14 @@ impl LayerSet {
         self.layers.iter().map(|l| &l.name)
     }
 
+    fn create_layer(&mut self, name: Name) -> &mut Layer {
+        let path = util::default_file_name_for_layer_name(&name, &self.path_set);
+        let layer = Layer::new(name, path);
+        self.path_set.insert(layer.path.to_string_lossy().to_lowercase());
+        self.layers.push(layer);
+        self.layers.last_mut().unwrap()
+    }
+
     /// Returns a new layer with the given name.
     pub fn new_layer(&mut self, name: &str) -> Result<&mut Layer, NamingError> {
         if name == DEFAULT_LAYER_NAME {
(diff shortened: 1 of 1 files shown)


madig:
I think I removed this method in https://github.com/linebender/norad/pull/255, also see the reasoning. If you come up with something robust, I'd want it back in :)

madig:
If you can create something that deals with the pain points we found earlier, certainly! We can look at it together.

src/layer.rs
```diff
@@ -144,11 +152,22 @@ impl LayerSet {
...
+        }
+    }
+
+    /// Returns a mutable reference to a layer, by name, creating the layer if it doesn't exist
+    pub fn get_or_create_layer(&mut self, name: &str) -> Result<&mut Layer, NamingError> {
+        let index = self.layers.iter().position(|l| l.name.as_str() == name);
+        match index {
+            Some(its_here) => Ok(&mut self.layers[its_here]),
+            None => {
+                // We can avoid the formal checks that new_layer does because by checking if it's already in the list,
```
cmyr, line 165: I think we still want to check that we aren't `DEFAULT_LAYER_NAME`, since it's possible that the default layer has another name.

See this test case that was deleted when we removed the previous version:

```rust
#[test]
     #[should_panic(expected = "reserved")]
     fn get_or_create_false_default_layer() {
         let mut layer_set = LayerSet::default();
         layer_set.rename_layer("public.default", "foreground", false).unwrap();
         layer_set.get_or_create("public.default");
     }
```


and... I think honestly it's easiest/hardest to screw up if t his *does* just call the old `new_layer` method if we find nothing. In practice this isn't called much, and iterating even small-10s of layers is still going to be very fast, and then we can be confident we're always using the same logic.

cmyr (approved):
So one of the reasons we got rid of this was that we weren't using it, and I don't like holding onto API that isn't being used. If you *are* using it, though, then that is no longer a concern.

In any case, the new method here feels much less of a wart than the previous one, I think because we didn't have `NamingError` originally?

I have one note inline, but I'm happy to merge this, and then I'll get a release out when that's done. :)

cmyr:
Thanks!

PR #299, opened 2023-06-09
Add nice to have methods

### LINKED ISSUE #298: Nice to have methods
Listing as I find them.

* LayerSet::iter_mut
* Layer::retain
* Layer::get_or_create_glyph

RickyDaMa:
Will close #298, merry Christmas @madig! 😛 

Since these were all just delegating to inner types that already have the functions, hopefully this shouldn't prove too controversial. I only spent 10 minutes on it lol.

I chose to expose `Layer::entry` instead of creating a `Layer::get_or_create_glyph` as I think it's a more versatile interface.

Documentation is copied from the standard library.

Supporting the `druid` feature makes this slightly less pleasant. I can either copy paste the `entry` method (plus documentation) to change the return type, or we can use the type alias I made (bikeshedding welcome): `MaybeArcGlyph`. I don't know of a way to `#[cfg]` change just the return type

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashSet};
+use std::collections::{btree_map, BTreeMap, HashSet};
 use std::fs;
 use std::path::{Path, PathBuf};
 
@@ -474,6 +474,11 @@ impl Layer {
     /// about it, if you like.
     pub fn path(&self) -> &Path {
         &self.path
+    }
+
+    /// Gets the given key's corresponding entry in the map for in-place manipulation.
+    pub fn entry(&mut self, glyph: Name) -> btree_map::Entry<Name, Glyph> {
+        self.glyphs.entry(glyph)
     }
 
     /// Returns a reference to the glyph with the given name, if it exists.
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -611,6 +611,14 @@ impl Layer {
         self.glyphs.values_mut()
     }
 
+    /// Retains only the elements specified by the predicate.
+    ///
+    /// In other words, remove all pairs `(k, v)` for which `f(&k, &mut v)` returns `false`.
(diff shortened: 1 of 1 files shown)


madig:
Maybe I'm missing it!

src/layer.rs
```diff
@@ -250,6 +255,11 @@ impl Default for LayerSet {
     }
 }
 
+#[cfg(not(feature = "druid"))]
+type MaybeArcGlyph = Glyph;
+#[cfg(feature = "druid")]
+type MaybeArcGlyph = Arc<Glyph>;
+
```
cmyr, line 262: instead of getting fancy here, let's just put the `entry` method behind a `#[cfg(not(feature = "druid"))]`? the druid stuff is pretty legacy at this point, and if in the future I need this feature I can deal with it then. :)

src/layer.rs
```diff
@@ -606,6 +621,14 @@ impl Layer {
         self.glyphs.values_mut()
     }
 
+    /// Retains only the elements specified by the predicate.
+    ///
+    /// In other words, remove all pairs `(k, v)` for which `f(&k, &mut v)` returns `false`.
+    /// The elements are visited in unsorted (and unspecified) order.
+    pub fn retain(&mut self, f: impl FnMut(&Name, &mut MaybeArcGlyph) -> bool) {
```
cmyr, line 628: ditto, we can say this is only turned on if the druid feature is disabled 🤷

cmyr (approved):
Oops sorry! this looks good, although I think we should avoid making things more complicated in order to support druid; notes inline. :)

cmyr (approved):
Thanks, this looks good! no worry about tests, this isn't adding any logic or anything. :)

src/layer.rs
madig: Should we do a pass to remove druid code then?

PR #301, opened 2023-06-14
Allow design space to contain no instances

### LINKED ISSUE #300: Generated designspace from fontmake doesn't load in norad
1. Clone git@github.com:googlefonts/fontc.git
1. Install fontmake in a venv
1. `fontmake -o=ufo resources/testdata/glyphs3/WghtVar.glyphs`
1. `cargo run -p fontc -- --source master_ufo/WghtVar.designspace`
   * Appears to be from https://github.com/googlefonts/fontc/blob/f805f0d509f0d2e636095487044cbdc1c8500631/ufo2fontir/src/source.rs#L148-L151
   * Fixed by manually added instance elements but would prefer not to

cmyr:
I could make the 'instances' field optional, but I've opted to just use an empty Vec if the instances are missing, since this is a non-breaking change.

- fixes #300

cc @madig, does this seem reasonable to you?

### DIFF
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -22,7 +22,7 @@ pub struct DesignSpaceDocument {
     #[serde(deserialize_with = "serde_impls::deserialize_sources")]
     pub sources: Vec<Source>,
     /// One or more instances.
-    #[serde(deserialize_with = "serde_impls::deserialize_instances")]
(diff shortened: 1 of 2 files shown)


rsheeter (approved):
LGTM, ty

madig (approved):

PR #304, opened 2023-06-29
Required is more like a suggestion

cmyr:
This updates DesignSpaceDocument to allow inputs where sources do not have names; we will autogenerate names if they are missing.

### DIFF
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -83,7 +83,7 @@ pub struct Source {
     #[serde(rename = "@stylename")]
     pub stylename: Option<String>,
     /// A unique name that can be used to identify this font if it needs to be referenced elsewhere.
-    #[serde(rename = "@name")]
+    #[serde(rename = "@name", default = "serde_impls::generate_missing_source_name")]
     pub name: String,
     /// A path to the source file, relative to the root path of this document.
     ///
@@ -162,6 +162,8 @@ impl DesignSpaceDocument {
 }
 
 mod serde_impls {
+    use std::sync::atomic::AtomicU64;
+
     use super::{Axis, Dimension, Instance, Source};
     use serde::{Deserialize, Deserializer};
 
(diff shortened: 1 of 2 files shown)


rsheeter:
Spec is wrong, fixed in https://github.com/fonttools/fonttools/pull/3194

src/designspace.rs
```diff
@@ -208,6 +210,17 @@ mod serde_impls {
...
+
+    /// Generate a unique name for a source.
+    ///
+    /// We do not make guarantees about what name will be assigned to a given
+    /// source, only that it will be unique within this process. If the designer
+    /// cares, they should provide a name explicitly.
+    pub fn generate_missing_source_name() -> String {
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+        let next_n = COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+        format!("unnamed_source_{next_n}")
```
rsheeter, line 222: Optional, we could match the existing authoritative implementation (https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/designspaceLib/__init__.py#L2311), though I cannot think of a reason it would matter.

src/designspace.rs
```diff
@@ -265,4 +278,12 @@ mod tests {
     fn load_with_no_instances() {
         DesignSpaceDocument::load("testdata/no_instances.designspace").unwrap();
     }
+
+    #[test]
+    fn required_was_like_just_a_suggestion_okay() {
```
rsheeter, line 283: load_with_no_source_name?

testdata/no_source_names.designspace
```diff
@@ -0,0 +1,35 @@
...
+            <location>
+        <dimension name="Weight" xvalue="400"/>
+      </location>
+    </source>
+    <source filename="TestFamily-Bold.ufo" familyname="Test Family" stylename="Bold">
+      <location>
+        <dimension name="Weight" xvalue="700"/>
+      </location>
+    </source>
+  </sources>
```
rsheeter, line 17: Optional: make this the most minimal source thatis value, e.g. no familyname or any other optional attribute

src/designspace.rs
cmyr: this would be more annoying, since we don't have access to serde's internal state (and hence don't know the actual index of the current source) and so we would need to do a post-processing step to assign the names after loading the whole designspace. 🤷

PR #308, opened 2023-07-20
Bump patch version

cmyr:


madig:
Is there something holding this up? I'd like to use it :)

cmyr:
It looks like we're going to change some of the details of the 'add placeholder names' code, so maybe we should skip this release, or at least hold on to it until we actually make those changes.

PR #309, opened 2023-07-21
Saving designspaces

RickyDaMa:
Adds `DesignspaceDocument::save`, as well as a lot of `serde` magic to make it happen

Main things I'm looking for feedback on is if I'm using `quick_xml` in a sane way (having never used it before, and it didn't come across at all intuitively), and if there are any `Option` or `Vec` fields I don't want to skip, even if they're `None`/empty

~~Also still working on getting the loading to work, hence the new test is failing. Oh and I need to fix conflicts with #307's changes~~ Edit: all sorted

If this could be released once merged that'd be mighty useful for me (maybe as part of #308 👀)


madig:
Lib is actually important. You can add a new field, it should be the same type as a Font.lib.

Hoolean:
Hello!

Is there a way for us to dodge writing automatically-generated source and instance names? This would avoid serialisation being non-deterministic based on when in the process the designspace was loaded (e.g. currently loading the same designspace twice in succession and serialising has different output).

I suppose this might not be possible without giving the `unnamed_source_` prefix a special semantic meaning, or tweaking the approach of #304 and #307, although the latter would have benefits in other places too: e.g. fixing PartialEq for the same designspace loaded at different times in the process :)

@cmyr for #304, did we home the default name implementation upstream in norad instead of in the compiler to avoid the breaking API change from `String` to `Option<String>`? 🔎

cmyr:
> Hello!
> 
> Is there a way for us to dodge writing automatically-generated source and instance names? This would avoid serialisation being non-deterministic based on when in the process the designspace was loaded (e.g. currently loading the same designspace twice in succession and serialising has different output).

What does the python impl do in this case?

> I suppose this might not be possible without giving the `unnamed_source_` prefix a special semantic meaning, or tweaking the approach of #304 and #307, although the latter would have benefits in other places too: e.g. fixing PartialEq for the same designspace loaded at different times in the process :)

I'm wondering if we even need to add these temp names at all? Maybe we can just let these things be unnamed? the issue in the compiler, as I understand it, is that we want *something* we can use to uniquely identify sources/instances, but as mentioned below this doesn't need to be norad's job.

> @cmyr for #304, did we home the default name implementation upstream in norad instead of in the compiler to avoid the breaking API change from `String` to `Option<String>`? 🔎

tbh I don't think I really thought about it too much? I'm not super worried about breaking API changes, where they are necessary, and I *do* think norad should be as dumb as possible, so maybe we want to back out #304 & #307?

Hoolean:
> What does the python impl do in this case?

From what I can see designspaceLib adds the `temp_master` prefixed name [on load](https://github.com/fonttools/fonttools/blob/04b31405/Lib/fontTools/designspaceLib/__init__.py#L2311-L2313), and then uses that same prefix to see which names should be dropped [on save](https://github.com/fonttools/fonttools/blob/04b31405/Lib/fontTools/designspaceLib/__init__.py#L1786-L1788).

> I'm wondering if we even need to add these temp names at all? Maybe we can just let these things be unnamed? the issue in the compiler, as I understand it, is that we want _something_ we can use to uniquely identify sources/instances, but as mentioned below this doesn't need to be norad's job.
> ...
> I'm not super worried about breaking API changes, where they are necessary, and I _do_ think norad should be as dumb as possible, so maybe we want to back out #304 & #307?

It definitely seems like the root cause is designspaceLib promising different things in the designspace schema (optional source name) and the API it exposes at runtime (guaranteed source name), although even then the latter is inconsistent as if you instantiate and add sources at runtime then they are not given automatically-generated names.

Overall it seems like this might be a coupling that arose between fontTools and the Python compiler stack, and so from a spiritual mindset of wanting us to carry as little 'magic' as possible from the Python world into the Rust world then I also wonder if we can drop it :grin:

madig:
I'm in favor of making names an Option<String> and changing the spec to reflect real politics.

cmyr:
Cool, I am always in favour of less code / less logic / being as dumb as possible. Someone want to PR these changes?

madig:
I'm ok with insta et al. You could do some light fuzzing by roundtripping all Designspace files on your puter.

cmyr:
I think snapshot testing infrastructure might be overkill? I would be happy with a single test case that covers all field types, and includes multi-item and nested arrays/dictionaries. If we can load this from file, and then write it to string and have them match, that's great. Once we have it working our serialization code should never really change, and it shouldn't be influenced by anything outside of our control (the cases where snapshots are particularly helpful)

src/designspace.rs
```diff
@@ -168,55 +170,130 @@ impl DesignSpaceDocument {
...
         quick_xml::de::from_reader(reader).map_err(DesignSpaceLoadError::DeError)
     }
+
+    /// Save a designspace.
+    pub fn save(&self, path: impl AsRef<Path>) -> Result<(), DesignSpaceSaveError> {
+        let mut buf = String::from("<?xml version='1.0' encoding='UTF-8'?>\n");
+        let mut xml_writer = quick_xml::se::Serializer::new(&mut buf);
+        xml_writer.indent(' ', 2);
+        self.serialize(xml_writer)?;
+        buf.push('\n'); // trailing newline
```
cmyr, line 180: do we always write unix newlines, or should we write `\r\n` on windows?

src/designspace.rs
```diff
@@ -168,55 +170,130 @@ impl DesignSpaceDocument {
...
+        pub(crate) fn serialize<S>(instances: &[Instance], serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: Serializer,
+        {
+            #[derive(Serialize)]
+            struct Helper<'a> {
+                instance: &'a [Instance],
+            }
+            let helper = Helper { instance: instances };
+            helper.serialize(serializer)
```
cmyr, line 241: you know i bet we could have a simple `macro_rules` macro to handle all the duplicate code, but let's leave that as follow-up.

src/error.rs
```diff
@@ -25,6 +25,18 @@ pub enum DesignSpaceLoadError {
     DeError(#[from] DeError),
 }
 
+/// An error that occurs while attempting to write a designspace file to disk.
+#[derive(Debug, Error)]
+pub enum DesignSpaceSaveError {
+    /// An [`std::io::Error`].
+    #[error("failed to open/write file")]
```
cmyr, line 32: It's useful to also print the inner error here, which may tell us a bit more
```suggestion
    #[error("failed to load designspace: '{0}'")]
```

src/error.rs
```diff
@@ -25,6 +25,18 @@ pub enum DesignSpaceLoadError {
...
+    Io(#[from] IoError),
+
+    /// A serialization error.
+    #[error("failed to deserialize")]
```
cmyr, line 36:
```suggestion
    #[error("failed to deserialize designspace: '{0}'")]
```

src/serde_xml_plist.rs
```diff
@@ -1,467 +1,657 @@
...
-    Real,
-    String,
-    Data,
-    Date,
-    True,
-    False,
-}
+    // the logic for deserializing a dict is a subset of the general deser logic,
+    // so we reuse this type for both cases.
+    pub(super) enum ValueVisitor {
```
cmyr, line 60: if this is shared between ser&de, I would just move it into the root module?

src/serde_xml_plist.rs
```diff
@@ -1,467 +1,657 @@
...
 }
 
-struct DictWrapper(Dictionary);
-struct ValueWrapper(Value);
-struct ArrayWrapper(Vec<Value>);
-struct IntWrapper(plist::Integer);
-
-/// The literal keyword 'key'.
-struct KeyKeywordLiteral;
+mod de {
```
cmyr, line 23: just an observation: I agree it makes sense to go with the ser/de mods, but moving all the existing code into a new module in this PR makes it very difficult to tell where the actual changes are, versus just changes in indent.

No need to change this now, but it's something to consider when making this kind of change, in the future. I would have probably left the existing code where it was, and then added a shim `de` module that just used the types that were declared in the module root; after the PR was merged it's an easy refactor to move things around. :)

src/serde_xml_plist.rs
```diff
@@ -1,467 +1,657 @@
...
+        {
+            match &self.0 {
+                Value::Array(a) => ArrayInnerHelper(a).serialize(serializer),
+                Value::Dictionary(d) => DictionaryInnerHelper(d).serialize(serializer),
+                Value::Boolean(_) => {
+                    unreachable!(
+                        "ValueInnerHelper should never serialize a boolean as it has no inner"
+                    )
+                }
+                Value::Data(d) => serializer.serialize_str(&base64_standard.encode(d)),
```
cmyr, line 429: when we get plist v1.6 on crates.io we'll want to replace this with `Data::to_xml_format`.

src/serde_xml_plist.rs
```diff
@@ -1,467 +1,657 @@
...
-        assert_eq!(
-            readme.lib.get("bestBefore").and_then(Value::as_date).map(|d| d.to_xml_format()),
-            Some("2345-01-24T23:22:21Z".into())
-        );
-        assert_eq!(
-            readme.lib.get("payload").and_then(Value::as_data),
-            Some("u got 0wned".as_bytes())
-        );
-    }
+    mod de {
```
cmyr, line 487: Is there any reason to have separate submodules in the tests? I'd just put all the tests in `test` unless there's some specific rationale.

testdata/a_bit_of_everything.plist
```diff
@@ -0,0 +1,61 @@
+<TestMe>
```
cmyr, line 1: to be a valid plist this should have the XML preamble etc; but of course, we aren't really deserializing an actual plist, or we wouldn't have this problem.

I avoided having to think about this by just writing the strings for my test data inline in the functions, so maybe you could do the same? Then we could also reuse the same test data for both ser&de.

cmyr (approved):
Overall this looks reasonable to me... ultimately the proof is in the roundtrip. :)

src/serde_xml_plist.rs
cmyr: Ah I assumed that the `pub(super)` was there so that this could be shared, but I think I just misunderstood, no worries. I'm not sure unifying is possible, and I'm not too worried about it as long as we have something that works.

src/serde_xml_plist.rs
cmyr: oh cool, TIL. Will look into that setting...

src/designspace.rs
```diff
@@ -311,4 +388,30 @@ mod tests {
...
+        assert!(!serialized.contains("<lib>"));
+        assert!(!serialized.contains("<lib/>"));
+    }
+
+    #[test]
+    fn load_save_round_trip() {
+        // Given
+        let dir = tempdir::TempDir::new("norad_designspace_load_save_round_trip").unwrap();
+        let mut ds_test_save_location = dir.path().to_path_buf();
+        ds_test_save_location.push("wght.designspace");
```
cmyr, line 406: silly, but you can just write this as,

```rust
let ds_test_save_location = dir.path().join("wght.designspace");
```

testdata/a_bit_of_everything.plist
cmyr: Ah okay I hadn't caught that these files are auto-generated. In that case I would just use the `.xml` suffix, and then as follow-up maybe we can reuse these files for the parsing test.

PR #311, opened 2023-07-25
Deserialize designspace lib fields

cmyr:
This was about as painful as expected, and requires us to manually glue together the xml data model with the plist model, using serde.

To briefly summarize the problem: XML data, by definition, does not have any defined semantics. Semantics are added on top of XML on a per-case basis, by indicating the document type inline
(<https://en.wikipedia.org/wiki/Document_type_declaration>).

The plist crate does implement derive for plist types, but they are based on the assumption that type information is being provided by some deserializer (e.g. for a specific file format). This is the right thing for the plist crate to do. Unfortunately quick_xml cannot provide this type information, since quick_xml does not know that we are deserializing a plist.

To make this all work, I have added a bunch of manual deserialize code. This code will only work when deserializing from XML; if in the future we wish to serialize to other formats, we will need to revise it.

Some notes:

- This only implements deserialization, not serialization. (This is fine, since we do not currently serialize anyway, but we'll have to figure that out eventually)
- this does not support the UID type, which is not accepted in XML plists

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ rayon = { version = "1.3.0", optional = true }
 kurbo = { version = "0.9.0", optional = true }
 thiserror = "1.0"
 indexmap = {version = "2.0.0", features = ["serde"] }
+base64 = "0.21.2"
 
 [dependencies.druid]
 default-features = false
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -4,7 +4,10 @@
 
 use std::{fs::File, io::BufReader, path::Path};
 
+use plist::Dictionary;
+
 use crate::error::DesignSpaceLoadError;
+use crate::serde_xml_plist as serde_plist;
 
 /// A [designspace].
 ///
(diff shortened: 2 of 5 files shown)


cmyr:
yea, serialization will likely be annoying too, but I'm not sure exactly *how* annoying. I guess we'll find out? 😬

cmyr:
Huh interesting, I honestly kind of appreciate the compactness of your version, maybe we should take that instead? I'm not even sure there's much performance difference to be honest. Want to PR it, maybe including the tests from this branch?

cmyr:
Although thinking about it a bit more, you are going to need more logic to handle date/data/integer, at least.

cmyr:
Thinking about this more I suspect that your approach just may not be tractable, because XML alone doesn't provide enough structural information to identify the appropriate types? I'm not positive, but if we're happy enough with my version then I'm okay with merging this as-is.

cmyr:
Okay I've talked myself into thinking this is the right approach, will merge and then you can use this to figure out how to save? Happy to help if you get stuck.

PR #313, opened 2023-07-25
[designspace] Stop generating source/instance names

### LINKED ISSUE #312: Stop generating placeholder names to sources/instances
After discussion in  https://github.com/linebender/norad/issues/309, we have come to the agreement that this should not be the responsibility of norad, and should be left up to the consumer.

cmyr:
closes #312

### DIFF
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -83,8 +83,8 @@ pub struct Source {
     #[serde(rename = "@stylename")]
     pub stylename: Option<String>,
     /// A unique name that can be used to identify this font if it needs to be referenced elsewhere.
-    #[serde(rename = "@name", default = "serde_impls::generate_missing_source_name")]
-    pub name: String,
+    #[serde(rename = "@name")]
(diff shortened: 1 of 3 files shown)


Hoolean:
Thanks for putting this PR together Colin, you beat me to it! Nice to see that only minimal changes are needed in fontc too :+1:

anthrotype:
> Nice to see that only minimal changes are needed in fontc too

fontc relies on those name attributes being set to non Optional values (e.g. it uses the source.name as a unique way to identify specific source elements, for the filename may well be the same for multiple sources referencing different layers from the same .ufo), so this is a breaking change.. If norad is unwilling to provide these unique name attributes fontc will have to do the same dance with AtomicU64. It's not sufficient to fallback to a constant "placeholder_source_name" because that might well clash.

cmyr:
I have a patch for fontc written, and I do think that this belongs there, instead of norad.

PR #314, opened 2023-08-02
Macro generated `serde` helpers for designspace fields

RickyDaMa:
From [here](https://github.com/linebender/norad/pull/309#discussion_r1280741822)

Two commits show two 'flavours' of the macro: either it can do absolutely everything (generating the module and two functions within it), or we could go for something more digestible and have individual macros for generating the ser/de impls and then write the modules ourselves

Either way it's a huge lines-of-code saving (until you factor in the giant doc comment lol)

### DIFF
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -184,97 +184,95 @@ impl DesignSpaceDocument {
 }
 
 mod serde_impls {
-
-    use super::{Axis, Dimension, Instance, Source};
-    use serde::{Serialize, Serializer};
-
-    macro_rules! de_from_field {
-        ($name:ident, $container:path) => {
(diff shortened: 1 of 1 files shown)


src/designspace.rs
```diff
@@ -184,117 +184,95 @@ impl DesignSpaceDocument {
...
+    ///     {
+    ///         #[derive(Serialize)]
+    ///         struct Helper<'a> {
+    ///             dimension: &'a [Dimension],
+    ///         }
+    ///         let helper = Helper { dimension };
+    ///         helper.serialize(serializer)
+    ///     }
+    /// }
+    /// ```
```
cmyr, line 237: 😆

cmyr (approved):
works for me!

PR #315, opened 2023-08-02
Bump version

RickyDaMa:
Also upgrades `quick-xml` to the latest as it's our only outdated dependency: https://deps.rs/repo/github/linebender/norad. No breakages or further changes required

Would like a new version released so I can make use of #309 :)


cmyr:
Adding the libs is a breaking change, so we should go to 0.11.

cmyr (approved):
Thanks!

PR #318, opened 2023-08-22
Ser/de Designspace rules

madig:
This reads and writes rules in Designspaces.

Does not support standalone `<condition>` elements outside a `<conditionset>` because it's annoying to implement.

Specification: https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#rules-element

### DIFF
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -23,8 +23,8 @@ pub struct DesignSpaceDocument {
     #[serde(with = "serde_impls::axes", skip_serializing_if = "Vec::is_empty")]
     pub axes: Vec<Axis>,
     /// One or more rules.
-    #[serde(default, with = "serde_impls::rules", skip_serializing_if = "Vec::is_empty")]
-    pub rules: Vec<Rule>,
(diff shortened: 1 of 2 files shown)


src/designspace.rs
```diff
@@ -78,6 +81,82 @@ pub struct AxisMapping {
...
+
+/// Describes a single set of substitution rules.
+///
+/// Does not support standalone `<condition>` elements outside a
+/// `<conditionset>`.
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct Rule {
+    /// Name of the rule.
+    #[serde(rename = "@name")]
+    pub name: Option<String>,
```
cmyr, line 118: let's use `GlyphName` as the type for these?

cmyr (approved):
One note, otherwise lgtm!

src/designspace.rs
madig: Hm, the rule name is not a glyph name, but I'll change the substitution names :)

cmyr: oops sorry yea just the glyph names. :)

PR #319, opened 2023-08-22
Bump version to 0.12.0

cmyr:


cmyr:
Adding new fields to a struct where all fields are public is a breaking change, because you might have previously had code that either created a new struct or destructured an existing one, which will no longer compile. :(

PR #320, opened 2023-08-24
Clear buf

rsheeter:
https://docs.rs/quick-xml/latest/quick_xml/reader/struct.Reader.html suggests "if we don't keep a borrow elsewhere, we can clear the buffer to keep memory usage low." We build data structures with copies so as far as I can tell we can clear buf in all the read event into loops.

While we're here, add magic incantation to Cargo.toml make `cargo bench -- --save-benchmark` work so one can compare against a baseline taken on master. `$ cargo bench -- --baseline master` generally tells me things are more better than worse. The results of reruns are not stunningly consistent but it mostly tells me a couple of things are faster and occasionally that one is slower.

Feel free to close as not convincing enough or merge as you see fit :)

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,11 @@ pretty_assertions = "1.0"
 expect-test = "1.4.1"
 criterion = "0.5"
 
+# https://github.com/bheisler/criterion.rs/issues/193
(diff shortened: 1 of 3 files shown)


cmyr (approved):
Looks good modulo the clippy complaints :)

PR #321, opened 2023-08-24
Make rules processing attribute optional

madig:
Brown-paper-bag PR.

### DIFF
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -89,7 +89,7 @@ pub struct AxisMapping {
 pub struct Rules {
     /// Indicates whether substitution rules should be applied before or after
     /// other glyph substitution features.
-    #[serde(rename = "@processing")]
+    #[serde(default, rename = "@processing")]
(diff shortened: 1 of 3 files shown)


src/designspace.rs
```diff
@@ -53,8 +53,7 @@ pub struct Axis {
     #[serde(rename = "@default")]
     pub default: f32,
     /// Records whether this axis needs to be hidden in interfaces.
-    #[serde(default)]
-    #[serde(rename = "@hidden")]
+    #[serde(default, rename = "@hidden", skip_serializing_if = "is_false")]
```
cmyr, line 56: maybe we can just do,
```suggestion
    #[serde(default, rename = "@hidden", skip_serializing_if = "std::convert::identity")]
```

🤔

cmyr: edit: I tried, it doesn't work :(

cmyr (approved):
👝

PR #329, opened 2023-10-13
Rename `LayerSet` to `LayerContents` & clarify ordering

### LINKED ISSUE #326: `LayerSet` is ordered when it is not implied or documented as such
A set is traditionally unordered, so one would expect `LayerSet` to be unordered. The fact that it is actually ordered wouldn't matter if it wasn't for the fact that its `PartialEq` implementation relies on ordering

This came up as a frustrating footgun while I was fixing a test within a bigger project, and until I questioned norad and looked into its internals I didn't know why I could compare individual layers and have them all be equal, but have two mismatching layer sets

Suggested solutions:
1. Making `LayerSet` actually unordered (though I daresay there's some reason why it is ordered)
2. Documenting that `LayerSet` is ordered, preferably changing it's name to not mislead
3. Make the `PartialEq` implementation be order-insensitive. This may be challenging without incurring allocations

RickyDaMa:
Will close #326 

Documentation also now explicitly states the layers are ordered

---

If we were interested in maintaining backwards compatibility, we could do a re-export under the old name:

```rs
#[deprecated(
    note = "old naming convention, switch to LayerContents instead",
    since = "0.12.2",
)]
pub use LayerContents as LayerSet;
```

### DIFF
--- a/src/font.rs
+++ b/src/font.rs
@@ -13,7 +13,7 @@ use crate::glyph::Glyph;
 use crate::groups::{validate_groups, Groups};
 use crate::guideline::Guideline;
 use crate::kerning::Kerning;
-use crate::layer::{Layer, LayerSet, LAYER_CONTENTS_FILE};
+use crate::layer::{Layer, LayerContents, LAYER_CONTENTS_FILE};
 use crate::name::Name;
 use crate::names::NameList;
 use crate::shared_types::{Plist, PUBLIC_OBJECT_LIBS_KEY};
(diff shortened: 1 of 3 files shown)


cmyr:
I'm happy either way, I'm not sure we have enough outside users to be too concerned about the breakage, but if you're feeling extra kind you should feel free, in which case I'd probably simplify the deprecation message to just "renamed to LayerContents" (but anything is fine really)

cmyr (approved):
Happy with or without the deprecation warning :)

PR #335, opened 2023-11-14
Wrap file writes with `close_already` (perf increase on Windows)

### LINKED ISSUE #334: Improving disk I/O performance on Windows
Full disclosure, this is basically an advert for a little crate I wrote over the weekend: [`close_already`](https://github.com/alpha-tango-kilo/close_already)

In short, by using my crate (under 1000 SLoC including the two transitive dependencies*) **I was able to speed up norad (without rayon) writing a big UFO to disk by 67%**, and a further ~10% when the rayon feature was enabled (despite a suboptimal implementation with conflicting threadpools). You can read more about how this works in the crate's README

I still need to benchmark read performance for exact numbers, but there should also be performance improvements there as well

I've made the library work as a no-op on other compile targets and tried to keep it simple & uninvasive to integrate, so I was wondering if `norad` would be interested in a PR to show what this looks like?

\* EDIT: when using the default `backend-threadpool`

RickyDaMa:
Per the benchmark results from my machine [here](https://github.com/linebender/norad/issues/334#issuecomment-1808503148), we see a 5% decrease in write time with `rayon` enabled, and **50%** decrease in write time with default features

If we don't want the benchmark/Roboto in-tree, let me know and I can remove it. The Roboto is the latest from the repo, with most/all lib keys removed from glyphs

Let me know if there are any writes I've missed wrapping!

If you want to read more and/or audit the new dependency: [alpha-tango-kilo/close_already](https://github.com/alpha-tango-kilo/close_already)

Will close #334

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ kurbo = { version = "0.10.0", optional = true }
 thiserror = "1.0"
 indexmap = { version = "2.0.0", features = ["serde"] }
 base64 = "0.21.2"
+close_already = "0.3"
 
 [dev-dependencies]
 failure = "0.1.6"
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -3,7 +3,7 @@
 #![deny(rustdoc::broken_intra_doc_links)]
 
 use serde::Serialize;
-use std::{fs, fs::File, io::BufReader, path::Path};
+use std::{fs::File, io::BufReader, path::Path};
 
 use plist::Dictionary;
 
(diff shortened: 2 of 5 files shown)


cmyr:
I'm open to merging this, but I would prefer not to add 10mb of test data to this repo. I don't think it's critical that we commit the benchmarks in any case? if we want to have some mechanism for timing the writing of a file, we do have the `load_save` example that will read and write a UFO, and can be passed an arbitrary font file.

cmyr (approved):
Thanks!

PR #336, opened 2023-11-21
Release version 0.14.0

RickyDaMa:
New major version due to the removal of the `druid` feature, and this version will also ship the speed-ups from `close_already`

@cmyr when you have time if you could release this onto crates.io I'd appreciate it :)


cmyr (approved):
oops sorry to let this sit. let's get #337 in as well and then I'll do a release :)

PR #337, opened 2023-11-30
Update quick-xml to 0.31.0

### LINKED ISSUE #330: updating to quick-xml 0.31 breaks our lib deserialization
this has been an unexpected rabbit hole, but basically: if we want to upgrade to the latest quick-xml, we need to figure out how to get our custom deserialization code working again, which I've just burnt a couple of hours on.

cmyr:
Something in quick-xml's internals changed in this version in a way that broke our admittedly hacky plist parsing.

Although the journey was long, the fix was easy: we need to explicitly handle the 'dict' key as a field when we are deserializing the lib.

This also includes a few more explicit 'skip_serializing_if' annotations, since once we were working again we were writing out empty strings in places where we hadn't been on the previous quick-xml.

One note on this impl that just occurs to me: it is possible now that deserialization will fail if the <lib> section of some object contains something other than a dict; we can discuss in review whether or not this is a problem.

- closes #330 
- ate six hours of my life

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ uuid = { version = "1.2", features = ["v4"] }
 serde = { version =  "1.0", features = ["rc", "derive"] }
 serde_derive = "1.0"
 serde_repr = "0.1"
-quick-xml = { version = "0.30.0", features = ["serialize"] }
+quick-xml = { version = "0.31.0", features = ["serialize"] }
 rayon = { version = "1.3.0", optional = true }
 kurbo = { version = "0.10.0", optional = true }
 thiserror = "1.0"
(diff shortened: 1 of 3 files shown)


cmyr:
Now that I think about it we were already enforcing that the lib contains only a dict, just previously it would have been simpler to change this and also accept other plist data types. So no changes in our behaviour. (And we're about to release a new minor anyway, so a breaking change wouldn't be a big deal anyway)

rsheeter:
I'm curious how you figured this out? - I played with it briefly and found it surprisingly opaque what the problem was. Admittedly I'm not over-familiar with norad or quick-xml but ... I'd still have expected more clues than I got.

PR #339, opened 2023-12-29
Fix DesignSpaceDocument lib serialization

### LINKED ISSUE #338: Serialization of Designspaces mangles lib
I noticed that code like

```rust
serde_json::to_value(norad::designspace::DesignSpaceDocument::load(path).unwrap()).unwrap()
```

returned mangled libs, that look like

```js
{array: {string: "_typoquote"}, dict: {dict: {key: "tH", string: "67"}, key: "F333AEC5-BB61-4D5E-9EDD-6124322D28F2"}, key: "public.skipExportGlyphs"}
```

when it should instead be https://gitlab.gnome.org/GNOME/cantarell-fonts/-/blob/cd2155cf5696082cc14706e0f93a9ad7a1b4dab1/src/Cantarell.designspace#L74-156. As in, it seems to mirror the verbatim plist structure of separate key and value elements. I don't understand why that happens...

madig:
TODO:
- [ ] Actually fix serialization
- [ ] Should serialization also remove the "@" character from field names that serves only as a hint for quick-xml?

Closes https://github.com/linebender/norad/issues/338.

### DIFF
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -460,6 +460,87 @@ mod tests {
     }
 
     #[test]
+    fn serialize_lib() {
+        use plist::Value;
+        use serde_test::{assert_tokens, Token};
+
+        // JSON representation of the test lib, abbreviated from Cantarell.designspace:
+        // {
+        //     "GSDimensionPlugin.Dimensions": {
+        //         "0B2E441B-685E-400D-9B9B-E078DEED62EF": {
+        //             "HH": "19",
+        //             "HV": "22"
+        //         },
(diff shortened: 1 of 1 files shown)


madig:
I now realize that the serializer is a special-case implementation for quick-xml. Should the XML serialization stuff be kept in a separate serializer maybe to not interfere with lib serialization?

cmyr:
how much do we care to resolve this? would you like me to pick it up?

madig:
I forgot why I wanted to fix this... maybe to make UFO JSON de/serialisation easier or even possible? If an LLM can quickly take care of this, that'd be nice.

PR #342, opened 2024-03-11
Always retain the default layer

madig:
Previously, you could drop the default layer if it had been renamed.

Code suggested by @RickyDaMa :)

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -237,12 +237,7 @@ impl LayerContents {
     where
         F: FnMut(&Layer) -> bool,
     {
-        self.layers.retain(|layer| {
-            if layer.is_default() {
-                return true;
-            }
-            predicate(layer)
-        });
+        self.layers.retain(|layer| layer.is_default() || predicate(layer));
     }
 
     /// Removes any layers that contain no glyphs


src/layer.rs
```diff
@@ -237,7 +237,15 @@ impl LayerContents {
     where
         F: FnMut(&Layer) -> bool,
     {
-        self.layers.retain(|layer| layer.name == DEFAULT_LAYER_NAME || predicate(layer))
+        // Always keep the default layer, which is always the first.
```
anthrotype, line 240: who says that a default layer has to be the first one? Default layer is simply the one that points to the default "glyphs" folder, it can have any name ("public.default" is simply the name for a default layer without an explicit name and is not sufficient to mark a layer as "default" ) and can be in any position not necessarily the first. Unless you reordered the layers (which you should have not).

anthrotype: > Unless you reordered the layers (which you should have not).

oh I see you actually do reorder the layers:

https://github.com/linebender/norad/blob/0b9e43334551861dbdd75732785d0cb5f3431568/src/layer.rs#L88-L93

I think this is incorrect, the UFO spec never says the default should be the first.

anthrotype:
The correct fix that doesn't depend on any specific layer name or ordering is to treat as default the layer whose path points to the default glyphs directory

src/layer.rs
anthrotype: Filed https://github.com/linebender/norad/issues/343

anthrotype (approved):

cmyr (approved):
looks good, thanks!

PR #347, opened 2024-05-21
Implement `TryFrom<String>` for `Name`

RickyDaMa:
Allows for re-using the allocated memory if you already have the owned type, and also avoids allocating in the error case 😎 

This addresses that invasive brain thought I get when I have to rely on the compiler to realise it can re-use an allocation, instead of explicitly having a code path for it

This PR also contains an unrelated commit to fix a clippy lint

### DIFF
--- a/src/serde_xml_plist.rs
+++ b/src/serde_xml_plist.rs
@@ -38,7 +38,7 @@ mod de {
 
     pub(super) struct DictWrapper(pub(super) Dictionary);
 
-    struct ValueWrapper(Value);
+    struct ValueWrapper;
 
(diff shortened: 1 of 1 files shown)


cmyr (approved):
This looks fine, but unfortunately I do not believe it reuses the allocation? For instance it doesn't appear that `From<String>` for `Arc<str>` reuses an allocation (I'm inferring this, but [the docs](https://doc.rust-lang.org/1.78.0/std/sync/struct.Arc.html#method.from-2) imply that the data is copied to the new allocation)

PR #350, opened 2024-07-08
Use explicit rather than implicit features.

waywardmonkeys:


waywardmonkeys:
This is planned to be an error in Rust 2024 edition: https://doc.rust-lang.org/edition-guide/rust-2024/cargo-remove-implicit-features.html

alerque (approved):

PR #351, opened 2024-07-22
Remove usages of `extern crate` syntax.

waywardmonkeys:


waywardmonkeys:
> Would this also mean that we can remove this dependency from the Cargo.toml?

No ... that's what makes it available for use. Just that since Rust 2018 edition, the `extern crate` is no longer needed, just having it in the `Cargo.toml` is enough.

> Thanks for your continual efforts to modernise the repo; is there a particular motivation behind this PR beyond just bringing norad more up-to-date?

Sadly, no! I just help out on a bunch of the linebender repos and this is in the linebender organization!

I don't even know what UFO files are (beyond what the readme says).

PR #357, opened 2024-10-04
Strip BOM from glif xml if present

cmyr:
TIL that utf-8 even _has_ a byte order mark 🚑

### DIFF
--- a/testdata/bom_glif.glif
+++ b/testdata/bom_glif.glif
@@ -1,0 +1,25 @@
+﻿<!-- this includes a byte order mark (https://en.wikipedia.org/wiki/Byte_order_mark)-->
+<?xml version="1.0" encoding="UTF-8"?>
+<glyph name=".notdef" format="2">
+  <advance width="600" height="1000"/>
+  <outline>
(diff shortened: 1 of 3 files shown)


testdata/bom_glif.glif
```diff
@@ -0,0 +1,25 @@
...
+      <point x="454" y="663" type="line"/>
+      <point x="454" y="51" type="line"/>
+    </contour>
+    <contour>
+      <point x="94" y="0" type="line"/>
+      <point x="505" y="0" type="line"/>
+      <point x="505" y="714" type="line"/>
+      <point x="94" y="714" type="line"/>
+    </contour>
+  </outline>
```
rsheeter, line 18: optional: guessing you don't even need an outline?

cmyr: was this intended to come with a +1?

rsheeter (approved):

testdata/bom_glif.glif
rsheeter: it was!

PR #364, opened 2025-01-24
Compile under wasm-pack

simoncozens:
This doesn't yet make norad do anything sensible under a WASM target, but it does make it compile and hence gives us a small hope of compiling fontc.


cmyr (approved):
Thanks!

PR #383, opened 2026-01-13
Add support for avar2 <mappings> element

simoncozens:
This adds support for the avar2 [`<mappings>` element](https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#mappings-element) which lives inside of `<axes>`. This of course means that `<axes>` is no longer a pure `Vec<Axis>`, which means we have to do some shenanigans to support the new element *and* maintain the existing API. `#[serde(from..., into...)]` provides the shenanigans, and we add a separate top-level `axis_mappings` field.

### DIFF
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -15,25 +15,21 @@ use crate::Name;
 ///
 /// [designspace]: https://fonttools.readthedocs.io/en/latest/designspaceLib/index.html
 #[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
-#[serde(rename = "designspace")]
+#[serde(from = "RawDesignSpaceDocument", into = "RawDesignSpaceDocument")]
 pub struct DesignSpaceDocument {
     /// Design space format version.
-    #[serde(rename = "@format")]
     pub format: f32,
     /// One or more axes.
-    #[serde(with = "serde_impls::axes", skip_serializing_if = "Vec::is_empty")]
     pub axes: Vec<Axis>,
+    /// Optional avar2-style axis mappings.
+    pub axis_mappings: Option<AxisMappings>,
     /// One or more rules.
-    #[serde(default, skip_serializing_if = "Rules::is_empty")]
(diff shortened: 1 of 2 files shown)


src/designspace.rs
```diff
@@ -291,6 +321,68 @@ impl Rules {
...
+    sources: Vec<Source>,
+    #[serde(default, with = "serde_impls::instances", skip_serializing_if = "Vec::is_empty")]
+    instances: Vec<Instance>,
+    #[serde(default, with = "serde_plist", skip_serializing_if = "Dictionary::is_empty")]
+    lib: Dictionary,
+}
+
+/// Internal container for axes and their optional mappings.
+#[derive(Clone, Debug, Default, PartialEq)]
+struct AxesContainer {
```
cmyr: I might call these `RawAxes` to match `RawDesignSpaceDocument`?

src/designspace.rs
```diff
@@ -379,8 +471,47 @@ mod serde_impls {
 
     serde_from_field!(location, dimension, crate::designspace::Dimension);
     serde_from_field!(instances, instance, crate::designspace::Instance);
-    serde_from_field!(axes, axis, crate::designspace::Axis);
     serde_from_field!(sources, source, crate::designspace::Source);
+
+    /// Custom (de)serialization for the axes container that holds both axes and mappings.
+    pub(super) mod axes_container {
```
cmyr: Is this necessary? Could we just derive `Serialize, Deserialize` for AxesContainer directly?

cmyr (approved):
Looks good, one little question inline but otherwise happy to get this merged. :)

src/designspace.rs
simoncozens: Oh hey, that's considerably simpler, yes...

cmyr (approved):
thanks!

cmyr:
@simoncozens would you like a release?

simoncozens:
Sure! Thanks!

Issue #386, opened 2026-02-13
Open files via hashmap


yanone:
I'm building a font editor and want to be able to open `.ufo/.designspace` files through `babelfont-rs`. Since wasm doesn't support native file system access and it can't be spoofed to it either, the simple solution is to build a hashmap and have `norad` open the files through it. 

Please see this branch comparison for the changes: https://github.com/linebender/norad/compare/master...yanone:norad:hasmap-loading?expand=1

I would create a PR if you are interested and then `babelfont-rs` would need its own PR to call the new methods.

Please note that I'm not very familiar with Rust, so feel free to request any changes. It works over here.

cc @simoncozens

PR #387, opened 2026-02-23
Add in-memory UFO loader and path helpers

yanone:
Introduce Font::load_entries to load a Font from in-memory UFO entries (HashMap of relative paths -> contents). The implementation normalizes virtual paths, detects an effective UFO root, parses metainfo/fontinfo/lib/groups/kerning/features, reads layer descriptors and glyph GLIFs, and normalizes negative openTypeOS2WinDescent values. Added utility functions normalize_virtual_path and join_virtual_path, plus necessary imports (Cursor, BTreeMap, HashMap). Also added DesignSpaceDocument::load_str and unit tests covering basic entry loading and negative OS/2 winDescent handling.

The purpose for this addition is that it allows tools with no file system access (such as wasm components) to parse .ufo/.designspace fonts.

I had opened [an issue](https://github.com/linebender/norad/issues/386) for that and thought I'd try to add some momentum with a PR. Cheers.

### DIFF
--- a/src/font.rs
+++ b/src/font.rs
@@ -3,7 +3,9 @@
 #![deny(rustdoc::broken_intra_doc_links)]
 
 use std::fs;
+use std::io::Cursor;
 use std::path::{Path, PathBuf};
+use std::{collections::BTreeMap, collections::HashMap};
 
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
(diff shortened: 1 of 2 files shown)


src/font.rs
```diff
@@ -212,6 +214,167 @@ impl Font {
...
+        })?;
+
+        let fontinfo_key = join_virtual_path(effective_root, FONTINFO_FILE);
+        if let Some(fontinfo_str) = normalized_entries.get(&fontinfo_key) {
+            let mut fontinfo_value: plist::Value =
+                plist::from_reader(Cursor::new(fontinfo_str.as_bytes())).map_err(|source| {
+                    FontLoadError::ParsePlist { name: FONTINFO_FILE, source }
+                })?;
+
+            if let Some(dict) = fontinfo_value.as_dictionary_mut() {
```
madig: This is curious, why is this manipulating data here?

yanone: That was a compatibility layer that accidentally remained in the code where I had to deal with a font with a negative winDescent value which broke the parsing. I just patched it.

yanone: I also updated the formatting. All checks pass now.

RickyDaMa:
Broader question than this PR: did we ought to have some file system abstraction trait in `norad`, to allow various use cases like this one? IIRC fontTools does something similar to allow it to read UFOs and UFOZ easily. A public trait would also then mean that other file system support wouldn't even necessarily need to be maintained here, but one could have their own crate `norad-wasm` to add the support they need. Just thinking aloud.

madig:
Are there existing filesystem abstractions in wider use?

RickyDaMa:
The closest thing would be accepting `impl io::Read` and similar, instead of FS-specific types

src/font.rs
```diff
@@ -212,6 +214,146 @@ impl Font {
...
 
+    /// Returns a [`Font`] object loaded from in-memory UFO entries.
+    ///
+    /// `path` is the virtual UFO root path (for example `MyFont.ufo` or `sources/MyFont.ufo`).
+    ///
+    /// `entries` should contain relative file paths and UTF-8 file contents.
+    pub fn load_entries(
+        path: impl AsRef<Path>,
+        entries: &HashMap<String, String>,
+    ) -> Result<Font, FontLoadError> {
```
cmyr, line 225: I want to think a bit about the API here and make sure that we get it right.

- Is there a reason to use `String` instead of `PathBuf` as a way of identifying the files? to me `PathBuf` would make more sense, even if we aren't referring to files on disk, because it handles all the logic for handling components, separators, etc.
- Do we need a root path, or could it be implied? Is it mostly useful for error messages?
- Instead of needing to pass in a `HashMap`, another option would be for this constructor to take a closure that would then return various resources, e.g. it could look like,

    ```rust
    fn load_entries(entires: impl Fn(&Path) -> Option<String>) -> Result<Font, Error>
    ```
    this isn't a huge difference, but it just makes the API more flexible at the call site, instead of requiring the caller to always put everything in a hashmap.


Curious to hear your thoughts, and we can iterate a bit from there?

src/font.rs
```diff
@@ -212,6 +214,146 @@ impl Font {
...
+            .ok_or(FontLoadError::MissingMetaInfoFile)?;
+        let root = root.trim_end_matches('/').to_string();
+
+        let rooted_metainfo_key = join_virtual_path(&root, METAINFO_FILE);
+        let effective_root =
+            if normalized_entries.contains_key(&rooted_metainfo_key) { root.as_str() } else { "" };
+
+        let metainfo_key = join_virtual_path(effective_root, METAINFO_FILE);
+        let metainfo_str =
+            normalized_entries.get(&metainfo_key).ok_or(FontLoadError::MissingMetaInfoFile)?;
```
cmyr, line 241: so is the idea here that the caller can provide either `MyFont.ufo/metainfo.plist` _or_ just `metainfo.plist`? I would prefer to enforce stricter requirements on the caller, and just insist that all paths are relative to the (implied) root of the ufo directory?

src/font.rs
```diff
@@ -212,6 +214,146 @@ impl Font {
...
+
+        let mut ufo = Font::new();
+        ufo.meta = plist::from_reader(Cursor::new(metainfo_str.as_bytes()))
+            .map_err(|source| FontLoadError::ParsePlist { name: METAINFO_FILE, source })?;
+
+        let fontinfo_key = join_virtual_path(effective_root, FONTINFO_FILE);
```
cmyr, line 247: yea it would be nice if we were just always using `FONTINFO_FILE` as the key and not worrying about a possible root..

cmyr (commented):
Sorry to let this sit! Overall I think that supporting this use-case makes sense, but I have some general questions about what the API should look like. I've left various questions inline, and once I make sure I'm understanding various decisions we can figure out what the best API would be.

yanone:
I'll propose another solution in a couple of days, maybe tomorrow, that's independent of the hashmaps, along the lines of what @RickyDaMa proposed. A general file system abstraction trait where I implement the hashmaps on my side while norad uses a file system wrapper that could wrap anything other than the native file system. It would also be the smallest-possible code change to norad. I'll be back.

cmyr:
So if you're going this route I will point you to how we handled this case in `fea-rs` (our FEA parser): [SourceResolver](https://docs.rs/fea-rs/latest/fea_rs/parse/trait.SourceResolver.html) is the abstraction, and then we have a particular implementation for ufos on the file system here at [FileSystemResolver](https://docs.rs/fea-rs/latest/fea_rs/parse/struct.FileSystemResolver.html).

yanone:
@cmyr 

Here are two proposals. I didn't create PRs yet, but you can compare the branches. Both are a bit verbose because norad reads files in many different places.

The [first](https://github.com/linebender/norad/compare/master...yanone:norad:source-resolver-loading?expand=1) follows your suggestion and implements **SourceResolver**.

Pro: explicit abstraction, cleaner long-term architecture, easiest to extend (different backends).
Con: larger API surface and slightly heavier review/maintenance burden.

The [second](https://github.com/linebender/norad/compare/master...yanone:norad:minimal-reader-loading?expand=1) is a minimal **callback-reader** approach. Less verbose changes, but SourceResolver appears to be the better architecture.

Pro: smaller and more merge-friendly; keeps current APIs mostly intact with one extra loading hook.
Con: less structured than a trait-based abstraction; weaker as a reusable public abstraction.

cmyr:
Honestly they both seem fine, so I'm happy with whichever you prefer. It might be that the trait-based abstraction has additional features that are not necessary for this case, which might suggest that the callback makes sense? But no strong feelings, feel free to open a PR with whichever you prefer :)

yanone:
Closing this PR in favour of https://github.com/linebender/norad/pull/390

PR #388, opened 2026-03-06
If a v1 .glif contains a note field, just drop it

cmyr:
Instead of erroring. This exists in a few real-world fonts, and doesn't seem worth failing over.

see https://github.com/BlackFoundryCom/Atma/blob/master/Sources/Mastering/TT/Atma-SemiBold.ufo/glyphs/bn_e.glif

noticed this was a cause of some failed compiles on crater.


dfrg (approved):
🤘

PR #390, opened 2026-03-13
SourceResolver loading

yanone:
**Summary**
This PR introduces a path-based source abstraction for UFO loading so norad can load from non-filesystem backends without exposing a HashMap-specific API.

**Motivation**
Downstream wasm and in-memory workflows need to provide UFO files from virtual storage.
The goal is to keep norad’s public API backend-agnostic while still supporting those use cases cleanly.

**What changed**
Added a source resolver abstraction for reading UFO resources by relative path.
Added a filesystem resolver implementation for current disk-based behavior.
Added resolver-based font loading flow, used by existing load entry points.
Added resolver-specific error handling for missing files and IO failures.
Kept default filesystem behavior intact for existing callers.

**Compatibility**
Existing filesystem loading behavior is preserved.
No HashMap-specific type is introduced in the public API.
This creates a clear extension point for custom loaders in downstream crates.

### DIFF
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -312,6 +312,11 @@ impl DesignSpaceDocument {
         close_already::fs::write(path, buf)?;
         Ok(())
     }
+
+    /// Returns a [`DesignSpaceDocument`] loaded from an XML string.
+    pub fn load_str(contents: &str) -> Result<DesignSpaceDocument, DesignSpaceLoadError> {
+        quick_xml::de::from_str(contents).map_err(DesignSpaceLoadError::DeError)
+    }
 }
 
 impl Rules {
--- a/src/error.rs
+++ b/src/error.rs
@@ -142,6 +142,17 @@ pub enum FontLoadError {
         /// The underlying error.
         source: PlistError,
     },
+    /// Failed to load a file through a custom source resolver.
+    #[error("failed to read '{path}' from source resolver")]
+    ResolverIo {
+        /// The requested path.
+        path: PathBuf,
+        /// The underlying error.
+        source: IoError,
+    },
+    /// Resolver-based loading currently supports only UFO v3 fonts.
+    #[error("resolver-based loading currently supports only UFO v3")]
+    ResolverUnsupportedFormatVersion,
     /// Norad can currently only open UFO (directory) packages.
     #[error("only UFO (directory) packages are supported")]
     UfoNotADir,
--- a/src/font.rs
+++ b/src/font.rs
@@ -2,7 +2,9 @@
 
 #![deny(rustdoc::broken_intra_doc_links)]
 
+use std::collections::BTreeMap;
 use std::fs;
+use std::io::{Cursor, ErrorKind};
 use std::path::{Path, PathBuf};
 
 use serde::{Deserialize, Serialize};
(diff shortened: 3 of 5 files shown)


src/designspace.rs
```diff
@@ -312,6 +312,11 @@ impl DesignSpaceDocument {
         close_already::fs::write(path, buf)?;
         Ok(())
     }
+
+    /// Returns a [`DesignSpaceDocument`] loaded from an XML string.
+    pub fn load_str(contents: &str) -> Result<DesignSpaceDocument, DesignSpaceLoadError> {
+        quick_xml::de::from_str(contents).map_err(DesignSpaceLoadError::DeError)
+    }
```
cmyr, line 319: total nit but I would prefer this to be right below the  main `load` fn, above.

src/error.rs
```diff
@@ -142,6 +142,17 @@ pub enum FontLoadError {
         /// The underlying error.
         source: PlistError,
     },
+    /// Failed to load a file through a custom source resolver.
+    #[error("failed to read '{path}' from source resolver")]
+    ResolverIo {
+        /// The requested path.
+        path: PathBuf,
+        /// The underlying error.
+        source: IoError,
```
cmyr, line 151: I don't think `io::Error` is the right error type, here; the error type on `SourceResolver` also isn't right, see below.

src/font.rs
```diff
@@ -34,6 +36,73 @@ static DEFAULT_METAINFO_CREATOR: &str = "org.linebender.norad";
...
 
+/// Abstraction over loading UTF-8 UFO files.
+///
+/// This can be implemented by filesystem-backed loaders, in-memory maps,
+/// or custom loaders.
+pub trait SourceResolver {
+    /// Return UTF-8 contents for the provided path.
+    ///
+    /// Returning `Ok(None)` indicates a missing file.
+    fn get_contents(&self, path: &Path) -> Result<Option<String>, FontLoadError>;
```
cmyr, line 47: What we want here is some very customizable error type that the implementor can use to communicate why the given resource isn't available; `FontLoadError` covers a lot of specific cases that aren't relevant to the job of "give me a string for this path".

The 'best' option is probably an associated type, so this would look like:


```rust
trait SourceResolver {
    type Error;
    fn get_contents(&self, path: &Path) -> Result<Option<String>, Self::Error>;
}
```

but this might be a bit fiddly if you're not that comfortable with rust? Because we would probably need bounds on the error type elsewhere, so that the top-level `Resolver` variant would contain a `Box<dyn std::error::Error>`...

src/font.rs
```diff
@@ -34,6 +36,73 @@ static DEFAULT_METAINFO_CREATOR: &str = "org.linebender.norad";
...
+
+    /// Resolve a raw request path.
+    fn resolve_raw_path(&self, path: &Path, _requested_from: Option<&Path>) -> PathBuf {
+        path.to_path_buf()
+    }
+
+    /// Canonicalize a path if needed.
+    fn canonicalize(&self, path: &Path) -> Result<PathBuf, FontLoadError> {
+        Ok(path.to_path_buf())
+    }
```
cmyr, line 57: both of these methods are important for resolving include statements in a FEA file but are (I believe) irrelevant for your use-case, and can be removed.

src/font.rs
```diff
@@ -34,6 +36,73 @@ static DEFAULT_METAINFO_CREATOR: &str = "org.linebender.norad";
...
+            return parent.join(path);
+        }
+        self.project_root.join(path)
+    }
+
+    fn canonicalize(&self, path: &Path) -> Result<PathBuf, FontLoadError> {
+        Ok(path.to_path_buf())
+    }
+}
+
```
cmyr, line 105: do you even need a file-system backed resolver? this is used in fea-rs because the resolver API is the only API for handling `include` statements, but norad already has an API for loading from the file system...

src/font.rs
```diff
@@ -212,6 +281,115 @@ impl Font {
...
+        request: DataRequest,
+        resolver: impl SourceResolver,
+    ) -> Result<Font, FontLoadError> {
+        Self::load_with_resolver_impl(request, &resolver)
+    }
+
+    fn load_with_resolver_impl(
+        request: DataRequest,
+        resolver: &dyn SourceResolver,
+    ) -> Result<Font, FontLoadError> {
```
cmyr, line 301: have a separate `_impl` method that we delegate too is unnecessary here (and is actually unnecessary in our main `load` fn too now, I believe it's a relic from an earlier time)

src/font.rs
```diff
@@ -212,6 +281,115 @@ impl Font {
...
+        let metainfo_str = required_file(resolver, Path::new(METAINFO_FILE), None)?;
```
cmyr, line 302: this `required_file` method is only called once, i'd just inline that bit?

src/font.rs
```diff
@@ -212,6 +281,115 @@ impl Font {
...
+        let mut meta: MetaInfo = plist::from_reader(Cursor::new(metainfo_str.as_bytes()))
+            .map_err(|source| FontLoadError::ParsePlist { name: METAINFO_FILE, source })?;
+
+        if meta.format_version != FormatVersion::V3 {
+            return Err(FontLoadError::ResolverUnsupportedFormatVersion);
+        }
+
+        let mut lib = if request.lib {
+            match optional_file(resolver, Path::new(LIB_FILE), None)? {
```
cmyr, line 311: I think `optional_file` can also just be removed, we don't need to worry about canonicalization etc so should just call `resolver.get_contents` directly?

src/font.rs
```diff
@@ -212,6 +281,115 @@ impl Font {
...
+                Some(lib_str) => plist::Value::from_reader(Cursor::new(lib_str.as_bytes()))
```
cmyr, line 312: you shouldn't need `Cursor` here, just `Value::from_reader_xml(lib_str.as_bytes())`?

src/font.rs
```diff
@@ -212,6 +281,115 @@ impl Font {
...
+
+        let features = if request.features {
+            optional_file(resolver, Path::new(FEATURES_FILE), None)?.unwrap_or_default()
+        } else {
+            Default::default()
+        };
+
+        let layers = load_layer_set_from_resolver(resolver, &request.layers)?;
+
+        let (groups, kerning) = (groups, kerning);
```
cmyr, line 376: ?

cmyr (commented):
so... lots of notes inline, but I think the general shape of this is okay?

- going through this does make me wonder if the closure-based version might not be preferable, since it's easier to get right without needing a lot of back-and-forth?
- In either case I think I'd like to see all of this new code live in a new module, like `non_file_loading.rs` or similar, just so that we can keep it a bit more contained.

yanone:
I'm addressing all of your remarks in a new branch. Before making a PR, see diffs here: https://github.com/linebender/norad/compare/master...yanone:norad:non-file-io-save-and-includes?expand=1

This branch is a follow-up over both #387 and #390.

Compared to #387, it keeps the non-file use case but drops the HashMap-specific API shape in favor of closure-based source/sink callbacks over UFO-root-relative paths.

Compared to #390, it removes the public resolver/filesystem-resolver abstraction and moves the non-file code into a separate module. It also generalizes the error path away from io::Error/resolver-specific types.

In addition, it goes beyond both earlier PRs in two ways that matter for my editor use case:
- feature file `include()` support from the same non-file source, while preserving included `.fea` files structurally instead of flattening them on save
- non-file writing APIs for both UFOs and designspace documents

So the current branch is basically: closure-based non-file load/save, isolated implementation, `include`-aware feature handling, and structural round-tripping of included feature files.

Please review. If you're generally happy with this, I'll make a PR and then we can discuss details.

cmyr:
yan please just open the new PR and I can review there :)

yanone:
Partially implemented by #405

PR #392, opened 2026-03-14
More lenient parsing of glyph points

### LINKED ISSUE #391: should `smooth="yes"` on off-curve points be a warning instead of an error?
norad currently rejects `.glif` files that have `smooth="yes"` on off-curve points with `ErrorKind::UnexpectedSmooth`. While the [UFO spec](https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#point) says smooth "must not be set for off-curve points", many real-world fonts in the wild have this, likely produced by older/buggy font editors. The attribute is meaningless on off-curve points and can be safely ignored.

Example from [etunni/arapey](https://github.com/etunni/arapey) (`Arapey-Regular.otf.ufo/glyphs/H_.glif`):

```xml
<point x="470" y="24" smooth="yes"/>
<point x="464" y="18" smooth="yes"/>
<point x="456" y="18" type="curve" smooth="yes"/>
```

The first two points have no `type` attribute (off-curve), but carry `smooth="yes"`. 166 out of ~260 glyph files in this font have the same issue.
Because of this, fontc fails to compile any UFO source containing this anti-pattern:

```
Reading source failed for '…/H_.glif': 'failed to parse glyph data: unexpected smooth attribute on an off-curve point'
```

Maybe norad could strip `smooth` on off-curve points during parsing (set it to `false`) and optionally emit a warning, rather than returning an error?

Note that `fontTools.ufoLib.glifLib` also has a similar validation and _would_ also reject these fonts, however fontmake fontmake by default loads UFOs with `validate_ufo=False` which in turn passes `validate=False` to glifLib's parsing functions, hence the smooth check is silently skipped and compilation completes regardless.

cmyr:
Specifically if an off-curve point has the 'smooth' attribute set, we will now log a warning instead of returning an error.

closes #391

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -536,6 +536,8 @@ pub enum ErrorKind {
     /// Has an unexpected move definition.
     UnexpectedMove,
     /// Has an unexpected smooth definition.
+    //TODO: no longer used, can be removed next non-patch release
+    #[deprecated(since = "0.18.2", note = "unused, will be removed")]
     UnexpectedSmooth,
     /// Has an unexpected element definition.
     UnexpectedElement,
@@ -606,6 +608,7 @@ impl std::fmt::Display for ErrorKind {
             UnexpectedMove => {
                 write!(f, "unexpected move point, can only occur at start of contour")
             }
+            #[allow(deprecated)]
             UnexpectedSmooth => write!(f, "unexpected smooth attribute on an off-curve point"),
             UnexpectedElement => write!(f, "unexpected element"),
             UnexpectedAttribute => write!(f, "unexpected attribute"),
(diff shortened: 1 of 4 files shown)


src/error.rs
```diff
@@ -536,6 +536,7 @@ pub enum ErrorKind {
     /// Has an unexpected move definition.
     UnexpectedMove,
     /// Has an unexpected smooth definition.
+    //TODO: no longer used, can be removed next non-patch release
```
RickyDaMa, line 539: Maybe stick a `#[deprecated(since = "new-ver", note = "will no longer be produced, scheduled for removal")]` on this in the interim?

cmyr: good call!

anthrotype (approved):

PR #394, opened 2026-03-20
Implement kerning lookups

RickyDaMa:
Every so often I get surprised that this isn't already part of norad, so I threw something together.

I've made a slow version for people only doing a couple kerning lookups and don't want the overhead of generating the reverse group lookup, and then there's also a `ReverseGroupsLookup` struct now which does exactly what it says on the tin and can get used to accelerate kerning lookups.

There is currently no check that you are using the `ReverseGroupsLookup` that corresponds to the `Font` you're calling the lookup on, but I'll document that caveat when I throw together some documentation ("garbage in, garbage out" or words to that effect). I could tie the two together with ✨ fancy ✨ API/types if we wanted.

I also made constants for the kerning group prefixes and did a quick & dirty replace throughout the codebase, so I can let autocomplete do the work instead of worrying that I've typo'd. Bikeshedding encouraged, I don't like how long their names are.

### DIFF
--- a/src/font.rs
+++ b/src/font.rs
@@ -15,7 +15,7 @@ use crate::fontinfo::FontInfo;
 use crate::glyph::Glyph;
 use crate::groups::{validate_groups, Groups};
 use crate::guideline::Guideline;
-use crate::kerning::Kerning;
+use crate::kerning::{Kerning, ReverseGroupsLookup};
 use crate::layer::{Layer, LayerContents, LAYER_CONTENTS_FILE};
 use crate::name::Name;
 use crate::names::NameList;
@@ -604,14 +604,47 @@ impl Font {
     pub fn guidelines_mut(&mut self) -> &mut Vec<Guideline> {
         self.font_info.guidelines.get_or_insert_with(Default::default)
     }
+    
+    #[inline]
+    pub fn get_reverse_groups_lookup(&self) -> ReverseGroupsLookup {
+        ReverseGroupsLookup::from(&self.groups)
+    }
+
+    pub fn kerning_lookup(&self, lookup: &ReverseGroupsLookup, first: &str, second: &str) -> Option<f64> {
+        // glyph name glyph name
+        if let Some(kern) = self.kerning_lookup_dumb(first, second) {
+            return Some(kern);
+        }
+
+        // group name glyph name
+        let first_group = lookup.get_first(first);
+        if let Some(first_group) = &first_group {
+            if let Some(kern) = self.kerning_lookup_dumb(first_group.as_str(), second) {
+                return Some(kern);
+            }
+        }
+
+        // glyph name group name
+        let second_group = lookup.get_second(second);
+        if let Some(second_group) = &second_group {
+            if let Some(kern) = self.kerning_lookup_dumb(first, second_group.as_str()) {
+                return Some(kern);
+            }
+        }
+
+        // group name group name
+        if let Some((first_group, second_group)) = first_group.zip(second_group) {
+            if let Some(kern) = self.kerning_lookup_dumb(first_group.as_str(), second_group.as_str()) {
+                return Some(kern);
+            }
+        }
+
+        None
+    }
 
     pub fn kerning_lookup_slow(&self, first: &str, second: &str) -> Option<f64> {
-        let basic_lookup = |first: &str, second: &str| {
-            self.kerning.get(first).and_then(|first| first.get(second)).copied()
-        };
-
         // glyph name glyph name
-        if let Some(kern) = basic_lookup(first, second) {
+        if let Some(kern) = self.kerning_lookup_dumb(first, second) {
             return Some(kern);
         }
 
@@ -623,7 +656,7 @@ impl Font {
                 },
             );
         if let Some(first_group) = first_group {
-            if let Some(kern) = basic_lookup(first_group.as_str(), second) {
+            if let Some(kern) = self.kerning_lookup_dumb(first_group.as_str(), second) {
                 return Some(kern);
             }
         }
@@ -636,19 +669,23 @@ impl Font {
                 },
             );
         if let Some(second_group) = second_group {
-            if let Some(kern) = basic_lookup(first, second_group.as_str()) {
+            if let Some(kern) = self.kerning_lookup_dumb(first, second_group.as_str()) {
                 return Some(kern);
             }
         }
 
         // group name group name
         if let Some((first_group, second_group)) = first_group.zip(second_group) {
-            if let Some(kern) = basic_lookup(first_group.as_str(), second_group.as_str()) {
+            if let Some(kern) = self.kerning_lookup_dumb(first_group.as_str(), second_group.as_str()) {
                 return Some(kern);
             }
         }
 
         None
+    }
+
+    fn kerning_lookup_dumb(&self, first: &str, second: &str) -> Option<f64> {
+        self.kerning.get(first).and_then(|first| first.get(second)).copied()
     }
 }
 
--- a/src/font.rs
+++ b/src/font.rs
@@ -620,7 +620,7 @@ impl Font {
 
     /// Retrieve the kerning value (if any) between a pair of elements.
     ///
-    /// The elments can be either individual glyphs (by name) or kerning groups
+    /// The elements can be either individual glyphs (by name) or kerning groups
     /// (by name), or any combination of the two.
     //  ^ note: this works without any special consideration in the code
     //          because glyph names are forbidden from using the group prefix,
@@ -673,7 +673,7 @@ impl Font {
 
     /// Retrieve the kerning value (if any) between a pair of elements.
     ///
-    /// The elments can be either individual glyphs (by name) or kerning groups
+    /// The elements can be either individual glyphs (by name) or kerning groups
     /// (by name), or any combination of the two.
     //  ^ note: this works without any special consideration in the code
     //          because glyph names are forbidden from using the group prefix,
--- a/src/font.rs
+++ b/src/font.rs
@@ -604,6 +604,52 @@ impl Font {
     pub fn guidelines_mut(&mut self) -> &mut Vec<Guideline> {
         self.font_info.guidelines.get_or_insert_with(Default::default)
     }
+
+    pub fn kerning_lookup_slow(&self, first: &str, second: &str) -> Option<f64> {
+        let basic_lookup = |first: &str, second: &str| {
+            self.kerning.get(first).and_then(|first| first.get(second)).copied()
+        };
+
+        // glyph name glyph name
+        if let Some(kern) = basic_lookup(first, second) {
+            return Some(kern);
+        }
+
+        // group name glyph name
+        let first_group =
+            self.groups.iter().filter(|(name, _)| name.starts_with("public.kern1.")).find_map(
+                |(name, members)| {
+                    members.iter().any(|glyph_name| glyph_name.as_str() == first).then_some(name)
+                },
+            );
+        if let Some(first_group) = first_group {
+            if let Some(kern) = basic_lookup(first_group.as_str(), second) {
+                return Some(kern);
+            }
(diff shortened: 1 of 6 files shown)


src/kerning.rs
```diff
@@ -14,6 +17,44 @@ use crate::Name;
 /// We use a [`BTreeMap`] because we need sorting for serialization.
 pub type Kerning = BTreeMap<Name, BTreeMap<Name, f64>>;
 
+#[derive(Debug)]
+pub struct ReverseGroupsLookup {
```
Hoolean: nit: lookup is quite an overloaded term in fonts, rename to cache, or even just `ReverseGroups`?

src/font.rs
```diff
@@ -604,6 +606,100 @@ impl Font {
     pub fn guidelines_mut(&mut self) -> &mut Vec<Guideline> {
         self.font_info.guidelines.get_or_insert_with(Default::default)
     }
+
+    #[inline]
+    pub fn get_reverse_groups_lookup(&self) -> ReverseGroupsLookup {
+        ReverseGroupsLookup::from(&self.groups)
+    }
```
Hoolean: Pseudo API suggestion to avoid the potential mismatch and abstract away some of the caching bits (probably similar to what you were proposing in the PR description):

```
struct KerningResolver { kerning: &Kerning, groups: &Groups, cache: Option<ReverseGroupsLookup> };

impl KerningResolver {
    pub fn lookup(first: &str, second: &str) -> Option<f64> {
        // [...] same as before, but using below, dumb lookup is just direct operation on kerning dict
    }

    pub fn first_group(glyph: &str) -> Option<&str> {
        self.cache.map(|cache| cache.first(first)).unwrap_or(|| manual stuff| ...)
    }
}

font.kerning_resolver().lookup("a", "b")
font.kerning_resolver_uncached().lookup("a", "b")
```

I trust your Rusty instincts over mine to arrive at something even more idiomatic than the above notes though 💪

Hoolean (commented):
Thanks for tackling, would love to have something like this in the shared Rust toolbox

src/font.rs
RickyDaMa: Yep this is moreorless exactly what I was thinking of if we were to go down this route. Chose to keep it simple for the first draft. I like the name `KerningResolver`!

src/font.rs
```diff
@@ -604,6 +606,100 @@ impl Font {
...
+        lookup: &ReverseGroupsLookup,
+        first: &str,
+        second: &str,
+    ) -> Option<f64> {
+        // glyph name glyph name
+        if let Some(kern) = self.kerning_lookup_dumb(first, second) {
+            return Some(kern);
+        }
+
+        // group name glyph name
```
RickyDaMa: Self-review: glyph name group name must come first: https://unifiedfontobject.org/versions/ufo3/kerning.plist/#exception-conflict-resolution

src/kerning.rs
```diff
@@ -1,9 +1,16 @@
...
 use serde::ser::{SerializeMap, Serializer};
 use serde::Serialize;
 
-use crate::Name;
+use crate::{Groups, Name};
+
+/// The UFO standard group prefix for kerns in the first position.
+pub const FIRST_KERNING_GROUP_PREFIX: &str = "public.kern1.";
+/// The UFO standard group prefix for kerns in the second position.
+pub const SECOND_KERNING_GROUP_PREFIX: &str = "public.kern2.";
```
RickyDaMa: Arguably should be in `groups.rs` instead

Hoolean: That would match [where it's defined in the spec](https://unifiedfontobject.org/versions/ufo3/groups.plist/), which is satisfying

src/groups.rs
```diff
@@ -18,7 +19,7 @@ pub(crate) fn validate_groups(groups_map: &Groups) -> Result<(), GroupsValidatio
             return Err(GroupsValidationError::InvalidName);
         }
 
-        if group_name.starts_with("public.kern1.") {
+        if group_name.starts_with(FIRST_KERNING_GROUP_PREFIX) {
             if group_name.len() == 13 {
```
Hoolean: nit: we could define this and the below in terms of the constant now

src/kerning.rs
RickyDaMa: Done, also cracked open the `groups` module to be public

src/kerning.rs
```diff
@@ -14,6 +17,45 @@ use crate::Name;
...
+impl From<&Groups> for ReverseGroupsLookup {
+    fn from(groups: &Groups) -> Self {
+        groups.iter().fold(ReverseGroupsLookup { first: HashMap::new(), second: HashMap::new() }, |mut rgl, (group_name, members)| {
+            let inverted = members.iter().map(|member| (member.clone(), group_name.clone()));
+            if group_name.starts_with(FIRST_KERNING_GROUP_PREFIX) {
+                rgl.first.extend(inverted);
+            } else if group_name.starts_with(SECOND_KERNING_GROUP_PREFIX) {
+                rgl.second.extend(inverted);
+            } else {
+                unreachable!("group didn't start with {FIRST_KERNING_GROUP_PREFIX} or {SECOND_KERNING_GROUP_PREFIX}");
```
Hoolean: Other custom group names are allowed for authors, [only the `public.`-prefix is reserved](https://unifiedfontobject.org/versions/ufo3/groups.plist/)

RickyDaMa: Thanks!

src/font.rs
```diff
@@ -604,6 +606,136 @@ impl Font {
...
+    /// Retrieve the kerning value (if any) between a pair of elements.
+    ///
+    /// The elements can be either individual glyphs (by name) or kerning groups
+    /// (by name), or any combination of the two.
+    //  ^ note: this works without any special consideration in the code
+    //          because glyph names are forbidden from using the group prefix,
+    //          thus meaning the group name lookup will always fail if a group
+    //          was passed in
+    ///
+    /// This method is accelerated by the [`ReverseGroupsLookup`], which can be
```
cmyr: I find 'accelerated' confusing here, since it's a required argument?

src/font.rs
```diff
@@ -604,6 +606,136 @@ impl Font {
...
+    ///
+    /// The elements can be either individual glyphs (by name) or kerning groups
+    /// (by name), or any combination of the two.
+    ///
+    /// ⚠️ This method forgoes pre-computing a reverse glyph groups lookup for the
+    /// use case where you're only after a couple of kerns and the overhead of
+    /// generating the lookup is not worth it. But in cases where you're
+    /// querying many kerns, you should definitely be using
+    /// [`Font::kerning_lookup`] over this method.
+    pub fn kerning_lookup_slow(&self, first: &str, second: &str) -> Option<f64> {
```
cmyr: I don't really love having two methods to do the same thing, I'd prefer to just figure out what our main use case is an support that well.

src/font.rs
cmyr: yes, this API makes more sense to me, I don't really think that `ReverseGroups` would need to be public API. I'm also not sure it's worth the complexity of having both a fast and slow API?

The only downside here is that the resolver holds a reference to things owned by the font, which means we can't mutate the font once it exists. I don't know how big of a problem this is?

cmyr (commented):
A couple notes in line, and the biggest point I'd make is already made in the earlier discussion: I think it would be nice if this was simplified to some kind of opaque `KerningResolver` type that ideally borrows data from the font. I'd also prefer to not have the two APIs, and just always compute the reverse map, unless there is some non-theoretical performance concern you're addressing.

Overall though, I'm happy with the idea of something like this being merged :)

src/font.rs
RickyDaMa: Yeah it would be nice to abstract away `ReverseGroups` entirely, however that then adds state to `Font`, which isn't something we currently have AFAIK, and personally I prefer the idea of the `norad::Font` remaining as pure data. Plus, if it's entirely internal, it's now on us to keep it up-to-date if a user changes groups/kerning. Doable of course.

`ReverseGroups` doesn't _need_ to hold a reference to things owned by the fonts, as `Name`s are cheaply cloneable. So on a technical level it's definitely possible

src/font.rs
RickyDaMa: Removed this phrasing, no longer relevant now there's only one public lookup method

RickyDaMa:
Hi @cmyr, I've updated the PR based on some of your comments. I've removed the slow version of the method, and updated the documentation of the remaining method.

I've made `ReverseGroups` depend on `Font` through lifetimes (which also saves cloning `Name`s).

> The only downside here is that the resolver holds a reference to things owned by the font, which means we can't mutate the font once it exists. I don't know how big of a problem this is?

Nothing stops the user from dropping their `ReverseGroups`. At least, in the current version of the API. I've kept `ReverseGroups` part of the public API for now, however from another [review comment](https://github.com/linebender/norad/pull/394#discussion_r3046261554) I'm relatively adverse to the idea:

> Yeah it would be nice to abstract away `ReverseGroups` entirely, however that then adds state to `Font`, which isn't something we currently have AFAIK, and personally I prefer the idea of the `norad::Font` remaining as pure data. Plus, if it's entirely internal, it's now on us to keep it up-to-date if a user changes groups/kerning. Doable of course.

src/font.rs
RickyDaMa: I've kept the fast method with the `ReverseGroups` and removed the other

src/font.rs
Hoolean: Thoughts on further renaming `ReverseGroups` to `KerningResolver`, and moving the `lookup` function from font to there too? I think that gets us the more-semantic public API we want, and stops the user from having to manage passing the object themselves

src/font.rs
```diff
@@ -604,6 +604,69 @@ impl Font {
...
+    /// The elements can be either individual glyphs (by name) or kerning groups
+    /// (by name), or any combination of the two.
+    //  ^ note: this works without any special consideration in the code
+    //          because glyph names are forbidden from using the group prefix,
+    //          thus meaning the group name lookup will always fail if a group
+    //          was passed in
+    ///
+    /// This method is requires a [`ReverseGroups`], which can be obtained from
+    /// [`Font::get_reverse_groups_lookup`].
+    pub fn kerning_lookup(
```
cmyr: okay sorry for all the back-and-forth, but what I had in mind was an API that looks something like,


```rust

struct LookupKerning<'a> {
    first_groups: HashMap<Name, Name>,
    second_groups: HashMap<Name, Name>,
    kerning: &'a Kerning,
}

impl LookupKerning<'_> {
    fn get(&self, first: &str, second: &str) -> Option<f64> {
        todo!("basically the kerning_lookup method in this PR")

impl Font {
    /// returns a thing you can lookup kerns in
    fn lookup_kerning<'a>(&'a self) -> LookupKerning<'a> {
        todo!("build the thing")
    }
}

/// and then it works like,

let my_font: Font = load_my_font();
let kemings = my_font.lookup_kerning();
let a_value = kemings.lookup("A", "V");



... does that make sense? Is there a reason to prefer the version you've provided here?

src/font.rs
RickyDaMa: Yep, definitely doable, I wouldn't mind that. Then there would be no way to not have `ReverseGroups`/`KerningResolver` not borrow from the parent `Font`, but we probably want that anyway.

src/font.rs
RickyDaMa: Back & forth is fine! Please don't feel bad, I actively enjoy discussing and refactoring this stuff. I like your/Harry's idea more too, I'll shift my implementation to match it 👍

src/font.rs
RickyDaMa: Resolving this thread as it's a duplicate of the later one with Colin's API sketch

cmyr (approved):
Great thanks, this looks good!

RickyDaMa:
Thanks for the merge! 🚀

PR #395, opened 2026-04-06
Non file io save and includes

yanone:
Follow up to https://github.com/linebender/norad/pull/390, a way to load and save UFO files from non-file sources such as hashmaps for instance in was, where native file system access doesn't exist (and can't be spoofed)


src/error.rs
```diff
@@ -35,6 +35,15 @@ pub enum DesignSpaceSaveError {
     /// A serialization error.
     #[error("failed to serialize designspace: {0}")]
     SeError(#[from] SeError),
+
+    /// Failed to write the designspace via a custom sink.
+    #[error("failed to write designspace file '{path}' via sink")]
```
RickyDaMa:
```suggestion
    #[error("failed to write designspace file '{path}' via sink: {source}")]
```

src/error.rs
```diff
@@ -142,6 +151,29 @@ pub enum FontLoadError {
...
+        path: PathBuf,
+    },
+    /// Included feature files form a cycle.
+    #[error("cyclic feature include detected for '{path}'")]
+    FeatureIncludeCycle {
+        /// The requested path.
+        path: PathBuf,
+    },
+    /// Failed to load a file through a custom non-file source.
+    #[error("failed to read '{path}' from non-file source")]
```
RickyDaMa:
```suggestion
    #[error("failed to read '{path}' from non-file source: {source}")]
```

src/error.rs
```diff
@@ -443,6 +475,14 @@ pub enum FontWriteError {
     /// There exists a `public.objectLibs` lib key when it should be set only by norad.
     #[error("the `public.objectLibs` lib key is managed by norad and must not be set manually")]
     PreexistingPublicObjectLibsKey,
+    /// Failed to write a file through a custom non-file sink.
+    #[error("failed to write '{path}' via non-file sink")]
```
RickyDaMa:
```suggestion
    #[error("failed to write '{path}' via non-file sink: {source}")]
```

src/error.rs
```diff
@@ -468,6 +508,14 @@ pub enum LayerWriteError {
     /// Failed to write out the layerinfo.plist file
     #[error("failed to write layerinfo.plist file")]
     LayerInfo(#[source] CustomSerializationError),
+    /// Failed to write a layer file via a custom sink.
+    #[error("failed to write layer file '{path}' via sink")]
```
RickyDaMa:
```suggestion
    #[error("failed to write layer file '{path}' via sink: {source}")]
```

src/font.rs
```diff
@@ -636,9 +700,12 @@ fn load_kerning(kerning_path: &Path) -> Result<Kerning, FontLoadError> {
     Ok(kerning)
 }
 
-fn load_features(features_path: &Path) -> Result<String, FontLoadError> {
-    let features = fs::read_to_string(features_path).map_err(FontLoadError::FeatureFile)?;
-    Ok(features)
+fn normalize_feature_text(contents: &str) -> Vec<u8> {
```
RickyDaMa: Is there a reason why you transform to a bytes buffer here? Based on the function name I would have expected `fn normalize_feature_text<'a>(contents: &'a str) -> Cow<'a, str>`, which allows you to return the original str if it's unedited, or an owned buffer otherwise

src/non_file_io.rs
```diff
@@ -0,0 +1,635 @@
...
+    })
+}
+
+fn read_required_text<F, E>(source: &mut F, path: &Path) -> Result<String, FontLoadError>
+where
+    F: FnMut(&Path) -> Result<Option<String>, E>,
+    E: StdError + Send + Sync + 'static,
+{
+    read_optional_text(source, path)?.ok_or(FontLoadError::MissingMetaInfoFile)
+}
```
RickyDaMa: This method's name makes it seem generic, but the error is specific to metainfo

src/non_file_io.rs
```diff
@@ -0,0 +1,635 @@
...
+fn load_layer_set_from_source<F, E>(
+    source: &mut F,
+    filter: &crate::data_request::LayerFilter,
+) -> Result<LayerContents, FontLoadError>
+where
+    F: FnMut(&Path) -> Result<Option<String>, E>,
+    E: StdError + Send + Sync + 'static,
+{
+    let layercontents_str = read_optional_text(source, Path::new(LAYER_CONTENTS_FILE))?
+        .ok_or(FontLoadError::MissingLayerContentsFile)?;
```
RickyDaMa: Unless I'm misreading the code, aren't you basically doing `read_required_text` with a different error variant?

src/non_file_io.rs
```diff
@@ -0,0 +1,635 @@
...
+                source: Box::new(LayerLoadError::ParsePlist { name: "contents.plist", source }),
+            })?;
+
+        for (_glyph_name, glif_relative_path) in glyph_files {
+            let glif_path = layer_dir.join(&glif_relative_path);
+            let glif_contents =
+                read_optional_text(source, &glif_path)?.ok_or(FontLoadError::Layer {
+                    name: layer_name.to_string(),
+                    path: glif_path.clone(),
+                    source: Box::new(LayerLoadError::MissingContentsFile),
```
RickyDaMa: Is this the right error variant?

src/non_file_io.rs
```diff
@@ -0,0 +1,635 @@
...
+            Component::CurDir => {}
+            Component::ParentDir => {
+                normalized.pop();
+            }
+            Component::Normal(part) => normalized.push(part),
+            Component::RootDir | Component::Prefix(_) => normalized.push(component.as_os_str()),
+        }
+    }
+    normalized
+}
```
RickyDaMa: Is [this function](https://doc.rust-lang.org/std/path/fn.absolute.html) doing the same thing as your impl here?

src/non_file_io.rs
```diff
@@ -0,0 +1,635 @@
...
+    F: FnMut(&Path, &[u8]) -> Result<(), E>,
+    E: StdError + Send + Sync + 'static,
+{
+    sink(path, bytes).map_err(|source| FontWriteError::Sink {
+        path: path.to_path_buf(),
+        source: Box::new(source),
+    })
+}
+
+fn normalize_feature_text(contents: &str) -> Vec<u8> {
```
RickyDaMa: How come this ended up duplicated?

RickyDaMa (commented):
Hey Yanone, was just having an idle nosy at your PR and thought I'd leave some feedback/questions on some of the low-hanging stuff I noticed. Very interested to see how a sans-io interface shapes up in `norad` 🤓 I'll let Colin guide the direction we take though

src/font.rs
yanone: Good point. I changed [normalize_feature_text](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to return [Cow<'_, str>](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and moved the byte conversion out to the write boundary. That matches the helper’s name better and avoids allocating when the contents don’t need normalization. I applied the same adjustment to the non-file save path as well so the two serialization paths stay consistent. Behavior is unchanged

src/non_file_io.rs
yanone: I renamed the helper to [read_required_metainfo_text](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) so it no longer reads like a generic required-file helper while still hardcoding the metainfo-specific error. That keeps the current behavior but makes the intent explicit.

src/non_file_io.rs
yanone: You weren’t misreading it. It was the same pattern with a different missing-file error. I changed it so [read_required_text](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is actually generic now: the caller provides the missing-file error variant, and both the metainfo and layercontents paths use the same helper.

src/non_file_io.rs
yanone: No, that wasn’t the right variant. [MissingContentsFile](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is specifically for a missing [contents.plist](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), not for a missing .glif. I changed the non-file path so a missing glyph file now comes through as [LayerLoadError::Glyph](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) with an [Io(NotFound)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) source, which matches the file-backed loader’s behavior.

src/non_file_io.rs
yanone: Not quite. [std::path::absolute](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) anchors relative paths to the host cwd and uses platform-specific absolute-path semantics, whereas this helper is just doing lexical normalization for UFO virtual paths coming from the non-file source. In particular, I do want to collapse . / .. inside that virtual namespace, not turn them into host-absolute paths. I renamed the helper to [normalize_virtual_path](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to make that distinction clearer.

src/non_file_io.rs
yanone: It ended up duplicated because the non-file save path was split out from the file-backed path and I copied the feature newline normalization along with it. I’ve removed the duplicate now: there’s a single shared [normalize_feature_text](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) helper in [non_file_io](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), and both save paths use it.

yanone:
@RickyDaMa I've addressed all raised issues. Happy for another review. Cheers

yanone:
Partially implemented by #405

PR #396, opened 2026-04-24
Skip empty glyph names in groups

cmyr:
This shows up in some real-world fonts, and doesn't seem like a helpful thing to error on.


dfrg (approved):
👍

PR #398, opened 2026-05-03
Support bare <condition> elements inside <rule>

cmyr:
The designspace spec allows <condition> elements directly inside <rule> without a <conditionset> wrapper for backwards compatibility. fonttools handles this by wrapping stray conditions into an implicit conditionset. This change mirrors that behavior using a RawRule intermediary struct, following the same from/into pattern used for RawDesignSpaceDocument.


dfrg (approved):
lgtm!

PR #401, opened 2026-05-06
Add no-fs feature for file-free loading

cmyr:
This adds an API for loading UFO fonts without directly interacting the filesystem. This is behind the 'no-fs' feature, and is not enabled by default.

@yanone See if this works for you for file-loading? Ultimately I thought it would be easier to just go back and forth with the LLM myself, instead of asking you to mediate. 🤓

Note that this only adds support for loading ufos, we can look at saving them subsequently?


yanone:
Hi @cmyr. Thanks.

For a font editor I need UFO and designspace loading and saving, as well as support for inline .fea file references, all of which my earlier PR supports.

This PR only covers a much narrower use case: loading of UFO files.

cmyr:
@yanone fair, but in any case this work would always need to be broken up into separate PRs; we'll start with loading and then we can look at saving.

yanone:
Okay, let's do it in stages. Since you said you'll be operating your LLM, I'm not doing anything until you ask me.

src/load_from_source.rs
```diff
@@ -0,0 +1,226 @@
...
+use crate::shared_types::Plist;
+use crate::DataRequest;
+
+pub(crate) fn load_font(
+    source: &impl FontSource,
+    request: DataRequest,
+) -> Result<Font, FontLoadError> {
+    // --- metainfo.plist (required) ---
+    let meta_str = read_required(source, Path::new(METAINFO_FILE))?;
+    let meta: MetaInfo = plist::from_reader(Cursor::new(meta_str.as_bytes()))
```
RickyDaMa, line 30: Nit: could just be doing `plist::from_bytes` here (the code is the same under the hood as is written here though, so not a big deal at all).

This pattern is repeated in several places, not just here.

src/font_source.rs
```diff
@@ -0,0 +1,61 @@
...
+/// let font = Font::load_from_source(DataRequest::all(), &source)?;
+/// # Ok(())
+/// # }
+/// ```
+pub trait FontSource {
+    /// The error type returned by [`read_contents`](FontSource::read_contents).
+    type Error: std::error::Error + Send + Sync + 'static;
+
+    /// Return the contents of the file at the given path, or `None` if it doesn't exist.
+    fn read_contents(&self, path: &Path) -> Result<Option<String>, Self::Error>;
```
RickyDaMa, line 48: My only consideration about this trait design is whether we'd also want to support a borrowed slice potentially, if we made it `FontSource<'a>` and the return type here `Result<Option<Cow<'a, str>>, Self::Error>`, or if that's just introducing more type tetris(tm) than the micro-optimisation is worth.

src/load_from_source.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+fn read_optional(
+    source: &impl FontSource,
+    path: &Path,
+) -> Result<Option<String>, FontLoadError> {
+    source
+        .read_contents(path)
+        .map_err(|e| FontLoadError::Source { path: path.into(), source: Box::new(e) })
+}
```
RickyDaMa, line 226: Is there a reason to keep these as private details instead of making them automatically-implemented methods of `FontSource`? I guess this way we prevent people from overwriting them.

src/load_from_source.rs
```diff
@@ -0,0 +1,226 @@
...
+        .read_contents(&contents_path)
+        .map_err(|e| LayerLoadError::Source {
+            path: contents_path.clone(),
+            source: Box::new(e),
+        })?
+        .ok_or(LayerLoadError::MissingContentsFile)?;
+
+    let contents: BTreeMap<Name, PathBuf> = plist::from_reader(Cursor::new(contents_str.as_bytes()))
+        .map_err(|source| LayerLoadError::ParsePlist { name: "contents.plist", source })?;
+    let path_set = contents.values().map(|p| p.to_string_lossy().to_lowercase()).collect();
```
RickyDaMa, line 164: Does it not matter here if paths get mangled because they're not UTF-8 and thus end up with substituted characters? I'm not intimately familiar with what this field of `Layer` is used for.

Or if by the power of the UFO spec we expect these to be UTF-8, I think that should be expressed in the code as (IIRC) `.to_str().expect("path was not UTF-8")` (or similar).

Cargo.toml
```diff
@@ -15,14 +15,16 @@ exclude = [
...
+features = ["kurbo", "no-fs"]
 
 [features]
 default = ["object-libs"]
 # allow us to add libs to objects, which requires potentially creating identifiers
 object-libs = ["uuid"]
 kurbo = ["dep:kurbo"]
 rayon = ["dep:rayon"]
+# Enable loading fonts from non-filesystem sources (e.g. in-memory, WASM).
+no-fs = []
```
RickyDaMa, line 27: If this feature isn't adding any new dependencies, is there a particular reason to feature-gate it? It doesn't seem intrusive to add to the API by default.

RickyDaMa (commented):
Just some passing questions / discussion points ✨ 

Also, this PR needs rebasing now `remove-string-interning` has been merged

cmyr:
going to go with the #405 approach.

PR #402, opened 2026-05-13
Preserve newlines around notes

madig:
@hoolean found an unfortunate roundtripping hindrance:

1. [ufoLib adds newlines around notes](https://github.com/fonttools/fonttools/blob/06e266ce70ec578d549c2df0e180a84d9323baf2/Lib/fontTools/ufoLib/glifLib.py#L949-L956)
2. [ufonormalizer preserves them](https://github.com/unified-font-object/ufoNormalizer/pull/86)
3. Norad strips them
4. ufonormalizer preserves that, too.

Norad trims all text in the .glif XML on read: https://github.com/linebender/norad/blob/fe37cc9427ce0f981fb63c6e83d9a60be264e9cb/src/glyph/parse.rs#L47

Maybe it should be more selective in its trimming, see https://docs.rs/quick-xml/latest/quick_xml/reader/struct.Config.html#method.trim_text.

### DIFF
--- a/src/glyph/tests.rs
+++ b/src/glyph/tests.rs
@@ -94,7 +94,9 @@ fn serialize_with_default_formatting() {
 			<string>I am a creative professional :)</string>
 		</dict>
 	</lib>
-	<note>durp</note>
+	<note>
+durp
+</note>
 </glyph>
 "#
     );
@@ -125,7 +127,9 @@ fn serialize_with_custom_whitespace() {
       <string>I am a creative professional :)</string>
     </dict>
   </lib>
-  <note>durp</note>
+  <note>
+durp
+</note>
 </glyph>
 "#
     );
@@ -155,7 +159,9 @@ fn serialize_with_single_quote_style() {
 			<string>I am a creative professional :)</string>
 		</dict>
 	</lib>
-	<note>durp</note>
+	<note>
+durp
+</note>
 </glyph>
 "#
     );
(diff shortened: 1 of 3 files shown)


RickyDaMa:
It sounds reasonable to me to have `norad` do the same `\n{note}\n` wrapping that the Python does, preserving the stripping on reading

cmyr:
so this does raise lots of questions like "is it okay to insert a newline if there wasn't one there when we loaded", but... maybe we don't need to overthink this? I think it's reasonable if we add them on save and strip them on load.

src/glyph/tests.rs
```diff
@@ -392,10 +400,17 @@ fn if_no_one_uses_your_lib_is_it_broken() {
...
-fn parse_note() {
+fn roundtrip_note() {
     let bytes = include_bytes!("../../testdata/note.glif");
     let glyph = parse_glyph(bytes).unwrap();
     assert_eq!(glyph.note, Some(".notdef".to_string()));
+
+    let buf = glyph.encode_xml().expect("encode failed");
+    let buf = std::str::from_utf8(&buf).unwrap();
+    assert!(
+        buf.find("<note>\n.notdef\n</note>").is_some(),
```
RickyDaMa: Nit

```suggestion
        buf.contains("<note>\n.notdef\n</note>"),
```

RickyDaMa (approved):

cmyr (approved):
thanks!

PR #403, opened 2026-05-13
Add better tests for lazy loading logic

cmyr:
This stuff is reasonably fancy and was definitely under-tested; I'm looking at adding support for wasm and ufoz, so I want to make sure that existing coverage is a bit better.

### DIFF
--- a/src/datastore.rs
+++ b/src/datastore.rs
@@ -552,4 +552,245 @@ mod tests {
 
         assert_eq!(store1, store2);
     }
+
+    // --- Helpers for synthetic on-disk stores ---
+
+    const PNG_HEADER: [u8; 8] = [137, 80, 78, 71, 13, 10, 26, 10];
+
+    fn png_data(extra: &[u8]) -> Vec<u8> {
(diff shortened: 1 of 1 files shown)


src/datastore.rs
```diff
@@ -552,4 +552,245 @@ mod tests {
...
+        let r1 = store.get(Path::new("ephemeral.txt")).unwrap();
+        assert!(r1.is_err());
+        assert!(matches!(
+            *store.items.get(Path::new("ephemeral.txt")).unwrap().borrow(),
+            Item::Error(_)
+        ));
+
+        // Second access: cached error (not re-attempted).
+        let r2 = store.get(Path::new("ephemeral.txt")).unwrap();
+        assert!(r2.is_err());
```
RickyDaMa: How does this verify the error is cached and it's not reattempted?

RickyDaMa (approved):
Looks reasonable to me, more tests never hurt

src/datastore.rs
cmyr: good call, updated this to actually check what's in the cell.

Issue #262, opened 2022-03-09
UFOZ support


madig:
More of a "maybe" issue. I personally have never seen UFOZ (zipped UFO packages) in the wild but apparently they do exist. Loading them would probably mean decoupling reading and writing files to disk and (de)serializing them into structures (i.e. have `from_path` and `from_bytes` methods on various structs where the former calls the latter.

PR #405, opened 2026-05-14
Add FontSource trait for source-agnostic UFO loading

cmyr:
Okay so after thinking about this more I realized that supporting this functionality would be a good opportunity for us to also implement #262, by generally decoupling reading from the file-system.

This PR does that; it closes #401.

@yanone sorry for the runaround, but this ultimately feels like the right approach, let me know if it works for your purposes?

### DIFF
--- a/src/datastore.rs
+++ b/src/datastore.rs
@@ -8,7 +8,7 @@ use std::{
 };
 
 use crate::error::{StoreEntryError, StoreError};
-use crate::font_source::FontSource;
+use crate::font_source::{DirEntry, FontSource};
 
 /// A generic file store for UFO [data][spec_data] and [images][spec_images],
 /// mapping [`PathBuf`] keys to [`Vec<u8>`] values.
@@ -141,13 +141,14 @@ impl DataType for Data {
                 .list_dir(&dir_path)
                 .map_err(|e| StoreEntryError::new(dir_path.clone(), e.into()))?;
 
-            for (entry_name, is_dir) in entries {
-                let full_rel = dir_path.join(&entry_name);
-                if is_dir {
-                    dir_queue.push(full_rel);
-                } else {
-                    let key = full_rel.strip_prefix(source_root).unwrap().to_path_buf();
-                    paths.push(key);
+            for entry in entries {
+                match entry {
+                    DirEntry::Dir(name) => dir_queue.push(dir_path.join(name)),
+                    DirEntry::File(name) => {
+                        let full_rel = dir_path.join(name);
+                        let key = full_rel.strip_prefix(source_root).unwrap().to_path_buf();
+                        paths.push(key);
+                    }
                 }
             }
         }
@@ -190,14 +191,11 @@ impl DataType for Image {
             .list_dir(source_root)
             .map_err(|e| StoreEntryError::new(source_root.to_path_buf(), e.into()))?;
 
-        for (entry_name, is_dir) in entries {
-            let full_rel = source_root.join(&entry_name);
-            if is_dir {
+        for entry in entries {
+            match entry {
                 // The spec forbids directories.
-                return Err(StoreEntryError::new(full_rel, StoreError::Subdir));
-            } else {
-                let key = full_rel.strip_prefix(source_root).unwrap().to_path_buf();
-                paths.push(key);
+                DirEntry::Dir(name) => return Err(StoreEntryError::new(name, StoreError::Subdir)),
+                DirEntry::File(name) => paths.push(name),
             }
         }
 
@@ -806,21 +804,18 @@ mod tests {
 
     /// A FontSource backed by in-memory data, with `as_path() -> None`.
     /// This triggers the eager-loading branch in `Store::new`.
+    #[derive(Default)]
     struct MemorySource {
         files: HashMap<PathBuf, Vec<u8>>,
-        dirs: HashMap<PathBuf, Vec<(PathBuf, bool)>>,
+        dirs: HashMap<PathBuf, Vec<DirEntry>>,
     }
 
     impl MemorySource {
-        fn new() -> Self {
-            MemorySource { files: HashMap::new(), dirs: HashMap::new() }
-        }
-
         fn add_file(&mut self, path: impl Into<PathBuf>, data: Vec<u8>) {
             self.files.insert(path.into(), data);
         }
 
-        fn add_dir(&mut self, path: impl Into<PathBuf>, entries: Vec<(PathBuf, bool)>) {
+        fn add_dir(&mut self, path: impl Into<PathBuf>, entries: Vec<DirEntry>) {
             self.dirs.insert(path.into(), entries);
         }
     }
@@ -830,7 +825,7 @@ mod tests {
             self.files.get(path).cloned().map(Ok)
         }
 
-        fn list_dir(&self, path: &Path) -> Result<Vec<(PathBuf, bool)>, std::io::Error> {
+        fn list_dir(&self, path: &Path) -> Result<Vec<DirEntry>, std::io::Error> {
             self.dirs.get(path).cloned().ok_or_else(|| {
                 std::io::Error::new(std::io::ErrorKind::NotFound, format!("{path:?} not found"))
             })
@@ -841,11 +836,9 @@ mod tests {
 
     #[test]
     fn data_eager_loading_from_memory_source() {
-        let mut source = MemorySource::new();
-        source.add_dir(
-            "data",
-            vec![(PathBuf::from("a.txt"), false), (PathBuf::from("b.txt"), false)],
-        );
+        let mut source = MemorySource::default();
+        source
+            .add_dir("data", vec![DirEntry::File("a.txt".into()), DirEntry::File("b.txt".into())]);
         source.add_file("data/a.txt", b"aaa".to_vec());
         source.add_file("data/b.txt", b"bbb".to_vec());
 
@@ -865,8 +858,8 @@ mod tests {
     #[test]
     fn image_eager_loading_from_memory_source() {
         let img = png_data(b"test");
-        let mut source = MemorySource::new();
-        source.add_dir("images", vec![(PathBuf::from("x.png"), false)]);
+        let mut source = MemorySource::default();
+        source.add_dir("images", vec![DirEntry::File("x.png".into())]);
         source.add_file("images/x.png", img.clone());
 
         let store = ImageStore::new(&source).unwrap();
@@ -879,8 +872,8 @@ mod tests {
 
     #[test]
     fn eager_load_invalid_image_fails_at_construction() {
-        let mut source = MemorySource::new();
-        source.add_dir("images", vec![(PathBuf::from("bad.png"), false)]);
+        let mut source = MemorySource::default();
+        source.add_dir("images", vec![DirEntry::File("bad.png".into())]);
         source.add_file("images/bad.png", b"not a png".to_vec());
 
         // Eager loading validates during construction, so this should fail.
@@ -890,10 +883,9 @@ mod tests {
 
     #[test]
     fn eager_load_nested_data_from_memory_source() {
-        let mut source = MemorySource::new();
-        source
-            .add_dir("data", vec![(PathBuf::from("top.txt"), false), (PathBuf::from("sub"), true)]);
-        source.add_dir("data/sub", vec![(PathBuf::from("deep.txt"), false)]);
+        let mut source = MemorySource::default();
+        source.add_dir("data", vec![DirEntry::File("top.txt".into()), DirEntry::Dir("sub".into())]);
+        source.add_dir("data/sub", vec![DirEntry::File("deep.txt".into())]);
         source.add_file("data/top.txt", b"top".to_vec());
         source.add_file("data/sub/deep.txt", b"deep".to_vec());
 
--- a/src/datastore.rs
+++ b/src/datastore.rs
@@ -8,15 +8,15 @@ use std::{
 };
 
 use crate::error::{StoreEntryError, StoreError};
+use crate::font_source::FontSource;
 
 /// A generic file store for UFO [data][spec_data] and [images][spec_images],
 /// mapping [`PathBuf`] keys to [`Vec<u8>`] values.
 ///
 /// The store provides a basic HashMap-like interface for checking data in and out.
-/// If initialized from disk, data can be loaded eagerly or lazily, as in, on access.
-/// It will remember the root data directory for this purpose. This complicates the
-/// accessor methods somewhat, because 1. access can fail with an IO error and 2.
-/// insertion can fail. Data is wrapped in a [`std::sync::Arc`] to help on-demand loading.
+/// If initialized from a filesystem-backed source, data is loaded lazily on access.
+/// Otherwise, data is loaded eagerly during construction.
+/// Data is wrapped in a [`std::sync::Arc`] to help on-demand loading.
 ///
 /// Note that it tracks files, not directories. Data paths you insert must not have
 /// any existing path in the store as an ancestor, or you would nest a file under a
@@ -67,7 +67,9 @@ use crate::error::{StoreEntryError, StoreError};
 #[derive(Debug, Clone)]
 pub struct Store<T> {
     items: HashMap<PathBuf, RefCell<Item>>,
-    ufo_root: PathBuf,
+    /// When `Some`, the source is filesystem-backed and lazy loading is used.
+    /// When `None`, all items were eagerly loaded during construction.
+    ufo_root: Option<PathBuf>,
     impl_type: T,
 }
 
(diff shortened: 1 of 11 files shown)


src/datastore.rs
```diff
@@ -186,36 +182,30 @@ impl DataType for Data {
...
+
+        for (entry_name, is_dir) in entries {
+            let full_rel = source_root.join(&entry_name);
+            if is_dir {
+                // The spec forbids directories.
+                return Err(StoreEntryError::new(full_rel, StoreError::Subdir));
             } else {
-                // ... and symlinks.
-                return Err(StoreEntryError::new(path, StoreError::NotPlainFile));
+                let key = full_rel.strip_prefix(source_root).unwrap().to_path_buf();
```
RickyDaMa: Instead of using `.strip_prefix` here, what's to stop you from just using `entry_name`? (Same in the impl for `Data`)

src/datastore.rs
```diff
@@ -793,4 +801,113 @@ mod tests {
...
+    /// This triggers the eager-loading branch in `Store::new`.
+    struct MemorySource {
+        files: HashMap<PathBuf, Vec<u8>>,
+        dirs: HashMap<PathBuf, Vec<(PathBuf, bool)>>,
+    }
+
+    impl MemorySource {
+        fn new() -> Self {
+            MemorySource { files: HashMap::new(), dirs: HashMap::new() }
+        }
```
RickyDaMa: For sake for tests, you could just use a `#[derive(Default)]` for this

src/font_source.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// List entries in a directory at the given relative path.
+    ///
+    /// Returns `(entry_name, is_dir)` pairs, where `entry_name` is the name
+    /// of each entry (not a full path). Callers should join with the directory
+    /// path to get the full relative path.
+    ///
+    /// The default implementation returns [`io::ErrorKind::Unsupported`],
+    /// meaning this source does not support directory enumeration. Data and
+    /// image stores will be empty for such sources.
+    fn list_dir(&self, _path: &Path) -> Result<Vec<(PathBuf, bool)>, io::Error> {
```
RickyDaMa: Could/Should this be an iterator instead of a pre-allocated list?

src/font_source.rs
```diff
@@ -0,0 +1,113 @@
...
+
+// Allow closures as FontSource for convenience.
+impl<F> FontSource for F
+where
+    F: Fn(&Path) -> Option<Result<Vec<u8>, io::Error>> + Sync,
+{
+    fn try_read(&self, path: &Path) -> Option<Result<Vec<u8>, io::Error>> {
+        self(path)
+    }
+}
```
RickyDaMa, line 142: I think a documentation comment for this would be helpful. It's not super discoverable written directly on the trait impl, but *somewhere* we should note that a closure implementing `try_read` (effectively) can be used as a `FontSource`

src/fontinfo.rs
```diff
@@ -496,23 +496,32 @@ impl FontInfo {
     /// converted by taking their absolute value. Fields that could be floats before and are
     /// integers now are rounded. Fields that could be floats before and are unsigned integers
     /// now are rounded before taking their absolute value.
+    #[allow(dead_code)]
```
RickyDaMa: If it's dead code and internal, why are we keeping it?

src/error.rs
```diff
@@ -303,6 +309,13 @@ impl StoreEntryError {
     pub(crate) fn new(path: PathBuf, source: StoreError) -> Self {
         Self { path, source }
     }
+
+    /// Returns `true` if the store could not be populated because the source
+    /// does not support directory listing or the directory does not exist.
+    pub(crate) fn is_missing_or_unsupported(&self) -> bool {
+        matches!(&self.source, StoreError::Io(e)
+            if matches!(e.kind(), std::io::ErrorKind::Unsupported | std::io::ErrorKind::NotFound))
```
RickyDaMa: This is more readable to me, but to each their own

```suggestion
        if let StoreError::Io(e) = &self.source {
            matches!(e.kind(), std::io::ErrorKind::Unsupported | std::io::ErrorKind::NotFound)
        } else {
            false
        }
```

RickyDaMa (commented):
I really like that you've rewritten the standard `Font` loading methods to use the new `FontSource` under the hood.

Have we benchmarked this PR at all to see how it compares for loading speed? Mainly curious

yanone:
> @yanone sorry for the runaround, but this ultimately feels like the right approach, let me know if it works for your purposes?

Sure, the direction works for me. What seems to be still missing here is the ability to open .fea includes(). In the end, after writing is supported, I need these to roundtrip cleanly.

Thanks.

src/font_source.rs
cmyr: we use this as a dyn trait, and that means  we can't have an associated type, and we also can't return an impl iterator; the only option would be `Box<dynamic Iterator>`, and at that point we're allocating and then paying for dynamic dispatch, so it probably isn't a meaningful win :(

src/font_source.rs
cmyr: yes good point, will add to the trait docs :)

src/datastore.rs
cmyr: good catch, was copy/pasta

cmyr:
okay I think this is ready to go; we can add designspace support afterwards.

@yanone I don't totally understand your issue with FEA includes, could you explain in a bit more detail? fea-rs does have support for handling includes in a way that doesn't require the file system.

It would be helpful for me if we can make sure that this branch will work before I commit to the new API, so if there's anything else you need to make that possible let me know, I have a bit more capacity right now and can be more responsive.

cmyr:
> I really like that you've rewritten the standard `Font` loading methods to use the new `FontSource` under the hood.
> 
> Have we benchmarked this PR at all to see how it compares for loading speed? Mainly curious

no measurable difference in time to load across ~40 test fonts.

src/datastore.rs
```diff
@@ -90,8 +92,8 @@ pub type ImageStore = Store<Image>;
 /// Defines custom behavior for data and images stores.
 #[doc(hidden)]
 pub trait DataType: Default {
```
RickyDaMa, line 94: Why is this gated on `Default`? General question, not PR specific / blocking

src/error.rs
```diff
@@ -162,6 +162,9 @@ pub enum LayerLoadError {
         /// The underlying error.
         source: GlifLoadError,
     },
+    /// An [`std::io::Error`].
+    #[error("failed to read layer data")]
+    Io(#[from] IoError),
```
RickyDaMa: Here (and below in `FontInfoLoadError`), did we not ought to include the inner error?


```suggestion
    #[error("failed to read layer data: {0}")]
    Io(#[from] IoError),
```

src/error.rs
```diff
@@ -188,6 +191,9 @@ pub enum FontInfoLoadError {
     /// The fontinfo.plist file contains invalid data.
     #[error("fontinfo.plist contains invalid data: {0}")]
     InvalidData(FontInfoErrorKind),
+    /// An [`std::io::Error`].
+    #[error("failed to read fontinfo.plist file")]
+    Io(#[from] IoError),
```
RickyDaMa: As above

```suggestion
    #[error("failed to read fontinfo.plist file: {0}")]
    Io(#[from] IoError),
```

src/font_source.rs
```diff
@@ -0,0 +1,129 @@
...
+///     }
+/// }
+/// ```
+pub trait FontSource: Sync {
+    /// Try to read the contents of a file at the given relative path.
+    ///
+    /// Returns `None` if the file does not exist, `Some(Ok(data))` if the
+    /// file was read successfully, or `Some(Err(..))` if the file exists but
+    /// could not be read.
+    fn try_read(&self, path: &Path) -> Option<Result<Vec<u8>, io::Error>>;
```
RickyDaMa, line 58: The choice of `io::Error` throughout this trait feels unintuitive to me considering that the aim of the PR was to support I/O-less workflows. Sure, you can build an `io::Error` with moreorless what you want, but maybe we should make the error type an associated constant so the implementer can choose it?

i.e.

```rs
pub trait FontSource: Sync {
    type Error;
    // -- snip --
}
```

src/font_source.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// The default returns `None`, meaning the store will eagerly load all data.
+    fn as_path(&self) -> Option<&Path> {
+        None
+    }
+
+    /// List entries in a directory at the given relative path.
+    ///
+    /// Returns `(entry_name, is_dir)` pairs, where `entry_name` is the name
+    /// of each entry (not a full path). Callers should join with the directory
+    /// path to get the full relative path.
```
RickyDaMa: I don't like that this has to be specified in documentation and isn't expressed in code.

Personally, I would make a struct like so

```rs
#[derive(Debug, Clone, /* etc. */)]
pub struct FontSourceEntry {
    pub relative_path: PathBuf,
    pub is_dir: bool,
}
```

and return that in the `Vec`.

src/font_source.rs
RickyDaMa, line 58: Less strong feelings on this one, but `Vec<u8>` is arguably a more prescriptive than it needs to be. We just need an owned buffer of bytes `Box<[u8]>`, not necessarily one that can be grown (the difference between the boxed slice and a `Vec<u8>`). Though a `Vec` is often used because it's convenient and easy to build.

I'd hesitate to add a generic for this because that makes more work for implementors, but equally switching to `Box<[u8]>` probably would too. Maybe I'm just rambling and we do nothing lol, it's probably fine

src/font_source.rs
```diff
@@ -0,0 +1,129 @@
...
+    fn list_dir(&self, _path: &Path) -> Result<Vec<(PathBuf, bool)>, io::Error> {
+        Err(io::Error::new(
+            io::ErrorKind::Unsupported,
+            "this FontSource does not support directory listing",
+        ))
+    }
+}
+
+/// A directory on disk implements [`FontSource`] directly.
+impl FontSource for &Path {
```
RickyDaMa, line 104: We could be really extra/helpful here, and easily update this impl to

```rs
impl<T: AsRef<Path>> FontSource for T {
    // -- snip --
}
```

Which would cover `PathBuf` and a bunch of string types too

src/layer.rs
RickyDaMa, line 358: Nit: it feels like there was a distinct adversity to using `?` when writing this code, but I think it'd make it a lot more readable, even if you have to state the return type of the closure explicitly

src/datastore.rs
cmyr: so we can construct one and call methods on it in a generic context:

https://github.com/linebender/norad/blob/99a9c4892e6d59e97eac773a14969ab126f83df7/src/datastore.rs#L234

(this code is very old and i'm not sure i'd do it the same today? but it seems... okay)

src/font_source.rs
cmyr: fair.. realistically i'm not sure what the error conditions would be in the wasm case? Moving to an associated type would introduce additional cascading changes, so my inclination right now would be to just leave this as a minor wart, and if it turns out to be an actual ergonomics problem in the wasm case then we can reconsider?

src/font_source.rs
cmyr: one better, we can just use an enum. will make that change.

src/font_source.rs
cmyr: I'd want to think about this pragmatically: where is the implementor getting these bytes, and is it ever _not_ from reading into a Vec? Converting Vec<u8> -> Box<[u8]> can cause a reallocation in some cases, although that isn't going to have a measurable impact on perf.

tldr; I'm open to this, but I'd want to understand that it is a concrete improvement on some practical axis, (e.g. ergonomics, performance, flexibility) and I don't see that at the moment.

src/font_source.rs
cmyr: we already have one blanket impl for the closure case, and adding a second breaks coherence, because 'in theory' you could have the same T that implements both `AsRef<Path> and Fn(&Path) -> thing`.

src/layer.rs
cmyr: I don't immediately see where `?` would work, here? If you want to offer a specific suggestion I'd be happy to consider, but playing around myself it isn't obvious.

src/font_source.rs
RickyDaMa: > Moving to an associated type would introduce additional cascading changes

Would it? For the existing I/O loading you'd just set the error type to be `io::Error` and then it's no different to the current state of the PR, but it opens up the flexibility to other implementors.

A lot of the basis for my comments is to try and get this in a good shape for its first version/release because it's easier to make changes while no one is depending on it. Up to you though.

src/font_source.rs
RickyDaMa: D'oh! Great idea

src/font_source.rs
RickyDaMa: Yep, fair argument, let's leave it.

src/font_source.rs
RickyDaMa: Urgh that's annoying. Personally I'd prefer `impl<T: AsRef<Path>> FontSource for T` over the closure case.

I don't know how often people are going to want to use a closure as their single source of logic for loading a font - it seems like the kind of thing you'd gravitate towards if you're writing example/test code and don't need any state - and even for this crate's own tests you sought additional state/functionality.

If people just want to implement a font loader in a single function they can effectively do that without this impl by making a ZST that implements `FontSource`. I'd want to optimise the ergonomics for the common case, which I think will still be path-likes even after we support sans-io.

src/layer.rs
RickyDaMa: Okay yeah fair enough, I tried to break it apart in my editor and see its pretty tightly woven to avoid repetition. Best I could do for readability is below, which I think is maybe marginally better. Equally, happy to leave it now that I know its as complex as it is out of necessity.

```rs
match source
    .read(&full_path)
    .map_err(GlifLoadError::Io)
    .and_then(|data| Glyph::parse(&data))
{
    Ok(mut glyph) => {
        glyph.name = name.clone();
        Ok((name, glyph))
    }
    Err(source) => Err(LayerLoadError::Glyph {
        name: name.to_string(),
        path: full_path,
        source,
    }),
}
```

src/font_source.rs
cmyr: bear in mind that the use of `&Path` as the impl type here is basically an internal plumbing detail, it isn't surfaced as public API; the public api all takes `impl AsRef<Path>`.

And although it remains tbd, I do think there's a case that the closure support is genuinely useful API, if (for instance) the caller just has a hashmap of path -> contents?

src/layer.rs
cmyr: yea this feels like a wash 🤷

src/font_source.rs
cmyr: We use `FontSource` as a dyn trait, and an associated type doesn't work in that context, so it would require some heftier re-architecting.

Realistically when I think about the wasm case, it's genuinely unclear what the error conditions would be? It's easy to imagine the case where a 'file' is missing, but barring that I'm genuinely unsure of the conditions under which a call to this API might fail? Happy for any insight on this..

src/font_source.rs
RickyDaMa: Ah yes okay I see what you mean, I lost sight of how little the public API touches a `FontSource` - it's basically only if you want something custom. Sorry! This is fine as-is then

src/font_source.rs
RickyDaMa: Yep again a fair point. I would be interested to hear from @yanone about what kind of failure modes we'd be likely to see from the WASM use case, but otherwise yeah I'm now satisfied this is sufficient for an initial release, and would be relatively unlikely to need revisiting

RickyDaMa (approved):
Thanks for graciously putting up with all my feedback/questions 🙏

cmyr:
Thanks for the review! will leave this open until @yanone has had a chance to actually try using it, but will merge if there's no problem there.

yanone:
Thank you.

The FontSource trait direction works well for me — it's a cleaner abstraction than the callback-based approach I took in #401, and having the standard Font::load rewired through it is a nice bonus.

For my font-editor use case, though, three things are still missing before this can replace my fork: a write/sink counterpart to FontSource (I need to save UFOs back out without a filesystem, mirroring load_from_source); resolution of .fea include() directives during load — not for compilation (fea-rs handles that), but so norad materializes the included source files into Font::feature_files and they round-trip cleanly on save; and non-filesystem writing for designspace files (reading already works via load_from_reader, but there's no string/sink-based write path). I've implemented all three in #401 and would be happy to port them onto the FontSource foundation here as follow-ups if you're open to it; if now isn't the right time, I'm also fine running a perpetual fork. I need the full functionality either way.

RickyDaMa:
Yeah to me those all seem like reasonable extensions that are compatible with the work on this PR, so we'd be fine to merge this as-is and build more onto it later. It seems like the design is good enough to support the additional requested features down the line.

cmyr:
Non-fs saving is a very reasonable followup for this, happy to take a look.

The one thing I'm confused about is the feature file resolution business. Can you explain this to me as if I were a small child?

In any case it sounds like this will work for the reading case, so I'm going to go ahead and merge :)

PR #406, opened 2026-05-14
Add DesignSpaceDocument::load_from_reader

cmyr:
This is specifically to support the no-fs use case (wasm, e.g.)

### DIFF
--- a/src/designspace.rs
+++ b/src/designspace.rs
@@ -3,7 +3,7 @@
 #![deny(rustdoc::broken_intra_doc_links)]
 
 use serde::{Deserialize, Serialize};
-use std::{fs::File, io::BufReader, path::Path};
+use std::path::Path;
 
 use plist::Dictionary;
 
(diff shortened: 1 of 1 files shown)


src/designspace.rs
```diff
@@ -335,7 +335,15 @@ pub struct Dimension {
...
     pub fn load<P: AsRef<Path>>(path: P) -> Result<DesignSpaceDocument, DesignSpaceLoadError> {
-        let reader = BufReader::new(File::open(path).map_err(DesignSpaceLoadError::Io)?);
+        let reader =
+            std::io::BufReader::new(std::fs::File::open(path).map_err(DesignSpaceLoadError::Io)?);
+        Self::load_from_reader(reader)
+    }
+
+    /// Load a designspace from a reader.
+    pub fn load_from_reader(
+        reader: impl std::io::BufRead,
```
RickyDaMa, line 345: Is there a reason we constrain on `BufRead` here, instead of just `Read`?

cmyr: it's because that's the trait bound used by the `quick_xml` crate, which we are passing this argument to. 🤷

RickyDaMa (approved):

PR #407, opened 2026-05-14
Add ufoz support

### LINKED ISSUE #262: UFOZ support

cmyr:
If passed a file (not a directory) and with the 'ufoz' feature enabled, norad will attempt to treat the file as a zip'd ufo directory.

- this is based on top of #405 
- closes #262

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ default = ["object-libs"]
 object-libs = ["uuid"]
 kurbo = ["dep:kurbo"]
 rayon = ["dep:rayon"]
+ufoz = ["zip"]
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 uuid = { version = "1.2", features = ["v4"], optional = true }
@@ -43,6 +44,7 @@ base64 = "0.22"
 close_already = "0.3"
 log = "0.4"
 smol_str = { version = "0.3", features = ["serde"] }
+zip = { version = "2", default-features = false, features = ["deflate"], optional = true }
 
 [dev-dependencies]
 failure = "0.1.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,15 +15,16 @@ exclude = [
 ]
 
 [package.metadata.docs.rs]
-features = ["kurbo"]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 default = ["object-libs"]
 # allow us to add libs to objects, which requires potentially creating identifiers
-object-libs = ["uuid"]
+object-libs = ["dep:uuid"]
 kurbo = ["dep:kurbo"]
 rayon = ["dep:rayon"]
-ufoz = ["zip"]
+ufoz = ["dep:zip"]
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 uuid = { version = "1.2", features = ["v4"], optional = true }
--- a/src/font.rs
+++ b/src/font.rs
@@ -217,11 +217,17 @@ impl Font {
 
     fn load_impl(path: &Path, request: DataRequest) -> Result<Font, FontLoadError> {
         let metadata = path.metadata().map_err(FontLoadError::AccessUfoDir)?;
-        if !metadata.is_dir() {
-            return Err(FontLoadError::UfoNotADir);
+        if metadata.is_dir() {
+            return Self::load_from_source(&request, &path);
         }
 
-        Self::load_from_source(&request, &path)
+        #[cfg(feature = "ufoz")]
+        if metadata.is_file() {
+            let source = crate::zip_source::ZipSource::open(path)?;
+            return Self::load_from_source(&request, &source);
+        }
+
+        Err(FontLoadError::UfoNotADir)
     }
 
     /// Returns a [`Font`] loaded from the given [`FontSource`].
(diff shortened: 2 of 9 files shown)


src/font.rs
```diff
@@ -217,11 +217,17 @@ impl Font {
...
         let metadata = path.metadata().map_err(FontLoadError::AccessUfoDir)?;
-        if !metadata.is_dir() {
-            return Err(FontLoadError::UfoNotADir);
+        if metadata.is_dir() {
+            return Self::load_from_source(&request, &path);
         }
 
-        Self::load_from_source(&request, &path)
+        #[cfg(feature = "ufoz")]
+        if metadata.is_file() {
```
RickyDaMa, line 225: Why are we only checking it's a file, not looking at the extension?

If we were ever to add JSON UFO loading, we're going to clash with this

src/zip_source.rs
```diff
@@ -0,0 +1,333 @@
...
+use crate::font_source::FontSource;
+
+/// A UFO source backed by a zip archive loaded into memory.
+pub(crate) struct ZipSource {
+    entries: HashMap<PathBuf, Vec<u8>>,
+}
+
+impl ZipSource {
+    /// Open a zip archive at the given path and load all file entries into memory.
+    pub fn open(path: &Path) -> Result<Self, FontLoadError> {
```
RickyDaMa: Would it be possible to just decompress what we need on the fly, instead of everything at once ahead of time?

If we store the archive in `ZipSource` instead of the decompressed entries. I guess the issue is that `FontSource` doesn't let you mutate `self` currently, but that could be changed or worked around I guess?

src/zip_source.rs
```diff
@@ -0,0 +1,333 @@
...
+    }
+}
+
+/// Detect if the zip has a single top-level directory wrapping all contents.
+///
+/// If so, return that directory name as the prefix to strip. Otherwise return
+/// `None`, meaning the zip contents are at the root level.
+pub(crate) fn detect_zip_root<R: io::Read + io::Seek>(
+    archive: &mut zip::ZipArchive<R>,
+) -> Option<PathBuf> {
```
RickyDaMa: Alternative impl that doesn't require always traversing the entire archive: pull out the first entry ahead of the loop, get its top-level dir (if it's a file early-return `None`), then run through everything else, immediately returning `None` if any other path doesn't have that same first component.

RickyDaMa (commented):
I think there are significant improvements that we could implement while everything is in flux. Sorry I don't have the time to offer to help beyond these little reviews 🙏

src/font.rs
cmyr: my thinking is basically: if it's a directory, try to load it as a ufo. If it's a file, try to load it as a ufoz. I believe in the wild I've seen zipped ufos with the .zip extension, and I'm not sure what doing the extension check really buys us?

RickyDaMa: Compatibility with e.g. JSONified UFOs or any other single-file UFO formats we may want to support in future. Granted they're not enshrined in the spec the same as I believe ufoz is, but JSONified UFOs are phenomenally useful to work with in build systems as it massively reduces I/O usage while still keeping human readable files for debugging.

cmyr: that makes sense, and i'm not opposed to adding json support if it would be genuinely helpful (and it might be worth even proposing as an extension to the spec, if you're finding it has real-world benefits. In any case I propose we punt on this until support for that materializes, and then we can do something fancier here like actually checking for the zip's magic bytes.

src/zip_source.rs
cmyr: Yea, so this is possible but we'd either need the trait to take &mut self, or we'd need interior mutability.

Ultimately it comes down to the tradeoffs: doing it all up front increases peak memory usage but is also more efficient, since we don't need to seek around the zip contents and do redundant deflating?

One issue I would point out here though is that the current design means we deflate everything even if we have a sparse `DataRequest`, and that does smell bad, but it's easy enough for us to thread `DataRequest` through to `ZipSource::open` and then avoid that work, so i'll make that change.

src/zip_source.rs
cmyr: Agree that's an improvement!

cmyr:
@RickyDaMa if this looks good to you i'll get it in :)

Cargo.toml
```diff
@@ -23,6 +23,7 @@ default = ["object-libs"]
 object-libs = ["uuid"]
 kurbo = ["dep:kurbo"]
 rayon = ["dep:rayon"]
+ufoz = ["zip"]
```
RickyDaMa: I believe otherwise we end up with an implicit feature called `zip`

```suggestion
ufoz = ["dep:zip"]
```

RickyDaMa (approved):
One final thing that could be nice to do is to add this to our Cargo.toml:

```toml
[package.metadata.docs.rs]
all-features = true
rustdoc-args = ["--cfg", "docsrs"]
```

So that the docs are generated for all features, and feature-specific API is labelled with what feature has to be enabled.

It's not the cleanest example because the underlying code is generated, but you can see this in the docs for [`static-lang-word-lists`' constants](https://docs.rs/static-lang-word-lists/latest/static_lang_word_lists/#statics)

PR #409, opened 2026-06-09
Add verticalOrigin accessor method

madig:
We thought we could have an official accessor for a glyph's vertical origin. Feel free to code golf this.

Harry helped flesh this out :)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -662,6 +662,11 @@ impl From<IoError> for StoreError {
 #[non_exhaustive]
 pub enum LibRetrievalError {
     /// A lib key has unexpected data.
-    #[error("the glif lib key '{0}' has invalid data, expected {1}")]
-    InvalidData(&'static str, &'static str),
+    #[error("the glif lib key '{key}' has invalid data, expected {expectation}")]
+    InvalidData {
+        /// They sought after lib key.
+        key: &'static str,
+        /// What data type was expected.
+        expectation: &'static str,
+    },
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -661,12 +661,10 @@ impl From<IoError> for StoreError {
 #[derive(Debug, Error)]
 #[non_exhaustive]
 pub enum LibRetrievalError {
-    /// A lib key has unexpected data.
-    #[error("the glif lib key '{key}' has invalid data, expected {expectation}")]
-    InvalidData {
-        /// They sought after lib key.
-        key: &'static str,
-        /// What data type was expected.
-        expectation: &'static str,
+    /// The `public.verticalOrigin` key has unexpected data.
+    #[error("'public.verticalOrigin' must be a number of reasonable size, but was '{value:?}'")]
+    BadVerticalOrigin {
+        /// The unexpected value.
+        value: plist::Value,
     },
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -656,3 +656,12 @@ impl From<IoError> for StoreError {
         StoreError::Io(std::sync::Arc::new(src))
     }
 }
+
+/// Errors for when retrieving lib data.
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum LibRetrievalError {
+    /// A lib key has unexpected data.
+    #[error("the glif lib key '{0}' has invalid data, expected {1}")]
+    InvalidData(&'static str, &'static str),
+}
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -247,28 +247,31 @@ impl Glyph {
     }
 
     /// Set the vertical origin of the glyph, and returns an attempted parse of
-    /// its previous value, if present. See `public.verticalOrigin` in the
-    /// [Common Key Registry].
+    /// its previous value, if present. If passing None, remove the vertical
+    /// origin of the glyph, and returns an attempted parse of its previous
+    /// value, if present. See `public.verticalOrigin` in the [Common Key
+    /// Registry].
     ///
     /// [Common Key Registry]:
     ///     https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#publicverticalorigin
-    pub fn set_vertical_origin(&mut self, value: f64) -> Option<Result<f64, LibRetrievalError>> {
-        let plist = if (value - value.round()).abs() < f64::EPSILON {
-            plist::Value::Integer((value as i64).into())
+    pub fn set_vertical_origin(
+        &mut self,
+        value: impl Into<Option<f64>>,
+    ) -> Option<Result<f64, LibRetrievalError>> {
+        let value = value.into();
+        if let Some(value) = value {
+            let plist = if value.fract() == 0.0 {
+                plist::Value::Integer((value as i64).into())
+            } else {
+                plist::Value::Real(value)
+            };
+            self.lib
+                .insert(PUBLIC_VERTICAL_ORIGIN.into(), plist)
+                .as_ref()
+                .map(plist_to_vertical_origin)
         } else {
-            plist::Value::Real(value)
-        };
-        self.lib.insert(PUBLIC_VERTICAL_ORIGIN.into(), plist).as_ref().map(plist_to_vertical_origin)
-    }
-
-    /// Remove the vertical origin of the glyph, and returns an attempted parse
-    /// of its previous value, if present. See `public.verticalOrigin` in the
-    /// [Common Key Registry].
-    ///
-    /// [Common Key Registry]:
-    ///     https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#publicverticalorigin
-    pub fn remove_vertical_origin(&mut self) -> Option<Result<f64, LibRetrievalError>> {
-        self.lib.remove(PUBLIC_VERTICAL_ORIGIN).as_ref().map(plist_to_vertical_origin)
+            self.lib.remove(PUBLIC_VERTICAL_ORIGIN).as_ref().map(plist_to_vertical_origin)
+        }
     }
 }
 
--- a/src/glyph/mod.rs
+++ b/src/glyph/mod.rs
@@ -259,19 +259,17 @@ impl Glyph {
         value: impl Into<Option<f64>>,
     ) -> Option<Result<f64, LibRetrievalError>> {
         let value = value.into();
-        if let Some(value) = value {
+        let old = if let Some(value) = value {
             let plist = if value.fract() == 0.0 {
                 plist::Value::Integer((value as i64).into())
             } else {
                 plist::Value::Real(value)
             };
-            self.lib
-                .insert(PUBLIC_VERTICAL_ORIGIN.into(), plist)
-                .as_ref()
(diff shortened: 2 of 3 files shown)


Hoolean:
(This repo felt like a nice home as the lib key is very well-defined in the spec 🏡 When this lands, there are a few projects downstream that can simplify too too, e.g. [the utility function I added to ufo2fontir in fontc can go](https://github.com/googlefonts/fontc/blob/ca5d0a6f7150b497c52bdbf7b6d15d1195c0043b/ufo2fontir/src/source.rs#L701-L718))

src/error.rs
```diff
@@ -656,3 +656,12 @@ impl From<IoError> for StoreError {
...
     }
 }
+
+/// Errors for when retrieving lib data.
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum LibRetrievalError {
+    /// A lib key has unexpected data.
+    #[error("the glif lib key '{0}' has invalid data, expected {1}")]
+    InvalidData(&'static str, &'static str),
```
RickyDaMa: I'd use named variant values here

src/glyph/mod.rs
```diff
@@ -233,6 +235,42 @@ impl Glyph {
...
+            "an integer or float of reasonable size",
+        )))
+    }
+
+    /// Set the vertical origin of the glyph. Use a value of `None` to unset it.
+    /// See `public.verticalOrigin` in the [Common Key Registry].
+    ///
+    /// [Common Key Registry]:
+    ///     https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#publicverticalorigin
+    pub fn set_vertical_origin(&mut self, value: Option<f64>) {
```
RickyDaMa: Personally I'd rather have an explicit remove method than this set-with-`Option` flavour

RickyDaMa: Also, we could return the previous value here, similar to [`plist::Dictionary::insert`](https://docs.rs/plist/latest/plist/dictionary/struct.Dictionary.html#method.insert) and `HashMap::insert`

src/error.rs
madig: You mean for both inner fields?

src/glyph/mod.rs
madig: I was wondering about this; what if we add more accessors for lib keys? Add a variant of all of these for each lib key?

src/error.rs
RickyDaMa: Yes, you can do:

```rust
enum LibRetrievalError {
    #[error("the glif lib key '{key}' has invalid data, expected {expected_type}")]
    InvalidData {
        key: &'static str,
        expected_type: &'static str,
    },
    // -- snip --
}
```

Not 100% confident on the syntax within `#[error]`, but you get the idea

src/glyph/mod.rs
RickyDaMa: If your aversion is creating boilerplate / excessive lines of code, then that's what declarative macros are for: stamping out repeated code patterns.

madig: That's one option 🤔 I had something like `pub fn get_lib(key: LibKey)`, etc. in mind where `LibKey` is an enum, but it struck me as probably a bad idea.

Hoolean: The problem with a shared `get_lib` is that not all of the spec-defined lib keys have the same type

RickyDaMa: Yeah I don't see a benefit to that over just having individual methods. Plus, at this point we're talking about one method for one lib key. We can procrastinate the issue until there's at least three of the things imo 😁

src/glyph/mod.rs
```diff
@@ -233,6 +235,50 @@ impl Glyph {
...
+        } else {
+            plist::Value::Real(value)
+        };
+        self.lib.insert(PUBLIC_VERTICAL_ORIGIN.into(), plist);
+    }
+
+    /// Remove the vertical origin lib key data and return its value, if any.
+    ///
+    /// Returns None if the data cannot be parsed into a number of reasonable
+    /// size.
```
RickyDaMa: We should note that even if the number doesn't fit into an `f64`, it will still have been removed.

src/glyph/mod.rs
```diff
@@ -233,6 +235,50 @@ impl Glyph {
...
+            plist::Value::Integer(integer) => integer.as_signed().map(|v| v as f64),
+            _ => None,
+        };
+        Some(parsed.ok_or(LibRetrievalError::InvalidData {
+            key: PUBLIC_VERTICAL_ORIGIN,
+            expectation: "an integer or float of reasonable size",
+        }))
+    }
+
+    /// Set the vertical origin of the glyph. Use a value of `None` to unset it.
```
RickyDaMa: This comment is outdated.

src/glyph/mod.rs
```diff
@@ -233,6 +235,50 @@ impl Glyph {
...
+    /// See `public.verticalOrigin` in the [Common Key Registry].
+    ///
+    /// [Common Key Registry]:
+    ///     https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#publicverticalorigin
+    pub fn set_vertical_origin(&mut self, value: f64) {
```
RickyDaMa: I still think it would be nice to return the old value here, even if just as an `Option<plist::Value>`

Hoolean: It seems like a leaky abstraction if we provide the old value here and in the `remove` function as an unvalidated `plist`. If someone really needs that functionality, they can just work on the `lib` themselves with the key constant we provide

Hoolean: I'd be in favour of removing it, but I think if we do still provide it on `remove` (where I can admittedly see the utility) then we should at least provide the same `Option<Result<_>>` validation

src/glyph/mod.rs
```diff
@@ -233,6 +235,53 @@ impl Glyph {
...
+    }
+
+    /// Set the vertical origin of the glyph, and returns an attempted parse of
+    /// its previous value, if present. See `public.verticalOrigin` in the
+    /// [Common Key Registry].
+    ///
+    /// [Common Key Registry]:
+    ///     https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#publicverticalorigin
+    pub fn set_vertical_origin(&mut self, value: f64) -> Option<Result<f64, LibRetrievalError>> {
+        let plist = if (value - value.round()).abs() < f64::EPSILON {
```
cmyr: so this is mostly academic, but there is a funny case here:

an f64 can represent values larger than i64; these values would pass the check here, but would be saturated when cast to i64.

ultimately I do not think that `1e300` is a very reasonable vertical origin, but thought i would point it out.

I think the epsilon check is also a bit overkill, and i'd just do `if value.fract() == 0.0`?

src/glyph/mod.rs
```diff
@@ -233,6 +235,53 @@ impl Glyph {
...
+    /// Remove the vertical origin of the glyph, and returns an attempted parse
+    /// of its previous value, if present. See `public.verticalOrigin` in the
+    /// [Common Key Registry].
+    ///
+    /// [Common Key Registry]:
+    ///     https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#publicverticalorigin
+    pub fn remove_vertical_origin(&mut self) -> Option<Result<f64, LibRetrievalError>> {
+        self.lib.remove(PUBLIC_VERTICAL_ORIGIN).as_ref().map(plist_to_vertical_origin)
+    }
+}
```
cmyr: idk? maybe you had this initially, but my actual preference here would be to make the `set_vertical_origin` fn take `impl Into<Option<f64>>`, and then the user can pass `None` if they want to remove? Did you try this and not like it?

src/glyph/mod.rs
```diff
@@ -233,6 +235,53 @@ impl Glyph {
...
+    ///
+    /// [Common Key Registry]:
+    ///     https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#publicverticalorigin
+    pub fn set_vertical_origin(&mut self, value: f64) -> Option<Result<f64, LibRetrievalError>> {
+        let plist = if (value - value.round()).abs() < f64::EPSILON {
+            plist::Value::Integer((value as i64).into())
+        } else {
+            plist::Value::Real(value)
+        };
+        self.lib.insert(PUBLIC_VERTICAL_ORIGIN.into(), plist).as_ref().map(plist_to_vertical_origin)
```
cmyr: slightly funny that this would return an error when replacing a bad value, even if we've succeeded. Fine, though..

src/error.rs
```diff
@@ -656,3 +656,17 @@ impl From<IoError> for StoreError {
         StoreError::Io(std::sync::Arc::new(src))
     }
 }
+
+/// Errors for when retrieving lib data.
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum LibRetrievalError {
```
cmyr, line 663: so the questions here:
- do we want an error at all, or do we just treat a bad origin as `None`? (probably an error)
- if we _do_ want an error, do we want _this one_? This is trying to generalize and imagine some future needs. Are those needs concrete? My preference might just be like,


```rust
#[error("public.verticalOrigin should be number type, found '{found_type}'"]
struct BadVerticalOrigin {
    found_type: &'static str,
}
```

alternatively, we could have an error that was just `struct BadLibValue`, and also bundled the field name.

src/error.rs
```diff
@@ -656,3 +656,17 @@ impl From<IoError> for StoreError {
...
+    /// A lib key has unexpected data.
+    #[error("the glif lib key '{key}' has invalid data, expected {expectation}")]
+    InvalidData {
+        /// They sought after lib key.
+        key: &'static str,
+        /// What data type was expected.
+        expectation: &'static str,
```
cmyr: what was expected is probably less useful as a diagnostic than what was found? Or at least equally useful.

cmyr (approved):
sure if you think this would be useful? Is there a particular reason this would be useful for verticalOrigin but not horizontal?

left some notes inline but i'm not opposed

src/glyph/mod.rs
RickyDaMa: I did not think of `impl Into<Option<f64>>`, initially it was just `Option<f64>`, which I objected to. The `impl Into` version I'd be okay with, though I still personally prefer separate methods

src/error.rs
RickyDaMa: imo it's potentially useful to distinguish between no value and a bad value; it should be up to the downstream user to collapse these two code paths if they want to handle them similarly. That doesn't mean I like `Option<Result<T, E>>` as a return value / function signature, but that's kinda what you end up with

src/glyph/mod.rs
madig: I agree, but Harry opined that this gives downstream options. We can still change it if we find it to be useless :)

madig:
> Is there a particular reason this would be useful for verticalOrigin but not horizontal?

The horizontal origin is always zero I think 😁

RickyDaMa (approved):

cmyr (approved):
looks good, sorry for the wait!

PR #410, opened 2026-06-19
Add links and badges to crates.io, docs, and CI

Hoolean:
I often come looking for these here and crave them, so putting a wee PR up. The README would probably benefit from some of the nice docs we have in the module already, but will leave that for a future PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+[![crates.io](https://img.shields.io/crates/v/norad.svg)](https://crates.io/crates/norad)
+[![docs.rs](https://img.shields.io/docsrs/norad.svg)](https://docs.rs/norad/latest/norad/)
+[![Continuous integration](https://github.com/linebender/norad/actions/workflows/rust.yml/badge.svg)](https://github.com/linebender/norad/actions/workflows/rust.yml)
+
 # Norad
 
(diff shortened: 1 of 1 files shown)


RickyDaMa:
Honestly you could just move the crate-level docs into the README, and then in-line the README.md into the docs, like [this](https://codeberg.org/alpha-tango-kilo/close_already/src/commit/c5265c3e7a15a7ed7f5dcc42a7d28a09142028ee/src/lib.rs#L1).

There are some rustdoc-isms you may have to translate, but it's a pretty quick process generally speaking

RickyDaMa (approved):

Hoolean:
> Honestly you could just move the crate-level docs into the README, and then in-line the README.md into the docs, like [this](https://codeberg.org/alpha-tango-kilo/close_already/src/commit/c5265c3e7a15a7ed7f5dcc42a7d28a09142028ee/src/lib.rs#L1).

That would be nice I think :+1: Let's see what people say; I've spun it out into #411, in case we decide to have one without the other.

RickyDaMa:
Going to merge this as it shouldn't conflict with #411, and I don't think pretty badges are going to be a controversial move

PR #411, opened 2026-06-19
Populate README with module docs, and share with an include

Hoolean:
In general, this PR tries to change as little of the content as possible, but has to make a few concessions with links to keep them working in both environments.


RickyDaMa (approved):
Personally I'd be really happy to see norad with a prettier, more fleshed out README!

cmyr (approved):
thanks!

PR #414, opened 2026-07-13
Treat empty guideline names as no name

tomconroy:
Norad failed to parse a real-world UFO that has a guideline with an empty name. This patch allows empty guideline names.

See: https://github.com/googlefonts/roboto-serif/blob/8346c2b0ecec6bf539be222ea84de5906ef1ade8/sources/Roboto_Serif_Micro_Cond-Super.ufo/fontinfo.plist#L22-L23

### DIFF
--- a/src/fontinfo.rs
+++ b/src/fontinfo.rs
@@ -1725,6 +1725,16 @@ mod tests {
     }
 
     #[test]
+    fn fontinfo_empty_guideline_name() {
+        let path = "testdata/fontinfo_empty_guideline_name.plist";
+        let font_info: FontInfo = plist::from_file(path).expect("failed to load fontinfo");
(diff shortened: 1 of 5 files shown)


RickyDaMa (commented):
This seems like a sensible workaround to me, although the current implementation is spec-compliant:

> `name`: An arbitrary name for the guideline. This attribute is optional. It must be at least one character long.

^ https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#guideline

Which basically says that if the field is going to be specified it should be non-empty.

No strong opinion from me either way.

cmyr (approved):
Thanks! given that the name field is explicitly optional I'm happy to skip empty strings.

Issue #416, opened 2026-07-29
Support DesignSpace v5.x data


madig:
* Axes elided fallback name https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#axes-element
* Discrete axes https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#axis-element
* Axis labels 
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#labels-element-axis
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#label-element-axis
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#labelname-element-axis-stat-label
* Mappings element 
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#mappings-element
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#mapping-element
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#input-element
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#output-element
* Top-level labels element
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#labels-element-top-level
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#label-element-top-level
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#location-element-top-level-stat-label
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#dimension-element
    * https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#labelname-element-top-level-stat-label
* Localised names for sources https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#familyname-element-localised-names-for-sources
* Variable fonts element and subelements https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#variable-fonts-element
* Instance element
    *  Optional location attribute https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#instance-element
    * fontinfo dicts https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#example-for-varlib

PR #418, opened 2026-09-22
designspace v5 (take 2)

cmyr:
Adds the v5 elements from #416: elidedfallbackname, axis STAT labels, top-level location labels, variable-fonts with axis-subsets, localised source family names, and instance location labels and localised names. The same structs handle 4.x and 5.x; the only API breakage is new pub fields on existing structs.

An axis-subset is modelled as a name plus SubsetValue (Whole, Tent or Discrete), parsed via a raw struct rather than an untagged enum, which quick-xml can't handle. On save, format is bumped to the minimum version the content needs, as fontTools does.

Based on #340 by @madig; the test file is fontTools' test_v5.designspace plus his additions.


madig:
Thank you for working on this! This LGTM. I added a tiny change to make sure the DSv5 test data exercises two probably seldom used attributes.

About the various libs that have been added to element structs, I had the thought of extracting and parsing the `public.fontInfo` key into a proper `FontInfo` struct, accessible next to an element's lib field (so e.g. `VariableFont` would have a `lib` and `font_info` field), but I'm not sure if that's a good idea or should be left to the compiler...

madig (approved):

src/designspace.rs
```diff
@@ -37,37 +52,345 @@ pub struct DesignSpaceDocument {
...
+                SubsetValue::Discrete(user_value)
+            }
+            None => match (user_minimum, user_default, user_maximum) {
+                (Some(minimum), Some(default), Some(maximum)) => {
+                    SubsetValue::Range { minimum, default, maximum }
+                }
+                (None, None, None) => SubsetValue::Full,
+                _ => {
+                    return Err(AxisSubsetError(
+                        "axis-subset element must have min/max/default values or none at all",
```
anthrotype: so.. this error matches fontTools designspace reader
https://github.com/fonttools/fonttools/blob/718b61b552841de3964ba5b691fcf6102a2ebb3c/Lib/fontTools/designspaceLib/__init__.py#L2377

however the spec https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#axis-subset-element explicitly makes userminimum, usermaximum and userdefault individually optional. The missing bounds come from the parent axis, whereas an omitted default should use the parent default clamped into the subset, something like `max(subset_min, min(parent_default, subset_max))`.

The problem is fontTools designspace writer is inconsistent with its reader: e.g. create an axis Weight 100/400/900 and a variable-font `RangeAxisSubsetDescriptor(name=”Weight”, userMinimum=700)`, then serialize. fontTools writes:

```xml
<axis-subset name="Weight" userminimum="700"/>
```

which both readers reject.

Since we are adding this here, might be worth following the spec and update fonttools impl separately?

anthrotype: fwiw I opened https://github.com/fonttools/fonttools/pull/4205 for the fontTools side.

For the model here, what if we fold `Full` into `Range` and make all three fields optional?

```rust
    pub enum SubsetValue {
        /// Missing values default to the parent axis (default clamped to the subset).
        Range { minimum: Option<f64>, default: Option<f64>, maximum: Option<f64> },
        Discrete(f64),
    }
```

That's how fontTools models it too (the whole axis is just a `RangeAxisSubsetDescriptor` with no bounds set), and it would avoid having two ways to spell the whole axis (Full vs an all-None Range). The TryFrom only needs to reject uservalue mixed with any of the range attributes. The all-None `Range` can be the `#[default]`, maybe with an `is_full()` helper so we don't lose the readability of `Full`.

anthrotype (approved):
LGTM, but see my comments above on partial axis-subset ranges

src/designspace.rs
cmyr: yea, that sounds good!

